Issue #10, opened 2015-11-10
We do not need Joda time library


buma:
I looked into the code and saw that Joda LocalDate is used only in JodaLocalDateSerializer/Deserializer otherwise it is just imported and not used. In OTP it made sense to use it but since Java 8 has its own LocalDate classes I would like to know if we specifically need joda LocalDate or is java [LocalDate](https://docs.oracle.com/javase/8/docs/api/java/time/LocalDate.html) enough. Since it also means one library less.

abyrd:
I noticed this too. My understanding is that Java 8 incorporates the Joda time API so we don't need the external dependency anymore. It's just a holdover from OTP. 

So feel free to remove it if you like.

buma:
I removed it in jodaRemoval branch, but I also noticed that it is still included since gtfs-lib uses it in 5 files and we call isServiceActive on gtfs-lib which has Joda LocalDate as a parameter.

abyrd:
We should switch all libraries to Java 8 and remove the Joda time dependency everywhere. Unless there's some conflict. @mattwigway?

mattwigway:
Agreed. I tried this once before and it was a little messy because in the current Analyst we have ProfileRequests that are serialized, but now that we're making a clean break that shouldn't be a problem.

mattwigway:
As I understand the Java date classes are based Joda so should provide very similar functionality.

abyrd:
The new Java time system was co-designed by the author of joda time to provide similar functionality. The Joda Time website asks users to migrate to java.time starting with Java 8.

buma:
I also removed joda from GTFS-lib in conveyal/gtfs-lib#5. So now we have 2 options. If there are no problems with Joda removal in gtfs-lib we can merge #35 otherwise we can merge #21 since it keeps joda in gtfs-lib.

PR #21, opened 2015-11-13
Joda removal

buma:
Removes all usage of Joda library from R5.

The only usage is in services in transitLayer to communicate with gtfs-lib more in #10.


abyrd:
We have decided to remove Joda from gtfs-lib, so this pull request is superseded by #35. I am closing it without merging.

PR #35, opened 2015-12-04
Joda complete removal

buma:
It removes all joda calls in R5 and also from calls to gtfs-lib since it depends on conveyal/gtfs-lib#5 to be merged.

It can be merged if there are no problems with removing joda in gtfs-lib otherwise #21 can be merged.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -386,11 +386,6 @@
       <version>1.10.28</version>
     </dependency>
     <dependency>
-      <groupId>joda-time</groupId>
-      <artifactId>joda-time</artifactId>
-      <version>2.8.2</version>
-    </dependency>
-    <dependency>
       <groupId>ch.qos.logback</groupId>
       <artifactId>logback-classic</artifactId>
       <version>1.1.3</version>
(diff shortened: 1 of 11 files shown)


buma:
Travis fails because GTFS-lib service.activeOn is called with assumption that joda is removed and because it isn't yet compile fails. I checked locally and it works.

mattwigway:
I'm in favor of removing joda-time; even the packages authors recommend moving to `java.time` in Java 8+:

> Joda-Time is the de facto standard date and time library for Java. From Java SE 8 onwards, users are asked to migrate to java.time (JSR-310).

The only problem I see with this is that we still have some code depending on GTFS-lib that hasn't been updated to use Joda. I think the way forward is to cut a release of gtfs-lib before migrating, so that all legacy codebases don't have to be migrated at once. In particular, gtfs-editor doesn't work with Java 8.

abyrd:
Yep, I totally agree. Let's cut a gtfs-lib release then apply the patch. I should be able to take care of this this weekend or Monday.

PR #92, opened 2016-03-07
Remove time from State

buma:
Because duration now exists in State current time is redundant so it is
removed.


buma:
Any objections to that or can I merge it?

abyrd:
There is no good reason to have redundant data in State, especially since we create so many of them. So it's just a choice between elapsed time and clock time. If you mainly need elapsed time, then I'm totally fine with this change.

buma:
The only place current time is needed is if we have time based turn restrictions (Which I think we decided not to have) and time dependent speeds (AKA based on current traffic).

abyrd:
Right, we've decided not to handle those things for now.

PR #111, opened 2016-04-05
McRaptorSuboptimal uses service times when routing

### LINKED ISSUE #110: McRaptorSuboptimalPathProfileRouter doesn't check if trip even runs on wanted date
@mattwigway noticed that R5 uses weekend service on weekdays in #107. This was fixed. But McRaptorSuboptimalPathProfileRouter doesn't check anywhere if tripPattern or trip even runs on specified dates. I noticed now when I have more debug output and I also see services of chosen trips. Sometimes in multi transit trip you get first transit trip on service which runs during a week and second part of service which runs on weekend.

buma:
So that it finds only trips that actually run on wanted day.

This also speeds up McRaptor from 100 ms to 25 ms in my small graph.

Fixes #110

### DIFF
--- a/src/main/java/com/conveyal/r5/profile/McRaptorSuboptimalPathProfileRouter.java
+++ b/src/main/java/com/conveyal/r5/profile/McRaptorSuboptimalPathProfileRouter.java
@@ -67,6 +67,7 @@ public class McRaptorSuboptimalPathProfileRouter {
     private BitSet touchedStops;
     private BitSet touchedPatterns;
     private BitSet patternsNearDestination;
+    private BitSet servicesActive;
 
     public McRaptorSuboptimalPathProfileRouter (TransportNetwork network, ProfileRequest req, TIntIntMap accessTimes, TIntIntMap egressTimes) {
         this.network = network;
@@ -76,6 +77,7 @@ public class McRaptorSuboptimalPathProfileRouter {
         this.touchedStops = new BitSet(network.transitLayer.getStopCount());
         this.touchedPatterns = new BitSet(network.transitLayer.tripPatterns.size());
         this.patternsNearDestination = new BitSet(network.transitLayer.tripPatterns.size());
+        this.servicesActive = network.transitLayer.getActiveServicesForDate(req.date);
     }
 
     /** Get a McRAPTOR state bag for every departure minute */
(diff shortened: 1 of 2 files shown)


src/main/java/com/conveyal/r5/profile/McRaptorSuboptimalPathProfileRouter.java
```diff
@@ -235,6 +245,10 @@ private boolean doOneRound () {
 
                         for (TripSchedule tripSchedule : pattern.tripSchedules) {
                             currentTrip++;
+                            //Skips trips which don't run on wanted date
```
buma: I'm not sure here if currentTrip needs to be incremented even if it is skipped. @mattwigway ?

mattwigway: Yes, we do need to increment currentTrip here.

src/main/java/com/conveyal/r5/profile/McRaptorSuboptimalPathProfileRouter.java
```diff
@@ -207,6 +213,10 @@ private boolean doOneRound () {
                 for (Map.Entry<StatePatternKey, McRaptorState> e : statesPerPatternSequence.entrySet()) {
                     int trip = tripsPerPatternSequence.get(e.getKey());
                     TripSchedule sched = pattern.tripSchedules.get(trip);
+                    //Skips trips which don't run on wanted date
+                    if(!servicesActive.get(sched.serviceCode)) {
+                        continue;
+                    }
```
mattwigway: We don't need this check since you've added a check before we put the trips into tripPerPatternSequence.

buma: Fixed. But can happen that statesPerPatternSequence have trip index as key that isn't inside tripPerPatternSequence?

mattwigway: I don't think so, they're both populated here: https://github.com/conveyal/r5/blob/mcRaptorDate/src/main/java/com/conveyal/r5/profile/McRaptorSuboptimalPathProfileRouter.java#L255

mattwigway:
Looks good to me. Merging.

Issue #109, opened 2016-03-31
Wheelchair support


abyrd:
Today @buma raised the issue of wheelchair support in R5.Every TripPattern already has a BitSet wheelchairAccessible (one bit per stop) and TripSchedules have a flag to show if that trip is accessible. In GTFS wheelchair accessibility is a boolean, per stop and per trip. There is no more granular information than this. We can think of cases where more granular information (stop_time level) might be necessary (high/low floor trams on high/low platforms) but GTFS cannot express this. Even if it could things would get awkward very fast: different trips on the same pattern should not have different handicap accessibility at different stops. If they did, we'd need to split them into different patterns. As it is, entire stops and trips can be skipped as if they didn't exist when a user requests a wheelchair trip.

So, we should first move the per-stop bitset out of the TripPattern and up to the top level, since this boolean is attached to the stops themselves and does not change from one pattern to the other. Then both the per-stop booleans and per-trip booleans should be taken into account when routing.

buma:
I'm not sure why build fails since I didn't change anything in tests that fails and It runs OK locally even with all of other changes stashed.

abyrd:
@buma the travis CI log says:

```
Tests in error: 
  testLoadShapefile(com.conveyal.r5.analyst.FreeFormPointSetTest): No code "EPSG:4326" from authority "EPSG" found for object of type "EngineeringCRS".
```

It seems like it is somehow not getting the Geotools projection database. http://docs.geotools.org/latest/userguide/library/referencing/hsql.html

buma:
Seems so. I was more concerned about a CSV problem which is previous error, but that seems to be the intended output.

mattwigway:
I fixed the build, it just required clearing the travis ci cache (under more options -> caches, delete the cache for the branch in question). This forces a redownload of all maven artifacts, etc. So @buma definitely not your fault.

PR #127, opened 2016-04-15
Move wheelchair filtering into raptor algoritm

buma:
So that filtering is done during the search itself instead at the end.
It skips stops and trips which don't have wheelchair support.
It also skips patterns that don't have transitModes that are requested.

Working on #109.

It seems to be working OK. Some different trips are found then when afterwards filtering is used. But it takes a little more time in raptor.

I added wheelchair checks on the same places that service checks were made. And I added one on stops.
@mattwigway: Could you please check if it is OK?

Then I'll remove afterwards filtering and it can be merged.


mattwigway:
Looks good to me.

PR #128, opened 2016-04-19
Fix for negative edges for edge graph

buma:
Should fix bug which appeared in
70fc77ccbe1f3136410f43e95bbafbc918555b62 and breaks P+R and bike
sharing since edges have negative ID.

It changes negative edge IDs to -1. P+R works again bike sharing doesn't (It finds very short bike part couple of meters and I have to check if this is the reason since distance was added to state in the meantime and other things)

Andrew and Matt had an idea that this could be solved that way:
>I think @mattwigway and I came to a conclusion on that question: it is algorithmically unnecessary to put these starting vertices “at a place” in the map, so they don’t need unique identifiers at all. We can just use -1 in all of them and not put them into the map.

### DIFF
--- a/src/main/java/com/conveyal/r5/streets/StreetRouter.java
+++ b/src/main/java/com/conveyal/r5/streets/StreetRouter.java
@@ -228,8 +228,8 @@ public class StreetRouter {
         queue.clear();
         // from vertex is at end of back edge. Set edge correctly so that turn restrictions/costs are applied correctly
         // at the origin.
-        State startState0 = new State(split.vertex0, split.edge + 1, streetMode);
-        State startState1 = new State(split.vertex1, split.edge, streetMode);
+        State startState0 = new State(split.vertex0, -1, streetMode);
+        State startState1 = new State(split.vertex1, -1, streetMode);
         // TODO walk speed, assuming 1 m/sec currently.
         startState0.weight = split.distance0_mm / 1000;
         startState1.weight = split.distance1_mm / 1000;
@@ -261,9 +261,7 @@ public class StreetRouter {
         queue.clear();
 
         bikeStations.forEachEntry((vertexIdx, bikeStationState) -> {
-            // backEdge needs to be unique for each start state or they will wind up dominating each other.
-            // subtract 1 from -vertexIdx because -0 == 0
-            State state = new State(vertexIdx, -vertexIdx - 1, streetMode);
+            State state = new State(vertexIdx, - 1, streetMode);
             state.weight = bikeStationState.weight+switchCost;
             state.durationSeconds = bikeStationState.durationSeconds+switchTime;
             state.isBikeShare = true;


abyrd:
@mattwigway aren't other changes necessary elsewhere to not insert these states in the map? With them all being "at" -1, dominance conditions would currently be applied to them which isn't what we want.

buma:
Interestingly bike shares have such short paths since same bikeshare is found on walk search from start to bike shares and in bicycle search from bike shares to bikeshares.

buma:
If instead of this fix use fix from ParkRideFix https://github.com/conveyal/r5/commit/c0907daae49ffb32c68d96381018dc64fbac31d4 both P+R and bikeSharing work correctly.

src/main/java/com/conveyal/r5/streets/StreetRouter.java
```diff
@@ -228,8 +228,8 @@ public boolean setOrigin (double lat, double lon) {
         queue.clear();
         // from vertex is at end of back edge. Set edge correctly so that turn restrictions/costs are applied correctly
         // at the origin.
-        State startState0 = new State(split.vertex0, split.edge + 1, streetMode);
-        State startState1 = new State(split.vertex1, split.edge, streetMode);
+        State startState0 = new State(split.vertex0, -1, streetMode);
+        State startState1 = new State(split.vertex1, -1, streetMode);
```
mattwigway: I'm not sure what the problem here was. I think we do want to set the backEdge on these states to be split.edge, so that turn restrictions are applied when leaving the edge that the origin was snapped to. Is there a reason this behavior was problematic?

src/main/java/com/conveyal/r5/streets/StreetRouter.java
```diff
@@ -261,9 +261,7 @@ public void setOrigin(TIntObjectMap<State> bikeStations, int switchTime, int swi
         queue.clear();
 
         bikeStations.forEachEntry((vertexIdx, bikeStationState) -> {
-            // backEdge needs to be unique for each start state or they will wind up dominating each other.
-            // subtract 1 from -vertexIdx because -0 == 0
-            State state = new State(vertexIdx, -vertexIdx - 1, streetMode);
+            State state = new State(vertexIdx, - 1, streetMode);
```
mattwigway: Don't bike stations each have a unique link edge? Can't we just set the backEdge to that?

src/main/java/com/conveyal/r5/streets/StreetRouter.java
buma: Problem was that when states were returned from StreetRouter they had negative edges ID with ID different from -1. And this broke all of getEdge function since negative edges weren't found in EdgeStore. But If I changed StreetSegment class (which reads states and gets edge geometry to create turn instruction) to skip all negative edges I get path with holes some edges get skipped.

I'm not sure if edges with negative IDs are even supposed to be returned from StreetRouter.

So the only solution that worked for both P+R and bike share was this: https://github.com/conveyal/r5/commit/c0907daae49ffb32c68d96381018dc64fbac31d4

src/main/java/com/conveyal/r5/streets/StreetRouter.java
buma: Probably. I just change the values to be the same as before the change that broke bike sharing but since this change also doesn't fix it it's probably something else.

buma:
I'm preparing correct fix.

PR #151, opened 2016-05-04
Should fix problem with negative edges

buma:
Which appeared in 70fc77c. P+R and bike sharing is working again.

Main source of troubles was in state creation from old state. Since first state had negative backEdges  (from setOrigins etc.) next state had those state as backState and there were troubles during path construction in response. Now state creation has backState only if those backState has positive backEdge.

And inside traverse a check needed to be changed from backEdge != -1 to backEdge >= 0 because otherwise we would get exception since first state can have negative edges.

With this changes there is no more negative edges in results.

There are lines [400](https://github.com/conveyal/r5/blob/master/src/main/java/com/conveyal/r5/streets/StreetRouter.java#L400) and [411](https://github.com/conveyal/r5/blob/master/src/main/java/com/conveyal/r5/streets/StreetRouter.java#L411) in StreetRouter where states with negative edges are still saved inside bestStatesAtEdge but since they are never read from it I didn't add conditional before insertion.

@mattwigway: I think this should be the correct fix.

### DIFF
--- a/src/main/java/com/conveyal/r5/api/util/StreetSegment.java
+++ b/src/main/java/com/conveyal/r5/api/util/StreetSegment.java
@@ -73,7 +73,7 @@ public class StreetSegment {
         double lastAngle = 0;
         for (StreetRouter.State state: path.getStates()) {
             int edgeIdx = state.backEdge;
-            if (edgeIdx != -1) {
+            if (edgeIdx >= 0) {
                 EdgeStore.Edge edge = path.getEdge(edgeIdx);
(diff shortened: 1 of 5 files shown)


buma:
The problem with missing car part in P+R Trips was in my code since Car part wasn't found because state with correct vertex was skipped since it had negative edge. Since last state of walk part needs to be the same vertex as start of CAR part so that path is found in different street routers. If instead of skipping backStates with negative edges I change them to -1. Everything works. 

And since currently no negative edges are inserted in bestStatesAtEdge this probably doesn't break anything.

PR #193, opened 2016-08-24
Saving stopTrees after building

buma:
This change save stopTrees after building so that they aren't rebuild everything graph is opened. Stoptrees also aren't build in point to point routing since they aren't needed there.


mattwigway:
We now have a similar feature in master.

PR #197, opened 2016-09-26
Split can now split backward edge

### LINKED ISSUE #196: Split only splits forward edges
Splitting only forward edges usually works. But It can have weird consequences if backward edges have more permissions then forward edges. For example when tagging `oneway=-1` backward edge has car permissions and forward edge doesn't. This can also happen if road is tagged with `bicycle:forward=use_sidepath/no` or any other tagging where backward edge has more permissions then forward edge. Split doesn't find this edge since it doesn't check backward edges. (Since only forward edges are in  geometry index).
For example how this looks like when oneway is tagged as `oneway=-1`:
image

I have a fix for this which also checks backward edge if forward doesn't have existing permissions. But I'm not sure if finding of splits would break if non forward edges would be returned.

buma:
if there is no required permissions on forward edge.

Fixes #196

### DIFF
--- a/src/main/java/com/conveyal/r5/streets/Split.java
+++ b/src/main/java/com/conveyal/r5/streets/Split.java
@@ -79,10 +79,36 @@ public class Split {
             //and route is never found since point is inaccessible because edges leading to it don't have required permission
             if (edge.getFlag(EdgeStore.EdgeFlag.LINK)) return true;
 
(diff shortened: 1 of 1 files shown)


abyrd:
I re-implemented this in 2e3517b to fit with some other recent changes.

PR #202, opened 2016-11-01
Scenario editor accessibility

mattwigway:
Add methods to compute regional accessibility in scenario editor.

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/GridCache.java
+++ b/src/main/java/com/conveyal/r5/analyst/GridCache.java
@@ -1,0 +1,64 @@
+package com.conveyal.r5.analyst;
+
+import com.amazonaws.AmazonServiceException;
+import com.amazonaws.ClientConfiguration;
+import com.amazonaws.services.s3.AmazonS3;
+import com.amazonaws.services.s3.AmazonS3Client;
+import com.amazonaws.services.s3.model.S3Object;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.BufferedInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.concurrent.ExecutionException;
+import java.util.zip.GZIPInputStream;
+
+/**
+ * Cache grids.
+ */
+public class GridCache {
+    private static final Logger LOG = LoggerFactory.getLogger(GridCache.class);
+
+    private AmazonS3 s3 = new AmazonS3Client();
+
+    /** How large the cache should be. Should be large enough to fit all field of a project */
+    private static final int CACHE_SIZE = 200;
+
+    private LoadingCache<String, Grid> cache = CacheBuilder.newBuilder()
+                .maximumSize(CACHE_SIZE)
+                // lambdas not legal here for whatever reason
+                .build(new CacheLoader<String, Grid>() {
+                    @Override
+                    public Grid load(String s) throws Exception {
+                        return loadGrid(s);
+                    }
+                });
+
+    private final String bucket;
+
+    public GridCache (String bucket) {
+        this.bucket = bucket;
+    }
+
+    private Grid loadGrid (String key) throws IOException {
+        S3Object obj = s3.getObject(bucket, key);
+        // no need to check if it exists; if it doesn't getObject will throw an exception which will be caught in the
+        // get function below
+        // Grids are gzipped on S3
+        InputStream is = new GZIPInputStream(new BufferedInputStream(obj.getObjectContent()));
+        return Grid.read(is);
+    }
+
+    public Grid get (String key) {
+        try {
+            return cache.get(key);
+        } catch (ExecutionException e) {
+            LOG.error("Error retrieving grid {}", key, e);
+            throw new RuntimeException(e);
+        }
+    }
+}
--- a/src/main/java/com/conveyal/r5/analyst/GridCache.java
+++ b/src/main/java/com/conveyal/r5/analyst/GridCache.java
@@ -18,7 +18,8 @@ import java.util.concurrent.ExecutionException;
 import java.util.zip.GZIPInputStream;
 
 /**
- * Cache grids.
+ * Cache opportunity density grids on S3, with a local cache for performance when reusing the same grids. Each instance
+ * works with a single S3 bucket.
  */
 public class GridCache {
     private static final Logger LOG = LoggerFactory.getLogger(GridCache.class);
--- a/src/main/java/com/conveyal/r5/analyst/GridComputer.java
+++ b/src/main/java/com/conveyal/r5/analyst/GridComputer.java
@@ -1,0 +1,154 @@
+package com.conveyal.r5.analyst;
+
+import com.amazonaws.services.s3.model.ObjectMetadata;
+import com.amazonaws.services.sqs.AmazonSQS;
+import com.amazonaws.services.sqs.AmazonSQSAsync;
+import com.amazonaws.services.sqs.AmazonSQSAsyncClient;
+import com.amazonaws.services.sqs.AmazonSQSClient;
+import com.amazonaws.services.sqs.model.MessageAttributeValue;
+import com.amazonaws.services.sqs.model.SendMessageRequest;
+import com.conveyal.r5.analyst.cluster.GridRequest;
+import com.conveyal.r5.analyst.cluster.TaskStatistics;
+import com.conveyal.r5.analyst.messages.OriginOuterClass;
+import com.conveyal.r5.api.util.LegMode;
+import com.conveyal.r5.profile.RaptorState;
+import com.conveyal.r5.profile.RaptorWorker;
+import com.conveyal.r5.profile.RepeatedRaptorProfileRouter;
+import com.conveyal.r5.profile.StreetMode;
+import com.conveyal.r5.streets.LinkedPointSet;
+import com.conveyal.r5.streets.PointSetTimes;
+import com.conveyal.r5.streets.StreetRouter;
+import com.conveyal.r5.transit.TransportNetwork;
+import com.conveyal.r5.transit.TransportNetworkCache;
+import com.conveyal.r5.util.S3Util;
+import com.google.common.io.LittleEndianDataOutputStream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.BufferedInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.PipedInputStream;
+import java.io.PipedOutputStream;
+import java.util.Base64;
+import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+import java.util.zip.GZIPOutputStream;
+
+/**
+ * Computes accessibility to grids and dumps it to s3.
+ */
+public class GridComputer  {
+    private static final Logger LOG = LoggerFactory.getLogger(GridComputer.class);
+
+    /** SQS client. TODO: async? */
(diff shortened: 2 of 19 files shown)


mattwigway:
@abyrd can you review when you have a chance?

src/main/java/com/conveyal/r5/analyst/cluster/GridResultAssembler.java
```diff
@@ -0,0 +1,201 @@
...
+                error = true;
+                LOG.error("Origin version mismatch for query %s, expected {}, found {}", ORIGIN_VERSION, version);
+            }
+
+            int x = data.readInt();
+            int y = data.readInt();
+            int iterations = data.readInt();
+
+            // convert the remainder to a delta coded byte array
+            ByteArrayOutputStream originData = new ByteArrayOutputStream(iterations * 4);
```
abyrd: Variable names are slightly confusing, impeding reading the code. Maybe use originByteOutputStream, which is wrapped by originDataOutputStream.

src/main/java/com/conveyal/r5/analyst/cluster/GridResultAssembler.java
```diff
@@ -0,0 +1,201 @@
...
+        } catch (Exception e) {
+            LOG.error("Error uploading results of regional analysis {}", request.jobId, e);
+        }
+    }
+
+    /** Process an SQS message */
+    public void handleMessage (Message message) {
+        try {
+            byte[] body = base64.decode(message.getBody());
+            ByteArrayInputStream bais = new ByteArrayInputStream(body);
```
abyrd: These variables need a naming scheme that is parallel with the origin output streams below, e.g. messageByteInputStream and messageDataInputStream. However even more advantageous would be to abstract the origin grids out into a separate class, which buffers the data for a single origin, runs header and sanity checks for that data, and has a method for writing out the delta-coded version to an arbitrary DataOutputStream. Or can write to an arbitrary region of a seekable buffer.

src/main/java/com/conveyal/r5/analyst/cluster/GridResultAssembler.java
```diff
@@ -0,0 +1,201 @@
...
+            ByteArrayOutputStream originData = new ByteArrayOutputStream(iterations * 4);
+            LittleEndianDataOutputStream origin = new LittleEndianDataOutputStream(originData);
+            for (int i = 0, prev = 0; i < iterations; i++) {
+                int current = data.readInt();
+                origin.writeInt(current - prev);
+                prev = current;
+            }
+            origin.close();
+
+            // write to the buffer in a synchronized block to ensure no threading issues
```
abyrd: "Write to the proper subregion of the unified results buffer for this origin number."

src/main/java/com/conveyal/r5/analyst/cluster/GridResultAssembler.java
```diff
@@ -0,0 +1,201 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.RandomAccessFile;
+import java.util.Base64;
+import java.util.zip.GZIPOutputStream;
+
+/**
+ * Assemble the results of GridComputer into AccessGrids.
+ *
```
abyrd: Rather than just naming the input and output classes (which are visible from function signatures), we should describe the data transform that occurs, and when it happens. "During distributed computation of access to gridded destinations, workers place raw accessibility results for single origins onto an Amazon SQS queue. These results contain one accessibility measurement per iteration (departure minutes and Monte Carlo draws) per origin grid cell. This class pulls results off the queue as they become available, and assembles them into a single large file containing a delta-coded version of the same data for all origin points."

AccessGrids should probably be abstracted out into another class as well, with OO-style "message" methods for reading and writing results for a given origin.

src/main/java/com/conveyal/r5/analyst/cluster/GridResultConsumer.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.stream.Collectors;
+
+/**
+ * Listens to an SQS queue and manages grid results coming back.
```
abyrd: Comment should emphasize the fact that this is a long-lived independent thread that waits for an entire regional analysis to finish.

src/main/java/com/conveyal/r5/analyst/scenario/AndrewOwenMeanGridSampler.java
```diff
@@ -0,0 +1,85 @@
...
+ * distribution of accessibility is well defined. The drawback of this method is that it treats opportunities as
+ * completely fungible, even day-to-day. If there are 50,000 jobs accessibility from 8:00 to 8:30 by taking an eastbound
+ * commuter train, and a different 50,000 accessible from 8:30 to 9:00 by taking a westbound commuter train, this measure
+ * will average out to 50,000, even though none of those 50,000 jobs are accessible for the whole time window, essentially
+ * assuming that people will choose their job based on what time they leave for work on a given day. For more on this issue,
+ * see Conway, Matthew Wigginton, Andrew Byrd, and Marco van der Linden. n.d. “Evidence-Based Transit and Land Use Sketch
+ * Planning Using Interactive Accessibility Methods on Combined Schedule and Headway-Based Networks,” under review for the
+ * the 96th Annual Meeting of the Transportation Research Board, January 2017.
+ */
+public class AndrewOwenMeanGridSampler {
```
abyrd: The name sort of implies that there are multiple GridSamplers behaving in different ways, but there is no GridSampler interface. I think all this new code would benefit from more abstraction and type definition. Of course it's understandable that that doesn't happen when prototyping, but as soon as we're sure what our basic work units are (AccessGrids etc.) we should give them existence as declared types.

src/main/java/com/conveyal/r5/analyst/scenario/AndrewOwenMeanGridSampler.java
```diff
@@ -0,0 +1,85 @@
...
+            header[i] = (char) input.readByte();
+        }
+
+        if (!"ACCESSGR".equals(new String(header))) {
+            throw new IllegalArgumentException("Input not in access grid format!");
+        }
+
+        int version = input.readInt();
+
+        if (version != ACCESS_GRID_VERSION) {
```
abyrd: Repeated code for reading and sanity checking "objects" that have no type definition.

src/main/java/com/conveyal/r5/transit/TransportNetworkCache.java
```diff
@@ -121,16 +121,29 @@ public synchronized TransportNetwork getNetworkForScenario (String networkId, Pr
             Scenario scenario;
             if (request.scenario == null && request.scenarioId != null) {
                 // resolve scenario
-                LOG.info("Retrieving scenario from S3");
-                // TODO offline mode
-                S3Object obj = s3.getObject(sourceBucket, String.format("%s_%s.json", networkId, scenarioId));
-                InputStream is = obj.getObjectContent();
+                LOG.info("Retrieving scenario stored separately");
```
abyrd: Not clear what "separately" means. If I understand correctly this is a multilevel cache where cache misses on the local filesystem fall through to S3.

src/main/java/com/conveyal/r5/analyst/ImprovementProbabilityGridSampler.java
```diff
@@ -0,0 +1,108 @@
...
+import com.amazonaws.services.s3.AmazonS3;
+import com.amazonaws.services.s3.AmazonS3Client;
+import com.amazonaws.services.s3.model.S3Object;
+import com.google.common.io.LittleEndianDataInputStream;
+
+import java.io.IOException;
+import java.util.zip.GZIPInputStream;
+
+/**
+ * Sample from an access grid of instantaneous accessibility to compute the probability of improvement.
```
mattwigway: Hmm, the grid of distributions is an interesting thought. Then you could say "what's the probability of an improvement greater than X"

src/main/java/com/conveyal/r5/analyst/GridCache.java
```diff
@@ -0,0 +1,64 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.io.BufferedInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.concurrent.ExecutionException;
+import java.util.zip.GZIPInputStream;
+
+/**
+ * Cache grids.
```
mattwigway: Yeah, that one gets the award for least informative comment of the year.

src/main/java/com/conveyal/r5/analyst/GridComputer.java
```diff
@@ -0,0 +1,149 @@
...
+        int millimetersPerSecond = (int) (request.request.walkSpeed * 1000);
+
+        for (TIntIntIterator it = reachedStops.iterator(); it.hasNext();) {
+            it.advance();
+            it.setValue(it.value() / millimetersPerSecond);
+        }
+
+        router.runRaptor(reachedStops, linkedTargets.eval(sr::getTravelTimeToVertex), new TaskStatistics());
+
+        int[] accessibilityPerIteration = new int[router.includeInAverages.cardinality()];
```
mattwigway: Which begs the question of why we are calculating those bounds at all, we could make the algorithm faster if we didn't.

src/main/java/com/conveyal/r5/analyst/PercentileGridSampler.java
```diff
@@ -0,0 +1,83 @@
...
+        for (int y = 0; y < height; y++) {
+            for (int x = 0; x < width; x++) {
+                int[] values = new int[nIterations];
+
+                for (int iteration = 0, val = 0; iteration < nIterations; iteration++) {
+                    values[iteration] = (val += input.readInt());
+                }
+
+                // compute percentiles
+                Arrays.sort(values);
```
mattwigway: Abstracting it out would make it less efficient as making a properly modular method would require copying the array before sorting it.

mattwigway: The Apache commons lib uses double[] and this is using int[], which appear to not be castable to each other.

src/main/java/com/conveyal/r5/analyst/cluster/GridResultAssembler.java
mattwigway: The thing with abstracting out access grids is that for reading it would preclude streaming access.

src/main/java/com/conveyal/r5/analyst/scenario/AndrewOwenMeanGridSampler.java
mattwigway: There are, but they don't have the same signature (i.e. for a percentile sampler you need to to know what percentile, for a mean sampler there are no parameters, etc.). We could make these into subclasses/interfaces that have a constructor where you set the parameter and then you call a run/compute method, but that seems needlessly verbose when these are basically classes wrapping a single function.

mattwigway:
@abyrd I made some changes to address most of your concerns. The main one I haven't addressed is abstracting out the access grid reading code; I'm not quite sure the best way to handle that since we need to read in a streaming fashion given the sizes of the files. A RandomAccessFile-based solution won't work as we store the files gzipped because they're quite large.

src/main/java/com/conveyal/r5/analyst/ImprovementProbabilityGridSampler.java
abyrd: I'll make an Asana ticket for this, for future consideration.

src/main/java/com/conveyal/r5/analyst/GridComputer.java
abyrd: It's seeming less and less necessary, and does add complexity. I'll also make an Asana ticket for this idea.

PR #204, opened 2016-11-21
Refactor transitive network construction

abyrd:
I suspected the problem with zoomed-in transitive rendering in TAUI was bad geometries where shapes aren't present, as in Marseille. I did a bunch of refactoring to break the TransitiveNetwork down into functional units, and eventually noticed that we were using stop IDs as vertex IDs without the intermediate lookup.

Although the rendering is much better than it was (not jumping around to random street vertices) for some reason the results are still bizarre. The weirdness seems more common and visible on inter-city trains. Things render properly in schematic mode but the destination station seems to be shifted one stop early (while retaining the true destination stop's name) in geo-accurate mode. It seems plausible that I've introduced some off-by-one error when generating the geometries but I don't see it.

### DIFF
--- a/src/main/java/com/conveyal/r5/transit/TripPattern.java
+++ b/src/main/java/com/conveyal/r5/transit/TripPattern.java
@@ -219,6 +219,9 @@ public class TripPattern implements Serializable, Cloneable {
                 previousCoordinate = currentCoordinate;
             }
         }
+        if (geometries.size() != stops.length - 1) {
+            throw new AssertionError("This function should always generate one less geometry than there are stops.");
+        }
         return geometries;
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/conveyal/r5/transit/TripPattern.java
```diff
@@ -56,7 +60,10 @@
     // This set includes the numeric codes for all services on which at least one trip in this pattern is active.
     public BitSet servicesActive = new BitSet();
 
-    /** index of this route in TransitLayer data. -1 if detailed route information has not been loaded */
+    /**
+     * index of this route in TransitLayer data. -1 if detailed route information has not been loaded
+     * TODO clarify what "this route" means. The route of this tripPattern?
```
mattwigway: Yes, it's the route associated with the TripPattern.

PR #207, opened 2017-01-05
Scenario editor

mattwigway:
I think this is ready to go.


src/main/java/com/conveyal/r5/analyst/Grid.java
```diff
@@ -335,4 +359,139 @@ public static Polygon getPixelGeometry (int x, int y, int zoom) {
...
+
+            for (String field : numericColumns) {
+                String value = reader.get(field);
+
+                double val;
+
+                if (value == null || "".equals(value)) {
+                    val = 0;
+                } else {
+                    val = parseDouble(value);
```
abyrd: If some column contains the occasional non-numeric junk, that columns will be deemed non-numeric, and won't be read in this second pass (though that prevents things from blowing up here). Maybe we should detect that case and throw a descriptive error that can reach the end user.

src/main/java/com/conveyal/r5/analyst/Grid.java
```diff
@@ -154,9 +159,27 @@ public void rasterize (Geometry geometry, double value) {
...
+     * all the web Mercator pixels that intersect it.
+     */
+    public void rasterize (Geometry geometry, double value) {
+        incrementFromPixelWeights(getPixelWeights(geometry), value);
+    }
+
+    public void incrementFromPixelWeights (TObjectDoubleMap<int[]> weights, double value) {
+        for (TObjectDoubleIterator<int[]> it = weights.iterator(); it.hasNext();) {
+            it.advance();
+            grid[it.key()[0]][it.key()[1]] += it.value() * value;
```
abyrd: Wow, this really makes me wish Java had built in tuple types. I guess as long as there's no major performance hit, it's sane to use a nice structured type like a Map from int[2] to double to pass around the pixel weights. Java seems pretty good at creating and destroying a lot of objects quickly. But `grid[it.key()[0]][it.key()[1]]` is still not really that readable. What about a list of (x, y, weight)? It requires a new class definition but might be clearer what's going on.

src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
```diff
@@ -759,6 +780,27 @@ public void finishPriorityTask(GenericClusterRequest clusterRequest, InputStream
     }
 
     /**
+     * Report to the broker that the task taskId could not be processed due to errors.
+     * The broker should then pass the errors back up to the client that enqueued that task.
+     * That objects are always the same type (TaskError) so the client knows what to expect.
+     */
+    public void reportTaskErrors(int taskId, List<TaskError> taskErrors) {
```
abyrd: I should add a status code parameter to this. The scenario application errors are usually the result of a bad request, but we're not sure randomly caught exceptions deserve a 400, and we may use this function elsewhere.

src/main/java/com/conveyal/r5/analyst/error/ScenarioApplicationException.java
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Created by abyrd on 2016-12-31
+ */
+public class ScenarioApplicationException extends RuntimeException {
+
+    // The structured error reports that can be sent back to the client via the broker.
+    public final List<TaskError> taskErrors = new ArrayList<>();
+
+    /**
+     * Pass in all the modifications that
```
abyrd: My comment needs fixing.

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
```diff
@@ -158,25 +160,33 @@ public LinkedPointSet(int[] edges, int[] distances0_mm, int[] distances1_mm, Poi
...
-                edges[p] = split.edge;
-                distances0_mm[p] = split.distance0_mm;
-                distances1_mm[p] = split.distance1_mm;
+            // When working with a scenario, skip all points that are not linked to a deleted street (i.e. one that has
+            // been split). At the current time, the only street network modification we support is splitting existing streets,
+            // so the only way a point can need to be relinked is if it is connected to a street which was split (and therefore deleted).
+            // FIXME when we permit street network modifications beyond adding transit stops we will need to change how this works,
+            // we may be able to use some type of flood-fill algorithm in geographic space, expanding the relink envelope until we
+            // hit edges on all sides or reach some predefined maximum.
+            if (all || streetLayer.edgeStore.temporarilyDeletedEdges != null && streetLayer.edgeStore.temporarilyDeletedEdges.contains(edges[p])) {
```
abyrd: I would lean toward grouping the sub-expression after || with parentheses rather than relying on the reader's operator precedence knowledge.

src/main/java/com/conveyal/r5/profile/RaptorWorker.java
```diff
@@ -224,7 +222,7 @@ public PropagatedTimesStore runRaptor (TIntIntMap accessTimes, PointSetTimes non
         // Now multiply the number of departure minutes by the number of Monte Carlo frequency draws per minute.
         if (data.hasFrequencies) {
             // We add 2 because we do two additional iterations for zero and maximal boarding times (not Monte Carlo draws).
-            iterations *= (monteCarloDraws + 2);
+            iterations *= (monteCarloDraws);
```
abyrd: Oh wait, that comment should be removed though, and a current explanation added

abyrd:
Shall we go ahead and merge this @mattwigway? If we keep feature branches open too long it tends to make development more complex to manage.

src/main/java/com/conveyal/r5/transit/TransportNetwork.java
```diff
@@ -137,6 +137,7 @@ private static TransportNetwork fromFiles (String osmSourceFile, List<String> gt
         // transitLayer.summarizeRoutesAndPatterns();
 
         // The street index is needed for associating transit stops with the street network.
+        // TODO why is this called twice?
```
abyrd: Added comment in code.

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
```diff
@@ -140,6 +140,22 @@ public LinkedPointSet(PointSet pointSet, StreetLayer streetLayer, StreetMode str
 
     }
 
+    /** used to create new linked pointsets that are subsets of existing ones */
+    public LinkedPointSet(int[] edges, int[] distances0_mm, int[] distances1_mm, PointSet pointSet,
```
abyrd: Resolved in 777dab2.

src/main/java/com/conveyal/r5/analyst/Grid.java
```diff
@@ -335,4 +359,139 @@ public static Polygon getPixelGeometry (int x, int y, int zoom) {
...
+
+    /** Create grids from a CSV file */
+    public static Map<String,Grid> fromCsv(File csvFile, String latField, String lonField, int zoom) throws IOException {
+        CsvReader reader = new CsvReader(new BufferedInputStream(new FileInputStream(csvFile)), Charset.forName("UTF-8"));
+        reader.readHeaders();
+
+        String[] headers = reader.getHeaders();
+        if (!Stream.of(headers).filter(h -> h.equals(latField)).findAny().isPresent()) {
+            LOG.info("Lat field not found!");
+            halt(400, "Lat field not found");
```
abyrd: Added 500 status codes in ececb79

src/main/java/com/conveyal/r5/analyst/Grid.java
```diff
@@ -335,4 +359,139 @@ public static Polygon getPixelGeometry (int x, int y, int zoom) {
...
+
+        if (!Stream.of(headers).filter(h -> h.equals(lonField)).findAny().isPresent()) {
+            LOG.info("Lon field not found!");
+            halt(400, "Lon field not found");
+        }
+
+        Envelope envelope = new Envelope();
+
+        // Keep track of which fields contain numeric values
+        Set<String> numericColumns = Stream.of(headers).collect(Collectors.toCollection(HashSet::new));
```
abyrd: Comment added about numeric field detection

src/main/java/com/conveyal/r5/analyst/Grid.java
```diff
@@ -335,4 +359,139 @@ public static Polygon getPixelGeometry (int x, int y, int zoom) {
...
+
+            for (Iterator<String> it = numericColumns.iterator(); it.hasNext();) {
+                String field = it.next();
+                String value = reader.get(field);
+                if (value == null || "".equals(value)) continue; // allow missing data
+                try {
+                    // TODO also exclude columns containing negatives?
+                    parseDouble(value);
+                } catch (NumberFormatException e) {
+                    it.remove();
```
abyrd: Ah, never mind. The columns are removed the first time a non-numeric value is encountered. That means only a few exceptions will occur.

src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
abyrd: 500 error codes added in ececb79

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
abyrd: Fixed in f9f5907

abyrd:
I pushed all the changes / fixes I suggested in past reviews of this PR.

src/main/java/com/conveyal/r5/analyst/PointSet.java
```diff
@@ -33,27 +34,31 @@
...
-                    // Build a new linkage from this PointSet to the supplied StreetNetwork,
-                    // initialized with the existing linkage to the base StreetNetwork when relevant.
-                    return new LinkedPointSet(PointSet.this, key.a, key.b, baseLinkage);
-                }
-            });
-
+            .build(new LinkageCacheLoader());
+
+    // This is pulled out into a named class because serializing the anonymous inner class was getting ugly.
+    private class LinkageCacheLoader extends CacheLoader<Tuple2<StreetLayer, StreetMode>, LinkedPointSet> implements Serializable {
```
mattwigway: Ah I guess we want to save the linked pointsets also.

src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
```diff
@@ -238,10 +232,6 @@ public AnalystWorker(Properties config, TransportNetworkCache cache) {
         // http://docs.aws.amazon.com/AWSSdkDocsJava/latest/DeveloperGuide/java-dg-roles.html
         s3 = new AmazonS3Client();
```
mattwigway: Don't we still need to set the region on this?

src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
```diff
@@ -784,13 +773,12 @@ public void finishPriorityTask(GenericClusterRequest clusterRequest, InputStream
...
-    public void reportTaskErrors(int taskId, List<TaskError> taskErrors) {
-        String url = BROKER_BASE_URL + String.format("/complete/error/%s", taskId);
+    public void reportTaskErrors(int taskId, int httpStatusCode, List<TaskError> taskErrors) {
+        String url = BROKER_BASE_URL + String.format("/complete/%d/%s", httpStatusCode, taskId);
         try {
             HttpPost httpPost = new HttpPost(url);
             httpPost.setHeader("Content-type", "application/json");
-            byte[] errorsAsJson = JsonUtilities.objectMapper.writeValueAsBytes(taskErrors);
-            httpPost.setEntity(new ByteArrayEntity(errorsAsJson));
+            httpPost.setEntity(JsonUtilities.objectToJsonHttpEntity(taskErrors));
```
mattwigway: We need to make some frontend changes here to handle errors that are not 400s.

src/main/java/com/conveyal/r5/transit/TransitLayer.java
```diff
@@ -143,6 +144,16 @@
     /** Map from feed ID to feed CRC32 to ensure that we can't apply scenarios to the wrong feeds */
     public Map<String, Long> feedChecksums = new HashMap<>();
 
+    /**
+     * A string uniquely identifying the contents of this TransitLayer among all TransitLayers.
+     * When no scenario has been applied, this field will contain the ID of the containing TransportNetwork.
+     * When a scenario has modified this StreetLayer, this field will be changed to the scenario's ID.
+     * We need a way to know what transit information is in the layer independent of object identity, which is lost in
+     * a round trip through serialization. This also allows re-using cached information for multiple scenarios that
+     * don't modify the transit network. (This never happens yet, but will when we allow street editing.)
```
mattwigway: Really though, we could preserve linkages any time a scenario creates no new stops, but it's probably not worth the effort.

src/main/java/com/conveyal/r5/transit/TransitLayer.java
```diff
@@ -682,18 +693,23 @@ public static TransitModes getTransitModes(int routeType) {
     /**
      * @return a semi-shallow copy of this transit layer for use when applying scenarios.
      */
-    public TransitLayer scenarioCopy(TransportNetwork newScenarioNetwork) {
+    public TransitLayer scenarioCopy(TransportNetwork newScenarioNetwork, boolean willBeModified) {
```
mattwigway: When would willBeModified be false? When a scenario only affects the street layer?

src/main/java/com/conveyal/r5/transit/TransportNetwork.java
```diff
@@ -37,55 +38,60 @@
...
     /**
-     * Cache a grid point set that covers the extents of this transport network. The PointSet itself caches linkages to
-     * street networks.
-     *
-     * Eventually we should save this point set and linkage but it's a little tricky since there's a linkage cache with
-     * references to streetlayers, a spatial index, etc. in there.
+     * A grid point set that covers the full extent of this transport network. The PointSet itself then caches linkages
+     * to street networks (the baseline street network, or ones with various scenarios applied). If they have been
+     * created, this point set and its linkage to the street network are serialized along with the network, which makes
+     * startup much faster. Note that there's a linkage cache with references to streetlayers in this GridPointSet,
```
mattwigway: It doesn't have a reference to streetlayers, right? I mean, you still don't want to serialize it because it will contain too many linkages, but it won't pull in a bunch of streetlayers.

mattwigway: Didn't we change it to have a reference to scenario IDs?

src/main/java/com/conveyal/r5/transit/TransportNetwork.java
```diff
@@ -285,25 +282,16 @@ public static InputFileType forFile(File file) {
...
-    public WebMercatorGridPointSet getGridPointSet() {
-        if (this.gridPointSet == null) {
-            synchronized (this) {
-                if (this.gridPointSet == null) {
-                    this.gridPointSet = new WebMercatorGridPointSet(this);
-                }
-            }
+    public void rebuildLinkedGridPointSet() {
+        if (gridPointSet == null) {
+            gridPointSet = new WebMercatorGridPointSet(this);
```
mattwigway: Do we need some synchronization here? I'd think this would be called from a bunch of threads simultaneously on worker startup.

src/main/java/com/conveyal/r5/transit/TransportNetwork.java
```diff
@@ -285,25 +282,16 @@ public static InputFileType forFile(File file) {
...
-        return this.gridPointSet;
-    }
-
-    /**
-     * @return an efficient implicit grid PointSet for this TransportNetwork, pre-linked to the street layer.
-     */
-    public LinkedPointSet getLinkedGridPointSet() {
-        // TODO don't hardwire walk mode
-        return getGridPointSet().link(streetLayer, StreetMode.WALK);
+        // Here we are bypassing the GridPointSet's internal cache of linkages.
```
mattwigway: Why are we bypassing the cache?

src/main/java/com/conveyal/r5/transit/TransportNetwork.java
```diff
@@ -345,37 +333,29 @@ public TransportNetwork applyScenario (Scenario scenario) {
...
+     * wasteful to copy. It is duplicated in such a way that it wraps the original lists, allowing them to be
+     * non-destructively extended. There will be some performance hit from wrapping these lists, but it's probably
+     * negligible.
+     *
+     * @return a copy of this TransportNetwork that is partly shallow and partly deep.
      */
     public TransportNetwork scenarioCopy(Scenario scenario) {
+        // Maybe we should be using clone() here but TransportNetwork has very few fields and most are overwritten.
         TransportNetwork copy = new TransportNetwork();
-        copy.networkId = scenario.id;
```
mattwigway: Have we confirmed that networkId is not used anyplace that will cause trouble if we have two networks with the same network ID?

src/main/java/com/conveyal/r5/transit/TransportNetworkCache.java
```diff
@@ -6,27 +6,31 @@
...
 import java.util.Map;
+import java.util.Set;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
 
 /**
  * This holds one or more TransportNetworks keyed on unique strings.
- * This is a replacement for ClusterGraphBuilder.
+ * Actually this currently only holds one single TransportNetwork, but that will eventually change.
  * TODO this should serialize any networks it builds, attempt to reload from disk, and copy serialized networks to S3.
```
mattwigway: This TODO is obsolete, right?

src/main/java/com/conveyal/r5/transit/TransportNetworkCache.java
```diff
@@ -108,15 +106,15 @@ public synchronized TransportNetwork getNetwork (String networkId) {
...
+     * Find or create a TransportNetwork for the given scenario ID.
      * By design a particular scenario is always defined relative to a single base graph (it's never applied to multiple
      * different base graphs). Therefore we can look up cached scenario networks based solely on their scenarioId
      * rather than a compound key of (networkId, scenarioId).
      *
      * The fact that scenario networks are cached means that PointSet linkages will be automatically reused when
      * the scenario is found by its ID and reused.
      *
-     * TODO LinkedPointSets keep a reference back to a StreetLayer which means that the network will not be completely garbage collected upon network switch
+     * TODO why are we passing in the request instead of the Scenario (which should contain its own ID)?
```
mattwigway: I believe because this function handles both scenarios in profile requests as well as scenarios pulled from S3.

src/main/java/com/conveyal/r5/util/ExpandingMMFBytez.java
```diff
@@ -0,0 +1,233 @@
...
+ */
+public class ExpandingMMFBytez extends MallocBytez {
+
+    private static final Method map0;
+    private static final Method unmap0;
+    private static final int MAP_READ_WRITE = 1;
+
+    static {
+        try {
+            map0 = FileChannelImpl.class.getDeclaredMethod("map0", int.class, long.class, long.class);
```
mattwigway: This code is accessing private methods on classes we do not control, are we sure this is wise?

src/main/java/com/conveyal/r5/util/ExpandingMMFBytez.java
```diff
@@ -0,0 +1,233 @@
...
+        T result = (T) in.readObject(TransportNetwork.class);
+        in.close();
+        coderFactory.free();
+        return result;
+    }
+
+    /**
+     * An instance of this object is passed to an FSTConfiguration when creating an FSTObjectInput or FSTObjectOutput.
+     * It provides the functionality of serializing/deserializing to/from a file.
+     * It can handle huge objects up to about 4GB without eating much memory by using the memory mapped file mechanism.
```
mattwigway: Should handle more than 4GB on non-Darwin systems, yeah?

src/main/java/com/conveyal/r5/analyst/PointSet.java
abyrd: Comments updated.

src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
abyrd: I added code to set the region on the s3 object and modified how the default is set, so the default will be set both when running outside EC2 and when running offline (though not sure we'd ever need it when running offline).

src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
abyrd: There is a ticket for that.

src/main/java/com/conveyal/r5/transit/TransitLayer.java
abyrd: This is an optimization and one that may not be worth the complexity given selective relinking.

src/main/java/com/conveyal/r5/transit/TransitLayer.java
abyrd: Yes. I added some comments here. It's questionable whether this optimization actually affects routing speed, but in theory it saves a comparison and an extra dereference every time we use the edge/vertex stores.

src/main/java/com/conveyal/r5/transit/TransportNetwork.java
abyrd: The WebMercatorGridPointset does not itself contain any references to a StreetLayer. Even if it did, it would be a reference to the network's default street layer, which is already being serialized. The serializer would spot the repeat reference and not serialize the streetLayer twice. The Guava loadingCache of linkages will not serialize its contents by design. That's why the TransportNetwork also has a linkedGridPointSet field. This means we're completely bypassing the cache for gridded analysis, but we've already essentially got two different methods, one gridded one not. Come to think of it, it would be more normalized to just store the linked grid pointset, and access the basic unlinked grid pointset via the linked pointset field.

src/main/java/com/conveyal/r5/transit/TransportNetwork.java
abyrd: It is called only in places that should already be synchronized: Scenario#applyToTransportNetwork and TransportNetworkCache#buildNetwork. The grid is usually built before the graph is serialized, and serialized along with the graph.

src/main/java/com/conveyal/r5/transit/TransportNetwork.java
abyrd: Comment added.

src/main/java/com/conveyal/r5/transit/TransportNetwork.java
abyrd: I don't think we'll ever have two network objects with the same ID. This change does not mean we're no longer setting it, the field is just called scenarioId now (here and in the street and transit layers) because it's not always a transit network ID. It's usually a scenario ID, and arguably when a scenario is not applied, the network ID is the ID of the baseline empty scenario.

src/main/java/com/conveyal/r5/transit/TransportNetworkCache.java
abyrd: TODO removed, and javadoc updated.

src/main/java/com/conveyal/r5/transit/TransportNetworkCache.java
abyrd: Ah, so the full content of the scenario may be embedded in the profile request, or we may only have an ID requiring download from S3. Updating the comment.

src/main/java/com/conveyal/r5/util/ExpandingMMFBytez.java
abyrd: It's not, but it's the only way to get a long-addressable buffer. The real solution is streaming serialization in FST, for which there is a ticket, but it's not implemented yet. Added comments to the code.

abyrd: I've got a pure safe java implementation on the back burner that we can drop in if ever needed. The shortcoming it that it's 31-bit addressable, so no bigger than 2GB.

src/main/java/com/conveyal/r5/util/ExpandingMMFBytez.java
abyrd: Updated comment.

PR #215, opened 2017-01-14
fix failing tests

evansiroky:


buma:
This seems OK. Thanks for fix.

PR #221, opened 2017-01-17
Isochrone endpoint

mattwigway:


src/main/java/com/conveyal/r5/point_to_point/PointToPointRouterServer.java
```diff
@@ -113,6 +121,8 @@ private static void run(TransportNetwork transportNetwork) {
         ObjectReader graphQlRequestReader = mapper.reader(GraphQlRequest.class);
         ObjectReader mapReader = mapper.reader(HashMap.class);
         staticFileLocation("debug-plan");
+        transportNetwork.transitLayer.buildDistanceTables(null);
```
abyrd: Generally we don't need to build these tables in point-to-point routing, so this may be adding a lot of startup time and memory consumption when someone just wants to do routing.

src/main/java/com/conveyal/r5/point_to_point/PointToPointRouterServer.java
```diff
@@ -1048,4 +1121,43 @@ private static float roundSpeed(float speed) {
         return Math.round(speed * 1000) / 1000;
     }
 
+    private static AnalystClusterRequest makeClusterRequest(TransportNetwork transportNetwork,
```
abyrd: The name of this function should probably indicate that this request serves a specific purpose. Even if its purpose within the PR or the commit is clear, as an element within PointToPointRouterServer.java it's not clear that this is intended only for generating isochrones.

src/main/java/com/conveyal/r5/analyst/ResultSet.java
```diff
@@ -230,6 +235,32 @@ public void writeIsochrones (JsonGenerator jgen) throws IOException {
         jgen.writeEndArray();
     }
 
+    public void writeIsochrones(List<GeoJsonFeature> features, boolean returnDistinctAreas) {
```
abyrd: Unlike the other function called writeIsochrones, I don't see this function writing any JSON. It appears to just create GeoJSON features and put them into the list that was passed in. It also may not be ideal to structure a function to have side effects on a collection passed in as a parameter, when it's still possible to return values from that function (void return type).

abyrd:
I added a few line comments. The main comment I have here is that all new functions should have documentation (comments) explaining what they do and how they are used, and care should be taken to make function names clearly describe the purpose of the function given the context of their enclosing class.

src/main/java/com/conveyal/r5/point_to_point/PointToPointRouterServer.java
evansiroky: If I don't do this, then I receive null pointer exceptions in all calls to the isochrone endpoint.  Also, when running this on my small Terre Haute network, it doesn't seem to take any time at all.

abyrd: Right, if you are going to do analysis work like making isochrones, you will need to build some additional tables like these. However, you have modified an existing tool that does point to point routing so that it will always build the tables when starting up. Creation of these analysis-oriented tables was intentionally made optional because they are so big and take so long to calculate. Obviously it's not a big deal on the small network you are working with, but it could take half an hour on a graph the size of the Netherlands. So if someone wants to build a graph quickly and do some routing in the Netherlands, this is making the task maybe more than twice as slow.

Currently we build the tables only in TransportNetworkCache, which is used by Analysis workers. When a network is built in that context, it contains the analysis-oriented tables which are serialized along with the graph.

When the network is built outside of an Analysis worker, those tables are not included. If you need them for your application, I think that you should have a command line switch to enable creating these extra tables. There should also be a check to see whether the deserialized graph already contains the tables, and maybe you'd want to create the tables before serialization (once) rather than after deserialization (repeatedly).

evansiroky:
@abyrd I addressed your comments in the above commit.

mattwigway:
@abyrd how do you feel about merging this?

abyrd:
Looks pretty good to me. @evansiroky thanks for adding the comments. I find it's better to just leave out trivial param and returns javadoc items. e.g. if the returns item does not express any information beyond the declared type of the function, there's no reason to include it in the Javadoc.

abyrd:
@mattwigway feel free to merge if it's ready

src/main/java/com/conveyal/r5/analyst/ResultSet.java
```diff
@@ -230,6 +234,53 @@ public void writeIsochrones (JsonGenerator jgen) throws IOException {
...
 
+    /**
+     * Reduces and simplifies (using DouglasPeuckerSimplifier) the inputted geometry.
+     *
+     * @param geometry  The geometry to modify
+     * @return          The new geometry that is simplified and then reduced
+     */
+    private Geometry reduceAndSimplifyGeometry(Geometry geometry) {
+        PrecisionModel precisionModel = new PrecisionModel(10000);
+        GeometryPrecisionReducer precisionReducer = new GeometryPrecisionReducer(precisionModel);
```
mattwigway: @evansiroky this isn't blocking on a merge, but we may want to put the precisionReducer in a static field so it isn't initialized whenever this function is called.

evansiroky:
Thanks for merging!

PR #223, opened 2017-01-27
Improve modeify

mattwigway:


mattwigway:
@abyrd I went ahead and merged this so we could move forward with some Modeify stuff, but can you take a look when you get a chance?

abyrd:
Yes, I've got it open now and will take a look tonight or in the morning.

src/main/java/com/conveyal/r5/profile/PathWithTimes.java
```diff
@@ -218,9 +225,22 @@ public int compareTo(Itinerary o) {
         }
     }
 
+    public int hashCode () {
+        return super.hashCode() ^ (accessMode.hashCode() * 31 + egressMode.hashCode() * 29);
+    }
+
+    public boolean equals (Object o) {
```
abyrd: Defining hash code and equals functions that do not compare / include all fields requires very clear documentation as to why this is the case, and how it is used. It appears to also be the case for the superclass.

mattwigway:
Yes, thanks for the comments, I will address the two issues you identified. This is to increase result variety but not computation time. The tickets that describe it are in Asana which you can't link to github commits, which is why they are not linked.

Issue #224, opened 2017-01-30
Buffer transport network when creating linked grid pointset


mattwigway:
When we create a WebMercatorGridPointset for a TransportNetwork, we use the bounds of that transport network, which means that origins outside of the TransportNetwork bounds throw an error "Sub-grid must lie fully inside the super-grid." This is an issue, for example, in Santa Barbara, CA (go gauchos!), where transit service ends a few blocks from the southernmost point (the beach) in downtown, so one cannot compute an isochrone from the beach because it's outside the bounds. We should buffer the bounds of the transport network with a reasonable distance for people coming from outside of the transport network. Of course, maybe if you live on the beach you are super chill and needn't worry about isochrones.

This only affects single-point mode as the walking travel times are passed to the client as a grid.

mattwigway:
I'm thinking a 1km buffer.

mattwigway:
This is particularly nasty in SB, because the urban area is so narrow (jammed onto a narrow plain between 🌊  and ⛰ ) that a haphazardly drawn box in analysis-ui around the region may have a centroid that lies outside the network.

abyrd:
This does seem reasonable, and it seems to me that the buffer should be the same as the walk distance to reach transit.

mattwigway:
So StaticComputer currently has the grid size used to report nontransit travel times hardwired at a 41-pixel-square grid, which means to be sure we avoid issues with this we need to make sure that the implicit pointset is 20 pixels bigger in each dimension than the most extreme origin point.

abyrd:
This should eventually not be hard-wired to 41 pixels, so it seems like we should be buffering not based on a number of pixels but rather the maximum walk distance from which the non-transit grid should be calculated.

PR #225, opened 2017-01-30
Buffer implied pointsets, and allow sub-pointsets to be uncontained b…

mattwigway:
This does two things:

1) Buffers the implied WebMercatorGridPointset so that it expands beyond the street network extent (#224). I used the max offstreet walk distance rather than the distance table size, because if points are further than that from the extent of the street network they are perforce unusable.
2) Allows sub linked pointsets to not completely overlap the superpointset. Currently generating Browsochrones data near the edge of the graph fails because the sub point set created around the origin to report non-transit times have points outside the implied pointset. This way those points outside the graph pointset are simply left unlinked. The reason this is a problem now and not in the past is that we recently made a change to simply subset the implied linked pointset, rather than generating a new linked pointset, when computing the non-transit times. Thus, we won't see walk paths to areas outside the extent of the implied pointset anymore, but this seems like a small loss - not linking a new, small grid pointset was a performance win.

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/WebMercatorGridPointSet.java
+++ b/src/main/java/com/conveyal/r5/analyst/WebMercatorGridPointSet.java
@@ -55,9 +55,9 @@ public class WebMercatorGridPointSet extends PointSet implements Serializable {
 
         this.zoom = DEFAULT_ZOOM;
 
-        double latBuffer = SphericalDistanceLibrary.metersToDegreesLatitude(TransitLayer.DISTANCE_TABLE_SIZE_METERS);
+        double latBuffer = SphericalDistanceLibrary.metersToDegreesLatitude(LinkedPointSet.MAX_OFFSTREET_WALK_METERS);
         double lonBuffer = SphericalDistanceLibrary
-                .metersToDegreesLongitude(TransitLayer.DISTANCE_TABLE_SIZE_METERS, transportNetwork.streetLayer.envelope.centre().y);
+                .metersToDegreesLongitude(LinkedPointSet.MAX_OFFSTREET_WALK_METERS, transportNetwork.streetLayer.envelope.centre().y);
 
         int west = lonToPixel(transportNetwork.streetLayer.envelope.getMinX() - lonBuffer);
         int east = lonToPixel(transportNetwork.streetLayer.envelope.getMaxX() + lonBuffer);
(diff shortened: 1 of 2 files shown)


src/main/java/com/conveyal/r5/analyst/WebMercatorGridPointSet.java
```diff
@@ -50,10 +54,15 @@ public WebMercatorGridPointSet(TransportNetwork transportNetwork) {
...
-        int west = lonToPixel(transportNetwork.streetLayer.envelope.getMinX());
-        int east = lonToPixel(transportNetwork.streetLayer.envelope.getMaxX());
-        int north = latToPixel(transportNetwork.streetLayer.envelope.getMaxY());
-        int south = latToPixel(transportNetwork.streetLayer.envelope.getMinY());
+
+        double latBuffer = SphericalDistanceLibrary.metersToDegreesLatitude(LinkedPointSet.MAX_OFFSTREET_WALK_METERS);
+        double lonBuffer = SphericalDistanceLibrary
+                .metersToDegreesLongitude(LinkedPointSet.MAX_OFFSTREET_WALK_METERS, transportNetwork.streetLayer.envelope.centre().y);
+
+        int west = lonToPixel(transportNetwork.streetLayer.envelope.getMinX() - lonBuffer);
```
abyrd: This could also be done with 

```
        Envelope envelope = new Envelope(transportNetwork.streetLayer.envelope);
        envelope.expandBy(lonBuffer, latBuffer);
```

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
```diff
@@ -174,10 +176,22 @@ public LinkedPointSet(LinkedPointSet sourceLinkage, WebMercatorGridPointSet subG
         distances0_mm = new int[nCells];
         distances1_mm = new int[nCells];
 
+        // mark all points as unlinked, in case part of the subGrid lies outside the super grid. Points within the super
+        // grid will be overwritten below
+        Arrays.fill(edges, -1);
```
abyrd: Maybe we should be using a symbolic constant like UNLINKED for this, but I guess we have hard-coded -1s all over.

abyrd:
@mattwigway I'm going through these older pull requests, hoping to merge some of them. Is this one still applicable? Will there be any side effects to merging it (interference with other changes in the front end for example)?

abyrd:
Revisiting this in October 2020, after combining r5 with analysis-backend. These changes may still be useful but my previous comment still stands - it will take some time to determine whether this is applicable or will necessitate changes elsewhere in the system.

abyrd:

ansoncfit:
This may be superseded by #684 and #688

PR #227, opened 2017-02-07
Faster isochrone generation

mattwigway:
A few changes here:
 * Fix bug with incorrect speeds being used for propagation that caused extra holes and shells in car isochrones
 * Simply outer rings to make finding the relevant rings for holes work
 * Get rid of really small rings


evansiroky:
build passes locally for me too

PR #236, opened 2017-02-14
New raptor worker

mattwigway:
@abyrd not ready for merge but please review.

### DIFF
--- a/src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
+++ b/src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
@@ -215,12 +215,6 @@ public class FastRaptorWorker {
                 // Einstein was probably wrong; God does in fact play dice with the universe, and so do we
                 offsets.randomize();
 
-                // re-mark all access stops
-                accessStops.forEachEntry((stop, time) -> {
-                    frequencyState[0].bestStopsTouched.set(stop);
-                    return true;
-                });
-
                 for (int round = 1; round <= request.maxRides; round++) {
                     frequencyState[round].min(frequencyState[round - 1]);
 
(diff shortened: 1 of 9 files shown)


mattwigway:
I'll rebase this tomorrow then it should be good to go.

mattwigway:
@abyrd should be good to go! Let's merge this after we merge #248 and #249 and release 2.1.0.

src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
```diff
@@ -0,0 +1,559 @@
...
+                timeInScheduledSearchFrequencyBounds += System.nanoTime() - frequencyStartTime;
+
+                long transferStartTime = System.nanoTime();
+                doTransfers(scheduleState[round]);
+                timeInScheduledSearchTransfers += System.nanoTime() - transferStartTime;
+            }
+            timeInScheduledSearch += System.nanoTime() - startTime;
+        }
+
+        if (transit.hasFrequencies) {
```
abyrd: Maybe tack a comment in here to remind the reader that the frequency search is building on the results of the scheduled search at this particular minute.

src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
```diff
@@ -0,0 +1,559 @@
...
+        accessStops.forEachEntry((stop, accessTime) -> {
+            initialState.setTimeAtStop(stop, accessTime + departureTime, -1, -1, 0, 0, true);
+            return true; // continue iteration
+        });
+    }
+
+    /** Perform one minute of a RAPTOR search */
+    private int[][] runRaptorForMinute(int iterationsPerMinute) {
+        // Run the scheduled search
+        // round 0 is the street search
```
abyrd: Perhaps augment this comment to clarify that the street search is reusing some state from a departure at the subsequent minute in the time window.

src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
```diff
@@ -0,0 +1,559 @@
...
+
+        // add initial stops
+        RaptorState initialState = scheduleState[0];
+        accessStops.forEachEntry((stop, accessTime) -> {
+            initialState.setTimeAtStop(stop, accessTime + departureTime, -1, -1, 0, 0, true);
+            return true; // continue iteration
+        });
+    }
+
+    /** Perform one minute of a RAPTOR search */
```
abyrd: Also please explain what is in the int[][] returned by this function. Looks like it would be an array of n arrays of travel times to every stop, where n= iterationsPerMinute.

src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
```diff
@@ -0,0 +1,559 @@
...
+                        }
+
+                        state.setTimeAtStop(targetStop, timeAtTargetStop, -1, stop, 0, 0, true);
+                    }
+                }
+            }
+        }
+    }
+
+    /** Get a list of the internal IDs of the patterns touched using the given index (frequency or scheduled) */
```
abyrd: Clarify what it means for a pattern to be "touched": the stop was updated in the search round that produced the state, because a faster way was found to reach it. So we're getting a bitset of all patterns that contain stops whose travel time was improved (updated) in the round that produced the supplied state.

src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
```diff
@@ -0,0 +1,559 @@
...
+    private BitSet getPatternsTouchedForStops(RaptorState state, int[] index) {
+        BitSet patternsTouched = new BitSet();
+
+        for (int stop = state.bestStopsTouched.nextSetBit(0); stop >= 0; stop = state.bestStopsTouched.nextSetBit(stop + 1)) {
+            final int finalStop = stop;
```
abyrd: Add comment to say this is a workaround for Java lambdas. I stared at this for about 30 seconds trying to figure out how every stop in the pattern could be the "final" one in the pattern :)

src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
```diff
@@ -0,0 +1,559 @@
...
+            transit.patternsForStop.get(stop).forEach(originalPattern -> {
+                int filteredPattern = index[originalPattern];
+
```
abyrd: We should be able to `return true` here immediately after seeing filteredPattern < 0, simplifying the conditional below, avoiding the sourcePatternIndex computation, and making it clearer why the iteration is being skipped over.

src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
```diff
@@ -0,0 +1,559 @@
...
+                int sourcePatternIndex = state.previousStop[finalStop] == -1 ?
+                        state.previousPatterns[finalStop] :
+                        state.previousPatterns[state.previousStop[finalStop]];
+
+                if (filteredPattern > -1 && sourcePatternIndex != originalPattern) {
```
abyrd: Why do we not want to consider the pattern "touched" if sourcePatternIndex == originalPattern? This seems to have something to do with not re-boarding the same patterns, but I can't quite put my finger on it.

src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
```diff
@@ -0,0 +1,559 @@
...
+    private final TransitLayer transit;
+
+    /** Times to access each transit stop using the street network (seconds) */
+    private final TIntIntMap accessStops;
+
+    /** The profilerequest describing routing parameters */
+    private final ProfileRequest request;
+
+    /** Frequency-based trip patterns running on a given day */
+    private TripPattern[] runningFrequencyPatterns;
```
abyrd: I believe these six fields could be collapsed down to just two bitsets for active frequency and active scheduled patterns (not doing any mapping, just pre-filtering and using the original indexes). Maybe I'm missing something though. I'd also be curious to see the speed difference with no filtering at all, just to see if the complexity is worth the speedup.

src/main/java/com/conveyal/r5/profile/Propagater.java
```diff
@@ -0,0 +1,88 @@
...
+        // NB if this is slow, parallelize
+        // This can use something akin to range-RAPTOR where we only update the travel time if it changes from one
+        // iteration to the next, but that requires working on clock times, so we'd have to pass more information down
+        // into this function; I'm not sure it's worth it.
+        int[] result = new int[travelTimesToStopsEachIteration.length];
+        for (int iteration = 0; iteration < travelTimesToStopsEachIteration.length; iteration++) {
+            int[] travelTimesToTargetsThisIteration = Arrays.copyOf(nonTransferTravelTimesToTargets, nonTransferTravelTimesToTargets.length);
+            int[] travelTimesToStopsThisIteration = travelTimesToStopsEachIteration[iteration];
+
+            // avoid float math in loop below
```
mattwigway: This is because we had this same loop in the previous propagation code and at some point determined it to be slow. It's probably not because of floating-point ops but rather casts, although I don't know if we ever tried it with single-precision floats.

src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
mattwigway: Is this something that blocks merge, or should we merge and then try it?

src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
```diff
@@ -0,0 +1,559 @@
...
+        LOG.info("Frequency search: {}s", timeInFrequencySearch / 1e9d);
+        LOG.info("  - Frequency component: {}s", timeInFrequencySearchFrequency / 1e9d);
+        LOG.info("  - Resulting updates to scheduled component: {}s", timeInFrequencySearchScheduled / 1e9d);
+        LOG.info("  - Transfers: {}s", timeInFrequencySearchTransfers / 1e9d);
+
+        return results;
+    }
+
+    /** Prefilter the patterns to only ones that are running */
+    private void prefilterPatterns () {
```
mattwigway: My thought was that filtering might get the patterns into the CPU cache, but since they're references and they won't all be on the same page perhaps not.

mattwigway:
@abyrd I believe I've addressed all your concerns.

src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
abyrd: No, it doesn't block the merge. Just observations as I was reviewing. I'll create a ticket for this after the merge.

abyrd:
Thanks for adding those comments @mattwigway, looks very good! Will merge this one right after #249

Issue #238, opened 2017-02-17
Browsochrones-style requests are not considered when deciding if a machine has been used recently


mattwigway:
Currently a machine will shut down if it hasn't received a single-point request in the last 15 minutes. We should also check if it has received a browsochrones-style single point request from the new UI.

PR #239, opened 2017-02-17
Consider browsochrones requests before shutting down AnalystWorker (fx #238)

mattwigway:


src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
```diff
@@ -145,7 +145,7 @@
     /** Information about the EC2 instance (if any) this worker is running on. */
     EC2Info ec2info;
 
-    /** TODO what's this number? */
+    /** The time the last high priority request was processed, used to check if the machine should be shut down. */
```
abyrd: I think we should always mention units on any number, even when it seems obvious.

src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
```diff
@@ -421,6 +421,7 @@ private void handleStaticSiteRequest (StaticSiteRequest.PointRequest request, Tr
 
         if (request.request.bucket != null) computer.run();
         else {
+            lastHighPriorityRequestProcessed = System.currentTimeMillis();
```
abyrd: Might be worth adding a comment about why it's only a highPriorityRequest when bucket != null (because batch regional analyses save results to buckets, while one-off analyses don't)

PR #243, opened 2017-02-23
Remove time when removing stops

mattwigway:
This also adds the ability to return warnings to the frontend.

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/broker/BrokerHttpHandler.java
+++ b/src/main/java/com/conveyal/r5/analyst/broker/BrokerHttpHandler.java
@@ -3,6 +3,7 @@ package com.conveyal.r5.analyst.broker;
 import com.conveyal.r5.analyst.cluster.AnalystWorker;
 import com.conveyal.r5.analyst.cluster.GenericClusterRequest;
 import com.conveyal.r5.analyst.cluster.WorkerStatus;
+import com.conveyal.r5.analyst.error.TaskError;
 import com.conveyal.r5.common.JsonUtilities;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
@@ -65,6 +66,10 @@ class BrokerHttpHandler extends HttpHandler {
         if (pathComponents.length < 2) {
             response.setStatus(HttpStatus.BAD_REQUEST_400);
             response.setDetailMessage("path should have at least one part");
+            TaskError error = new TaskError(null, "Path should have at least one part", "");
+            OutputStream os = response.getOutputStream();
+            mapper.writeValue(os, new TaskError[] { error });
+            return;
         }
 
         String command = pathComponents[1];
@@ -167,6 +172,10 @@ class BrokerHttpHandler extends HttpHandler {
                             !task.workerVersion.equals(exemplar.workerVersion)) {
                             response.setStatus(HttpStatus.BAD_REQUEST_400);
                             response.setDetailMessage("All tasks must be for the same graph, job, and worker commit.");
+                            TaskError error = new TaskError(null, "All tasks must be for the same graph, job, and worker commit.", "");
+                            OutputStream os = response.getOutputStream();
+                            mapper.writeValue(os, new TaskError[] { error });
+                            os.close();
                             return;
                         }
                     }
@@ -242,14 +251,27 @@ class BrokerHttpHandler extends HttpHandler {
                 } else {
                     response.setStatus(HttpStatus.BAD_REQUEST_400);
                     response.setDetailMessage("Delete is only allowed for tasks and jobs.");
+                    TaskError error = new TaskError(null, "Delete is only allowed for tasks and jobs", "");
+                    OutputStream os = response.getOutputStream();
+                    mapper.writeValue(os, new TaskError[] { error });
+                    os.close();
                 }
             } else {
                 response.setStatus(HttpStatus.BAD_REQUEST_400);
                 response.setDetailMessage("Unrecognized combination of HTTP method and command.");
+                TaskError error = new TaskError(null, "Unrecognized combination of HTTP method and command.", "");
+                OutputStream os = response.getOutputStream();
+                mapper.writeValue(os, new TaskError[] { error });
+                os.close();
             }
         } catch (JsonProcessingException jpex) {
             response.setStatus(HttpStatus.BAD_REQUEST_400);
             response.setDetailMessage("Could not decode/encode JSON payload. " + jpex.getMessage());
+            TaskError error = new TaskError(null, "Could not decode/encode JSON payload. ", jpex.getMessage());
+            OutputStream os = response.getOutputStream();
+            mapper.writeValue(os, new TaskError[] { error });
+            os.close();
+            
             LOG.info("Error processing JSON from client", jpex);
         } catch (Exception ex) {
             response.setStatus(HttpStatus.INTERNAL_SERVER_ERROR_500);
--- a/src/main/java/com/conveyal/r5/analyst/scenario/Modification.java
+++ b/src/main/java/com/conveyal/r5/analyst/scenario/Modification.java
@@ -69,7 +69,9 @@ public abstract class Modification implements Serializable {
     /**
      * Any warnings that should be presented to the user but which do not prevent scenario application should appear here.
      */
-    public transient final Set<String> warnings = new HashSet<>();
+    // TODO this should be transient as well but previously wasn't. R5 modifications are stored in MongoDB in regional
+    // analyses, which means that marking it transient causes deserialization headaches.
+    public final Set<String> warnings = new HashSet<>();
 
     /**
      * Apply this single modification to a TransportNetwork.
--- a/src/main/java/com/conveyal/r5/analyst/scenario/Modification.java
+++ b/src/main/java/com/conveyal/r5/analyst/scenario/Modification.java
@@ -64,12 +64,12 @@ public abstract class Modification implements Serializable {
      * The "resolve" method is called on each Modification before it is applied. If any problems are detected, the
      * Modification should not be applied, and this Set should contain Strings describing all the problems.
      */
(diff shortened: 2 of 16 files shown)


src/main/java/com/conveyal/r5/analyst/scenario/Scenario.java
```diff
@@ -113,7 +117,16 @@ public TransportNetwork applyToTransportNetwork (TransportNetwork originalNetwor
...
                 throw new ScenarioApplicationException(Arrays.asList(modification));
             }
+
+            if (!modification.warnings.isEmpty()) {
+                modificationsWithWarnings.add(modification);
+            }
         }
+
+        copiedNetwork.scenarioApplicationWarnings = modificationsWithWarnings.stream()
+                .map(m -> new TaskError(m, m.warnings))
```
abyrd: It seems like this constructor should take one object of type Modification (since the second argument is implicit in the first), and then we should call  `modificationsWithWarnings.stream().map(TaskError::new)`

src/main/java/com/conveyal/r5/analyst/broker/BrokerHttpHandler.java
```diff
@@ -65,6 +66,10 @@ public void service(Request request, Response response) throws Exception {
         if (pathComponents.length < 2) {
             response.setStatus(HttpStatus.BAD_REQUEST_400);
             response.setDetailMessage("path should have at least one part");
+            TaskError error = new TaskError(null, "Path should have at least one part", "");
```
abyrd: This would usually be caused by an error in how our components communicate with each other, so I'm not sure we need to wrap the error message in a data model object that's intended to show user-oriented errors.

abyrd: Then again I see that you've done this throughout the class, so I guess the idea is to change the purpose/meaning of TaskError to represent all errors.

src/main/java/com/conveyal/r5/analyst/scenario/Modification.java
```diff
@@ -65,7 +64,14 @@ public final void setType (String type) {
      * The "resolve" method is called on each Modification before it is applied. If any problems are detected, the
      * Modification should not be applied, and this Set should contain Strings describing all the problems.
      */
-    public final Set<String> warnings = new HashSet<String>();
+    public transient final Set<String> errors = new HashSet<>();
```
abyrd: I'm glad to see the name of this field change, it was inaccurate.

src/main/java/com/conveyal/r5/analyst/scenario/RemoveStops.java
```diff
@@ -47,6 +48,16 @@
     /** Stops to remove from the routes. */
     public Set<String> stops;
 
+    /**
+     * Number of seconds (in addition to dwell time, if any) to remove from the schedule at each removed stop,
+     * to account for acceleration, deceleration, merging, and so on.
```
abyrd: Maybe we should say "vehicle acceleration and deceleration" and "merging into traffic" to make it clearer that we're talking about time saved by physical changes in bus operations and not something in code.

src/main/java/com/conveyal/r5/analyst/scenario/RemoveStops.java
```diff
@@ -47,6 +48,16 @@
...
+     *
+     * If removing this time would make the hop time between two stops that remain <= 0, an error will be surfaced.
```
abyrd: It's actually a warning that's surfaced in the UI.

src/main/java/com/conveyal/r5/analyst/scenario/RemoveStops.java
```diff
@@ -147,23 +158,63 @@ public TripPattern processTripPattern (TripPattern originalTripPattern) {
...
                         // Only accumulate dwell time at the beginning of the output trip, to preserve the time offset of
                         // the first arrival in the output. After that, dwells are not retained for skipped stops.
                         accumulatedRideTime += dwellTime;
                     }
                 } else {
-                    newSchedule.arrivals[j] = prevOutputDeparture + accumulatedRideTime + rideTime;
+                    accumulatedRideTime += rideTime;
+                    int secondsToRemove = secondsSavedAtEachStop * nStopsRemovedSinceLastStop;
+                    if (nStopsRemovedSinceLastStop > 0 && accumulatedRideTime < secondsToRemove) {
+                        // figure out which stops are causing the problem
```
abyrd: The comment says we're finding stops that cause a problem, but doesn't say that this conditional block handles a problem, or what that problem is.

src/main/java/com/conveyal/r5/analyst/scenario/RemoveStops.java
```diff
@@ -147,23 +158,63 @@ public TripPattern processTripPattern (TripPattern originalTripPattern) {
...
                 int dwellTime = originalSchedule.departures[i] - originalSchedule.arrivals[i];
                 prevInputDeparture = originalSchedule.departures[i];
                 if (removeStop[i]) {
-                    // The current stop will not be included in the output. Record the travel time.
+                    // The current stop will not be included in the output. Record the travel time and subtract off the
+                    // seconds saved. This may cause accumulatedRideTime to go negative, which may be fine since we're putting
+                    // all the time savings before the stop. After we've accumulated ride time for a whole segment,
+                    // we check to make sure it's positive.
                     accumulatedRideTime += rideTime;
+                    nStopsRemovedSinceLastStop++;
```
abyrd: It seems odd to me to count how many stops have been removed and then multiply that by the number of stops removed. Shouldn't it be possible to subtract secondsSavedAtEachStop out of accumulatedRideTime at each iteration, and just keep an eye on the accumulated total?

src/main/java/com/conveyal/r5/analyst/scenario/RemoveStops.java
```diff
@@ -147,23 +158,63 @@ public TripPattern processTripPattern (TripPattern originalTripPattern) {
...
+                    int secondsToRemove = secondsSavedAtEachStop * nStopsRemovedSinceLastStop;
+                    if (nStopsRemovedSinceLastStop > 0 && accumulatedRideTime < secondsToRemove) {
+                        // figure out which stops are causing the problem
+                        int[] problemStops = new int[nStopsRemovedSinceLastStop];
+                        for (int removedStop = i - 1, index = nStopsRemovedSinceLastStop - 1; removedStop >= i - nStopsRemovedSinceLastStop; removedStop--, index--) {
+                            problemStops[index] = originalTripPattern.stops[removedStop];
+                        }
+
+                        // map them back to names
+                        String problemStopNames = IntStream.of(problemStops)
```
abyrd: Having the error handling code inserted in the conditional body inside the loop stretches out the blocks and makes it a bit harder to see the overall flow. It seems sufficient to just record all the offending stops in a set and handle the error generation outside the inner blocks, potentially even in a function.

src/main/java/com/conveyal/r5/analyst/scenario/Scenario.java
```diff
@@ -93,11 +95,13 @@ public TransportNetwork applyToTransportNetwork (TransportNetwork originalNetwor
         // Check all the parameters before applying any modifications.
         // Might some parameters may become valid/invalid because of previous modifications in the list?
         List<Modification> badModifications = new ArrayList<>();
```
abyrd: Parallelism: if the next field is called modificationsWithWarnings maybe this one should be renamed modificationsWithErrors.

src/main/java/com/conveyal/r5/analyst/scenario/Scenario.java
mattwigway: The second argument is actually not implicit in the first, because a task error can describe either the errors or the warnings.

src/main/java/com/conveyal/r5/analyst/broker/BrokerHttpHandler.java
mattwigway: The reason I did this is because anything that causes a 400 error is expected by the frontend to have a JSON-formatted error payload so we can display the error.

src/main/java/com/conveyal/r5/analyst/scenario/RemoveStops.java
mattwigway: Sure, although I'm not sure what code operation is implied by acceleration :)

src/main/java/com/conveyal/r5/analyst/scenario/Scenario.java
mattwigway: Fixed.

src/main/java/com/conveyal/r5/analyst/scenario/RemoveStops.java
mattwigway: And of course merging into traffic could be confusing once we have a branch called traffic :) :)

mattwigway: But seriously, thanks, I updated the comment.

src/main/java/com/conveyal/r5/analyst/scenario/RemoveStops.java
mattwigway: Fixed.

src/main/java/com/conveyal/r5/analyst/scenario/RemoveStops.java
mattwigway: Yes, we could do it that way. I didn't because we need to know the number of stops removed so we can generate the warning messages, but there wouldn't be any harm in just keeping that separate from the actual calculation.

mattwigway:
@abyrd I believe I have addressed your concerns, please re-review. I did not change how the time removed get subtracted from accumulated ride time; see my response to your comment and let me know if you'd still like me to make that change.

abyrd:
@mattwigway thanks, at the end of the day I suppose it's a question of style and the functionality is needed soon, so let's just merge it as is.

PR #248, opened 2017-03-01
Replace island removal code with Tarjan's algorithm.

### LINKED ISSUE #27: All Non walkable street are removed when islands are removed
I was debugging missing motorways and found that since [removeDisconnectedSubgraph](https://github.com/conveyal/r5/blob/master/src/main/java/com/conveyal/r5/streets/StreetLayer.java#L381) uses streetRouter which uses WALK mode by default ever since permissions were added non walkable streets are removed. This is mostly motorways. When I build graph without removing islands motorways are back.

When I added support for permission in TN previously same thing happened. I added switchMode parameter to request which used any mode available during routing. [Commit](https://github.com/buma/OpenTripPlanner-Maribor/commit/d500365570367a08653d65c8ccd22a2f070b01f7) if this would be OK I can do the same here.

In OTP edges which are supposed to be prunned are first checked if they have traversable permissions and are removed only if they can't be walked or cycled. If I understand the [code](https://github.com/opentripplanner/OpenTripPlanner/blob/master/src/main/java/org/opentripplanner/common/StreetUtils.java#L123) correctly.

### LINKED ISSUE #228: Car routing fails when origin gets snapped to one-way cul-de-sac
Car routing fails when the origin is snapped to [this way](http://www.openstreetmap.org/way/207971576) which is a one-way street that dead-ends into the entrance of a parking garage, but the interior of the parking garage is not mapped.

We currently have an island removal process which makes sure that every part of the graph is part of a component of at least size _n_. What we really need to do is make sure that all edges and vertices are members of _strong_ components, but our simple flood-fill algorithm does not work in this case.

One could argue that this is purely a data problem and should not be fixed on the R5 side, but that is also true of island removal. This is just a more general case of island removal.

mattwigway:
This finds all strong components, meaning one-way cul-de-sacs will be eliminated.
It also is applied to all modes, so car/bike islands will be eliminated as well.

Fixes #228 and fixes #27.

@abyrd if you can review that would be great, ideally we'd also try building an NL graph to make sure the algorithm is tractable on a larger graph.

@evansiroky just tagging you since you reported the original issue.

### DIFF
--- a/src/main/java/com/conveyal/r5/streets/TarjanIslandPruner.java
+++ b/src/main/java/com/conveyal/r5/streets/TarjanIslandPruner.java
@@ -1,0 +1,326 @@
+package com.conveyal.r5.streets;
+
+import com.conveyal.r5.profile.StreetMode;
+import gnu.trove.iterator.TIntIterator;
+import gnu.trove.list.TIntList;
+import gnu.trove.set.TIntSet;
+import gnu.trove.set.hash.TIntHashSet;
+import gnu.trove.stack.TIntStack;
+import gnu.trove.stack.array.TIntArrayStack;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.BitSet;
+import java.util.List;
+import java.util.function.Consumer;
+
+/**
+ * Prune islands from a graph using Tarjan's strong-components algorithm, described in
+ * Tarjan, R. “Depth-First Search and Linear Graph Algorithms.” SIAM Journal on Computing 1, no. 2 (1972): 146–60. doi:10.1137/0201010.
+ * and summarized at
+ * https://en.wikipedia.org/wiki/Tarjan%27s_strongly_connected_components_algorithm
+ *
+ * We prune islands by mode; what is an island for pedestrians is probably not an island for cars.
+ *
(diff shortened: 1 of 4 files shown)


abyrd:
I just built a graph with the data for all of the Netherlands. It seems perfectly tractable, and necessary given the number of islands removed.

CAR:
557967 strong components in 26 sec
removed 556960 islands (601232 vertices) in 0.163 sec

WALK:
78590 strong components in 4 sec
removed 78524 islands (97007 vertices) in 0.031 sec

BICYCLE:
85708 strong components in 2.3 sec
removed 85642 islands (95811 vertices) in 0.025 sec

src/main/java/com/conveyal/r5/streets/TarjanIslandPruner.java
```diff
@@ -0,0 +1,326 @@
...
+            return true; // continue iteration over outgoing edges
+        });
+    }
+
+    /** Remove the permissions around a vertex for the desired mode. Returns the number of edges affected */
+    public void removePermissionsAroundVertex (int vertex) {
+        for (TIntList edgeList : new TIntList[] { streets.outgoingEdges.get(vertex), streets.incomingEdges.get(vertex) }) {
+            edgeList.forEach(eidx -> {
+                edgeCursor.seek(eidx);
+                switch (mode) {
```
abyrd: As mentioned above, this switch could probably be pulled out into a method on Edge to make things more readable.

PR #249, opened 2017-03-06
handle high priority browsochrones requests via side channels.

mattwigway:
Previously the side channel would only open for old-style analyst requests, meaning that single point jobs would not complete when a regional job was running on the same graph. This fixes that and also cleans up the auto-shutdown code.

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
+++ b/src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
@@ -300,7 +300,7 @@ public class AnalystWorker implements Runnable {
 
             // Enqueue high-priority (interactive) tasks first to ensure they are enqueued
             // even if the low-priority batch queue blocks.
-            tasks.stream().filter(AnalystWorker::isHighPriorityTask)
+            tasks.stream().filter(GenericClusterRequest::isHighPriority)
                     .forEach(t -> highPriorityExecutor.execute(() -> {
                         LOG.warn("Handling single point request via normal channel, side channel should open shortly.");
                         this.handleOneRequest(t);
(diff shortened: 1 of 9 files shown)


src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
```diff
@@ -825,6 +821,20 @@ private void logQueueStatus() {
         LOG.debug("Waiting tasks: high priority: {}, batch: {}", highPriorityExecutor.getQueue().size(), batchExecutor.getQueue().size());
     }
 
+    public static boolean isHighPriorityTask (GenericClusterRequest req) {
```
abyrd: Conditional-based polymorphism looks a bit suspicious to me. Couldn't this be handled with an isHighPriority method on GenericClusterRequest?

src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
```diff
@@ -300,15 +300,15 @@ public void run() {
...
                         this.handleOneRequest(t);
                     }));
 
             // Enqueue low-priority (batch) tasks; note that this may block anywhere in the process
             logQueueStatus();
-            tasks.stream().filter(t -> !(t instanceof AnalystClusterRequest) || ((AnalystClusterRequest) t).outputLocation != null)
+            tasks.stream().filter(t -> !isHighPriorityTask(t))
                 .forEach(t -> {
                     // attempt to enqueue, waiting if the queue is full
                     while (true) {
```
abyrd: This isn't part of the PR but looks pretty questionable to me. The thread sleeps when insertion is rejected? It seems like the executor should handle that kind of thing.

mattwigway:
@abyrd should be good to go, please re-review.

abyrd:
Great, I find this very readable after the changes! It looks like you added a log statement that was making the log too long, which caused the build to fail. I removed that and will merge.

PR #261, opened 2017-04-04
Fix modeify time estimates by recomputing statistics using all patter…

### LINKED ISSUE #260: Point-to-point routing can produce extremely long travel times
We're seeing some unrealistically long travel times coming out of the point to point router:

image

(for http://localhost:5000/planner?from=211%20M%20Street%20Ne%2C%20Washington%2C%20DC%2C%20USA%2020002%2C%20Washington%2C%20District%20of%20Columbia&to=512%20N%20Alfred%20St%2C%20Alexandria%2C%20VA%2C%20USA%2022314%2C%20Alexandria%2C%20Virginia&modes=TRAINISH%2CWALK&start_time=7&end_time=9&days=M%E2%80%94F )

It appears that this is due to incorrect computation of the `waitStats` and `rideStats` objects on a ProfileOption. They are simply computed using the first set of patterns, rather than combining all sets of patterns. This is most problematic when there is a pattern that starts running at the end of the time window (which is why we're seeing it now on the Yellow Line, all kinds of crazy stuff is happening with the current SafeTrack work). This would have been a problem earlier, but usually it just means waiting a few extra minutes to get a train on a particular pattern (for instance, a Greenbelt rather than Mt Vernon train), so we didn't notice it.

mattwigway:
…ns in an option, fixes #260.

@buma as far as I can tell it was just using the stats from the first transit path encountered which led to some weird results when the first path was infrequent.

My main question is how the access and egress time maps work.

### DIFF
--- a/src/main/java/com/conveyal/r5/R5Main.java
+++ b/src/main/java/com/conveyal/r5/R5Main.java
@@ -17,6 +17,13 @@ import java.util.Arrays;
  */
 public class R5Main {
     public static void main (String... args) throws Exception {
+        System.out.println("__________________\n" +
+                "___  __ \\__  ____/\n" +
+                "__  /_/ /_____ \\  \n" +
+                "_  _, _/ ____/ /  \n" +
+                "/_/ |_| /_____/   \n" +
+                "                ");
+
         // Pull argument 0 off as the sub-command,
         // then pass the remaining args (1..n) on to that subcommand.
         String command = args[0];
--- a/src/main/java/com/conveyal/r5/api/ProfileResponse.java
+++ b/src/main/java/com/conveyal/r5/api/ProfileResponse.java
@@ -200,4 +200,9 @@ public class ProfileResponse {
             }
         }
     }
+
+    /** Recompute stats for all options, should be done once all options have been added */
+    public void recomputeStats (ProfileRequest request) {
+        this.options.forEach(o -> o.recomputeStats(request));
+    }
 }
--- a/src/main/java/com/conveyal/r5/api/util/ProfileOption.java
+++ b/src/main/java/com/conveyal/r5/api/util/ProfileOption.java
@@ -306,8 +306,8 @@ public class ProfileOption {
                         req,
                         // TODO is this intended to handle multiple access modes to the same stop?
                         // do these ever have more or less than one value?
-                        accessIndexes.values().stream().findFirst().orElse(null),
-                        egressIndexes.values().stream().findFirst().orElse(null),
+                        access.get(0).duration,
+                        egress.get(0).duration,
                         transit.size(),
                         fullItineraries);
 
--- a/src/main/java/com/conveyal/r5/api/util/ProfileOption.java
+++ b/src/main/java/com/conveyal/r5/api/util/ProfileOption.java
@@ -1,6 +1,9 @@
 package com.conveyal.r5.api.util;
 
+import com.conveyal.r5.profile.Path;
 import com.conveyal.r5.profile.PathWithTimes;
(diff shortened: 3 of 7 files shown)


buma:
access is map of all access paths to stops and direct paths. First all direct paths from origin and destination are added to the map in addDirect function, then all the paths from origin to first stop. Access paths to stops are unique based on mode and stop streetIndex. Which means that if there are multiple transit paths which starts with same mode and stop they are only saved once in access. Which transit path has which accessIndex is saved in accessIndexes in Profile option. Key is ModeStopIndex(LegMode, stopIndexStreetVertex).

To find out which accessIndexes are for access to transit you can look at itinerary list and get access index from its pointToPointConnection on all with 0 transit time or read the values of accessIndexes (since all those indexes are indexes for access times for transit, and keys are mode and start stop streetVertex).

Egress is deduplicated the same way as access. Indexes are in egressIndexes. The only difference is that it can be null (no transit, or destination is on stopVertex) and it doesn't have direct paths.

mattwigway:
@buma but doesn't any single profile option only have a single access and egress leg (or just a single access leg if it represents a direct option?) I don't think we're combining transit journeys that don't use the same stops, or combining transit journeys with non-transit journeys in a single option, are we?

buma:
I ran some tests and there is first Profile option which is for NonTransit and is usually more then one. Usually it has CAR and BICYCLE, since for longer routing WALK is too far.

Most of transit seems to have only 1 access. But some can have more. I found one that have 2: BICYCLE and BICYCLE_RENT.

We are combining transit journeys that use same stops. This happens in ProfileOption addTransit (hasSameStops check). So rarely it can happen that there are multiple access parts. (I just found a bug that for some reason for 2 access parts 6 itineraries are generated and they are duplicated for some reason).

Multiple access legs can be seen in /new.html when there are multiple itineraries under one summary.

Example request with multiple access points:
```java
profileRequest.fromLat = 38.88942947447528;
profileRequest.fromLon = -77.03338623046875;
profileRequest.toLat = 38.94338908847991;
profileRequest.toLon = -77.00660705566406; //timezone -5
profileRequest.wheelchair = false;
profileRequest.bikeTrafficStress = 4;
profileRequest.setTime("2016-06-05T07:30-05:00", "2016-06-05T10:30-05:00");

profileRequest.transitModes = EnumSet.of(TransitModes.TRANSIT);
profileRequest.accessModes = EnumSet.of(LegMode.BICYCLE_RENT,LegMode.WALK,LegMode.BICYCLE,LegMode.CAR_PARK);
profileRequest.egressModes = EnumSet.of(LegMode.WALK);
profileRequest.directModes = EnumSet.of(LegMode.WALK,LegMode.BICYCLE,LegMode.CAR);
```

"routes Red, E4 via FORT TOTTEN METRO STATIO" and "routes Red, 64 via FORT TOTTEN METRO STATION" have 2 access modes.

buma:
I know why there are same itineraries repeated multiple times. They aren't the same. Start and end time is different since they use different times of transit. for example 4 different with BICYCLE access time and 3 different with BIKE_RENT. Then we have 7 different itineraries with 2 access legs 1 egress and 1 transit.

abyrd:
@mattwigway at least according to our original definition of profile routing, we should be including trips that pass through different stops in the same option, as long as they pass through the same stop clusters. This ensures that when a single logical stop is split into several platforms or boarding locations, the itineraries using those different stops don't land in separate options.

abyrd:
Do we have a conclusion here? Should we merge this into the next point release?

abyrd:
Actually it looks like the pull request #273 off @demory's cfaz-deploy branch includes this PR. Is that the case?

PR #262, opened 2017-04-11
Turn restriction when edges are split

buma:
Fixes problems when turn restrictions aren't applied if from/via edge is split.

Also support splitting edges in scenarios when stops are added. If this happens original network stays the same since turnRestriction lists are copied. TurnRestrictions itself are copied only if from/via edge need to be changed.


abyrd:
@buma will you be adding more commits to this PR/branch since we talked about simplifying all turn restrictions (forward/backward) as "no-turn" restrictions, with no "only" restrictions? I just wanted to check before I review / consider merging this.

buma:
I think this can be merged. I'll do only turn -> no turn change on different branch, since I need to firstly simplify splitting. (Currently we have getOrCreateVertex and splitEdge) which both do basically the same. Only second is used only in PR linking.

abyrd:
Hi @buma I'm merging some pull requests now, but unfortunately won't have time to review this in detail right away. Hopefully I will merge it tomorrow in a point release. I want to make sure we get one release out with no deep changes to street routing, just so we don't risk accidentally changing analysis results.

abyrd:
The only thing @landonreed changed when resolving conflicts in his merge 35342b9 was a few imports.

abyrd:
Revisiting this in October 2020, after combining r5 with analysis-backend. These changes may still be applicable but it will take some time to evaluate correctness and determine whether this will necessitate changes elsewhere in the system. This is getting into some complex areas that probably do deserve attention but may take several days to work through properly.

abyrd:

PR #263, opened 2017-04-12
Regional analysis medians block bootstrap

mattwigway:
Not quite ready for merge I think, but ready for code review!


abyrd:
@mattwigway realistically I can't give this the proper amount of attention tonight. Please cherry pick the major bugfix if you'd like to take care of that today, and I'll review the rest tomorrow morning.

src/main/java/com/conveyal/r5/profile/PerTargetPropagater.java
```diff
@@ -0,0 +1,107 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Arrays;
+
+/**
+ * This is similar to the Propagater class, but instead of the reducer function being called with each iteration,
+ * it is called with the travel times to a particular target for all iterations.
+ *
+ * The reducer will be called with the target index, and a boolean array of whether that target is reachable at each iteration.
```
abyrd: Clarify: what is a "target index"? A 1D offset into an array representing a 2D grid of cells?

src/main/java/com/conveyal/r5/profile/PerTargetPropagater.java
```diff
@@ -0,0 +1,107 @@
...
+        long startTimeMillis = System.currentTimeMillis();
+        // avoid float math in loop below
+        // float math was previously observed to slow down this loop, however it's debatable whether that was due to
+        // casts, the operations themselves, or the fact that the operations were being completed with doubles rather
+        // than floats.
+        int speedMillimetersPerSecond = (int) (request.walkSpeed * 1000);
+
+        for (int targetIdx = 0; targetIdx < targets.size(); targetIdx++) {
+            boolean[] perIterationResults = new boolean[travelTimesToStopsEachIteration.length];
+            // fill with non-transit time as upper bound on travel time
```
abyrd: This is not filling it with the time, it's filling it with booleans representing whether this target cell is reachable without transit. The idea is analogous to an upper bound but I don't think we can apply the term to a boolean; it's sort of a bit pattern to be OR'ed with the transit results.

src/main/java/com/conveyal/r5/profile/PerTargetPropagater.java
```diff
@@ -0,0 +1,107 @@
...
+            Arrays.fill(perIterationResults, nonTransferTravelTimesToTargets[targetIdx] < cutoffSeconds);
+
+            if (perIterationResults[0]) {
```
abyrd: Feels a bit weird to be testing the zeroth value in an array to get the value you just filled it with. I'd prefer to calculate `boolean reachableWithoutTransit = nonTransferTravelTimesToTargets[targetIdx] < cutoffSeconds;` then both fill the array and conditionally `continue;` based on the value of that boolean.

src/main/java/com/conveyal/r5/profile/PerTargetPropagater.java
```diff
@@ -0,0 +1,107 @@
...
+                // reachable on foot, no need to even check if it's reachable via transit
+                reducer.accept(targetIdx, perIterationResults);
+                continue;
+            }
+
+            TIntIntMap pointToStopDistanceTable = targets.pointToStopDistanceTables.get(targetIdx);
+
+            // effectively final nonsense
```
abyrd: Elaborate in comment: Java can only close lambdas over effectively final variables, and an array of one value meets that criterion while still allowing the value to be modified.

It's too bad Trove maps don't have a method that allows you to map-reduce all their entries (pass each pair of entries through a function and collect the results like streams).

src/main/java/com/conveyal/r5/profile/PerTargetPropagater.java
```diff
@@ -0,0 +1,107 @@
...
+            boolean[] targetEverReached = new boolean[] { nonTransferTravelTimesToTargets[targetIdx] <= cutoffSeconds };
+
+            // don't try to propagate transit if there are no nearby transit stops,
+            // but still call the reducer below with the non-transit times, because you can walk even where there is no
+            // transit
+            if (pointToStopDistanceTable != null) {
+                for (int iteration = 0; iteration < perIterationResults.length; iteration++) {
+                    final int effectivelyFinalIteration = iteration;
+                    pointToStopDistanceTable.forEachEntry((stop, distanceMm) -> {
```
abyrd: Mm would be mega-meters, which is why I break with camelCase variable naming convention and put underscores before unit names (to preserve capitalization) or spell out the unit prefixes.

src/main/java/com/conveyal/r5/profile/PerTargetPropagater.java
```diff
@@ -0,0 +1,107 @@
...
+                        int timeAtStop = travelTimesToStopsEachIteration[effectivelyFinalIteration][stop];
+
+                        if (timeAtStop > cutoffSeconds) return true; // avoid overflow
+
+                        int timeAtTargetThisStop = timeAtStop + distanceMm / speedMillimetersPerSecond;
+
+                        if (timeAtTargetThisStop < cutoffSeconds) {
+                            perIterationResults[effectivelyFinalIteration] = true;
+                            targetEverReached[0] = true;
+                            return false; // stop iteration over stops
```
abyrd: Comment: mention that this is an optimization, that once we know the target can be reached from any stop in less than the time threshold, we can break out, since we only care about reachability rather than the list of specific travel times.

src/main/java/com/conveyal/r5/profile/Propagater.java
```diff
@@ -25,17 +25,21 @@
     /** Times at targets using the street network */
     public final int[] nonTransferTravelTimesToTargets;
 
+    /** The maximum allowable travel time in this regional analysis, to avoid propagating things that won't be useful */
+    public final int cutoffSeconds;
```
abyrd: The expression "maximum allowable travel time" is not clear. At first I took this to mean the maximum representable or displayable travel time, but reading the code I think this refers to the accessibility indicator travel time threshold value. 

The phrase "propagating things that won't be useful" could be clarified as well. "Things" means "times", and I wouldn't say those times "won't be useful", it's more that the action of propagation is unnecessary, because propagation can only increase travel time, and our end result is just whether travel times are over or under the threshold, not the travel times themselves. This allows an optimization.

src/main/java/com/conveyal/r5/profile/Propagater.java
```diff
@@ -56,8 +60,12 @@ public Propagater (int[][] travelTimesToStopsEachIteration, int[] nonTransferTra
         // iteration to the next, but that requires working on clock times, so we'd have to pass more information down
         // into this function; I'm not sure it's worth it.
         int[] result = new int[travelTimesToStopsEachIteration.length];
+        int[] travelTimesToTargetsThisIteration = new int[nonTransferTravelTimesToTargets.length];
```
abyrd: This would be more readable if we pulled nIterations and nTargets out into named final int variables and used those to specify the array dimensions. Likewise below in the loop termination conditions.

src/main/java/com/conveyal/r5/profile/Propagater.java
```diff
@@ -70,7 +78,9 @@ public Propagater (int[][] travelTimesToStopsEachIteration, int[] nonTransferTra
             for (int stop = 0; stop < travelTimesToStopsThisIteration.length; stop++) {
                 int travelTimeToStop = travelTimesToStopsThisIteration[stop];
 
-                if (travelTimeToStop == RaptorWorker.UNREACHED) continue;
+                // don't bother to even propagate stops that take too long to get to
+                // this will also skip unreached stops
+                if (travelTimeToStop > cutoffSeconds) continue;
```
abyrd: If I understand correctly, this means that the Propagater can no longer be used to get all the travel times, and like the PerTargetPropagater is no longer in any way generic: it only makes sense for reducer functions with very specific properties (ones that have no effect when all travel times are above the cutoff). If it is not generic, why does it accept reducer function references instead of just defining them as named methods on the Propagator class?

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
```diff
@@ -71,6 +71,9 @@
     /** For each transit stop, the distances to nearby PointSet points as packed (point_index, distance) pairs. */
     public List<int[]> stopToPointDistanceTables;
 
+    /** For each pointset point, the reachable stops, as a map from StopID to distance in millimeters */
```
abyrd: "the stops reachable from that point on foot"

src/main/java/com/conveyal/r5/analyst/ExtractingGridStatisticComputer.java
```diff
@@ -0,0 +1,67 @@
...
+import com.google.common.io.LittleEndianDataInputStream;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.zip.GZIPInputStream;
+
+/**
+ * Extract the nth sample from a grid. Mostly useful when the grid contains a distribution rather than samples.
+ */
+public class ExtractingGridStatisticComputer {
```
mattwigway: There are several other classes that have similar functions in that they compute single value grids, but they have different method signatures (cf. . We could make a superinterface and then put the extra arguments in a constructor, but that seems needlessly dogmatic for something that can easily be done with a static class. It computes a whole grid, not a single value, which is why I termed it a GridComputer and not a SampleExtractor. It seems like that would be just semantics because if we then renamed the AndrewOwenMeanGridStatisticComputer an AndrewOwenGridSampleExtractor we'd have the same problem.

src/main/java/com/conveyal/r5/analyst/ExtractingGridStatisticComputer.java
```diff
@@ -0,0 +1,67 @@
...
+
+    /** Version of the access grid format we read */
+    private static final int ACCESS_GRID_VERSION = 0;
+
+    public static Grid compute (String bucket, String key, int index) throws IOException {
+        S3Object accessGrid = s3.getObject(bucket, key);
+
+        LittleEndianDataInputStream input = new LittleEndianDataInputStream(new GZIPInputStream(accessGrid.getObjectContent()));
+
+        char[] header = new char[8];
```
mattwigway: Good point, this could move to a superclass.

mattwigway: Done

src/main/java/com/conveyal/r5/analyst/ExtractingGridStatisticComputer.java
mattwigway: Disregard previous comment, based on the duplicated header code it made sense to make a super class.

src/main/java/com/conveyal/r5/analyst/ExtractingGridStatisticComputer.java
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        int zoom = input.readInt();
+        int west = input.readInt();
+        int north = input.readInt();
+        int width = input.readInt();
+        int height = input.readInt();
+        int nIterations = input.readInt();
+
+        if (index >= nIterations || index < 0) throw new ArrayIndexOutOfBoundsException("Extraction parameter out of bounds");
```
mattwigway: This code no longer exists to to refactor after previous comments.

src/main/java/com/conveyal/r5/analyst/ExtractingGridStatisticComputer.java
```diff
@@ -0,0 +1,67 @@
...
+
+        Grid grid = new Grid(zoom, width, height, north, west);
```
mattwigway: Done.

src/main/java/com/conveyal/r5/analyst/ExtractingGridStatisticComputer.java
```diff
@@ -0,0 +1,67 @@
...
+        if (version != ACCESS_GRID_VERSION) {
+            throw new IllegalArgumentException(String.format("Version mismatch of access grids, expected %s, found %s", ACCESS_GRID_VERSION, version));
+        }
+
+        int zoom = input.readInt();
+        int west = input.readInt();
+        int north = input.readInt();
+        int width = input.readInt();
+        int height = input.readInt();
+        int nIterations = input.readInt();
```
mattwigway: Done.

src/main/java/com/conveyal/r5/analyst/ExtractingGridStatisticComputer.java
```diff
@@ -0,0 +1,67 @@
...
+
+        if (index >= nIterations || index < 0) throw new ArrayIndexOutOfBoundsException("Extraction parameter out of bounds");
+
+        Grid grid = new Grid(zoom, width, height, north, west);
+
+        for (int y = 0; y < height; y++) {
+            for (int x = 0; x < width; x++) {
+                int val = 0;
+                for (int iteration = 0; iteration <= index; iteration++) {
+                    val += input.readInt();
```
mattwigway: This is because the input data is delta coded; I'll add a comment.

src/main/java/com/conveyal/r5/analyst/GridComputer.java
```diff
@@ -17,29 +18,41 @@
...
- * separately for each iteration of the range RAPTOR algorithm (each departure minute and randomization of the schedules
- * for the frequency-based routes) and all of these different values of the indicator are retained to allow
- * probabilistic scenario comparison. This does freeze the travel time cutoff and destination grid in the interest of
- * keeping the results down to an acceptable size. The results are placed on an Amazon SQS queue for collation by
- * a GridResultConsumer and a GridResultAssembler.
+ * a Web Mercator density grid. Both grids must be at the same zoom level. The accessibility is calculated using
+ * median travel time (there are plans to change this to allow use of arbitrary percentiles in the future). In order
+ * to facilitate probabilistic comparison of scenarios, many accessibility values are returned, representing the
+ * sampling distribution of the accessibility. The first value is the value computed using all the Monte Carlo frequency
+ * draws with equal weight. Subsequent values are produced by bootstrapping the Monte Carlo iterations to simulate
```
mattwigway: Better?

abyrd:
@mattwigway I see you made some changes. Do you plan to revise the comments further? Ping me when you consider your revisions complete.

mattwigway:
Yes, I ran out of time on Friday so I'll complete then tomorrow.

abyrd:
No problem @mattwigway, just checking. I think we can do the release as soon as this one's merged though.

src/main/java/com/conveyal/r5/analyst/GridComputer.java
```diff
@@ -106,41 +119,107 @@ public void run() throws IOException {
...
+                    }
+                }
+            } else if (foundReachable && !foundUnreachable) {
+                // this destination is always reachable and will be included in all bootstrap samples, no need to do the
+                // bootstrapping
+                for (int i = 0; i < samples.length; i++) samples[i] += grid.grid[gridx][gridy];
+            } // otherwise, this destination is never reachable, no need to do bootstrapping or increment samples
         });
 
+        int[] intSamples = DoubleStream.of(samples).mapToInt(d -> (int) Math.round(d)).toArray();
```
mattwigway: I think I'm going to leave this and add a comment because this way it's explicit that we're rounding rather than casting/flooring.

src/main/java/com/conveyal/r5/analyst/cluster/GridResultAssembler.java
```diff
@@ -135,17 +132,17 @@ public void handleMessage (Message message) {
             Origin origin = Origin.read(bais);
 
             // if this is not the first request, make sure that we have the correct number of iterations
-            if (buffer != null && origin.accessibilityPerIteration.length != this.nIterations) {
+            if (buffer != null && origin.samples.length != this.nIterations) {
```
mattwigway: I updated the comment to match "samples" but I think samples is a good term since these can be either instantaneous accessibility values or bootstrapped values, depending on the version of r5.

src/main/java/com/conveyal/r5/profile/Propagater.java
mattwigway: That's correct, however the Propagater class is now unused so I've removed it.

src/main/java/com/conveyal/r5/profile/PerTargetPropagater.java
```diff
@@ -0,0 +1,107 @@
...
+
+import com.conveyal.r5.streets.LinkedPointSet;
+import gnu.trove.map.TIntIntMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Arrays;
+
+/**
+ * This is similar to the Propagater class, but instead of the reducer function being called with each iteration,
```
mattwigway: The reason I pass it as a function is so that the times can be dealt with and summarized in GridComputer; the array of times is too large to hold in memory all at once, so it needs to be dealt with piecemeal. A member function of PerTargetPropagater would not have access to the variables in GridComputer (unless we made them public or package-private class fields and passed in the GridComputer instance, but I'm not sure that's better).

src/main/java/com/conveyal/r5/analyst/cluster/GridResultAssembler.java
abyrd: This updated version is fine since it provides more detail, I was more concerned about the word being used alone.

PR #265, opened 2017-04-15
Remove the custom AMI requirement

mattwigway:
This just passes in a user data and can be used with a standard Amazon Linux AMI, e.g. `ami-e5083683` in `eu-west-1`.

H/T @dfconway for helping with the shell scripting.

XKCD

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/broker/Broker.java
+++ b/src/main/java/com/conveyal/r5/analyst/broker/Broker.java
@@ -384,7 +384,7 @@ public class Broker implements Runnable {
 
         // Read in the startup script
         try {
-            InputStream scriptIs = this.getClass().getClassLoader().getResourceAsStream("worker.sh");
+            InputStream scriptIs = this.getClass().getClassLoader().getResourceAsStream("/worker.sh");
             ByteArrayOutputStream scriptBaos = new ByteArrayOutputStream();
             ByteStreams.copy(scriptIs, scriptBaos);
             scriptIs.close();
--- a/src/main/java/com/conveyal/r5/analyst/broker/Broker.java
+++ b/src/main/java/com/conveyal/r5/analyst/broker/Broker.java
@@ -384,7 +384,7 @@ public class Broker implements Runnable {
 
         // Read in the startup script
         try {
-            InputStream scriptIs = this.getClass().getClassLoader().getResourceAsStream("/worker.sh");
+            InputStream scriptIs = Broker.class.getClassLoader().getResourceAsStream("worker.sh");
             ByteArrayOutputStream scriptBaos = new ByteArrayOutputStream();
             ByteStreams.copy(scriptIs, scriptBaos);
             scriptIs.close();
--- a/src/main/java/com/conveyal/r5/analyst/broker/Broker.java
+++ b/src/main/java/com/conveyal/r5/analyst/broker/Broker.java
@@ -12,6 +12,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.TreeMultimap;
+import com.google.common.io.ByteStreams;
 import gnu.trove.map.TIntObjectMap;
 import gnu.trove.map.TObjectLongMap;
 import gnu.trove.map.hash.TIntObjectHashMap;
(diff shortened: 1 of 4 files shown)


src/main/java/com/conveyal/r5/analyst/broker/Broker.java
```diff
@@ -380,8 +379,27 @@ public void createWorkersInCategory (WorkerCategory category) {
             throw new RuntimeException(e);
         }
 
+        String workerConfigString = cfg.toString();
+        String scriptTemplate;
```
abyrd: This variable leaks out of the only scope it is used in, across a try block. Personally I'd probably put the variable declaration and everything up to req.setUserData(userData); inside the try block.

src/main/java/com/conveyal/r5/analyst/broker/Broker.java
```diff
@@ -380,8 +379,27 @@ public void createWorkersInCategory (WorkerCategory category) {
...
+            scriptIs.close();
+            scriptBaos.close();
+            scriptTemplate = scriptBaos.toString();
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+
+        String logGroup = workerConfig.getProperty("log-group");
+
+        String script = MessageFormat.format(scriptTemplate, workerDownloadUrl, logGroup, workerConfigString);
```
abyrd: This MessageFormat seems to be intended for internationalized user messages, and the copying of streams into byte arrays etc. is a bit strange, but maybe there's no better way to do this. Positionally filling in the fields is a little brittle, but there are only four of them so I guess it'll do the job.

src/main/java/com/conveyal/r5/profile/ProfileRequest.java
```diff
@@ -236,6 +237,7 @@ This parameter compensates for the fact that GTFS does not contain information a
 
     //If true current search is reverse search AKA we are looking for a path from destination to origin in reverse
     //It differs from searchType because it is used as egress search
+    @JsonInclude(JsonInclude.Include.NON_DEFAULT)
```
abyrd: Is this change related to worker startup? If not can it be separated out of the PR?

src/main/resources/worker.sh
```diff
@@ -0,0 +1,77 @@
...
+# first things first: set up logging
+LOGFILE=/var/log/analyst-worker.log
+
+echo Starting analyst worker at `date` > $LOGFILE
+
+# make it so that the worker can write to the logfile
+chown ec2-user:ec2-user $LOGFILE
+chmod 664 $LOGFILE # Log agent needs to read log file
+
+cat > /etc/awslogs/awslogs.conf <<EOF
```
abyrd: It would be good to have a short comment mentioning "heredocs" to save someone in the future confusion when they've never seen this before.

src/main/resources/worker.sh
```diff
@@ -0,0 +1,77 @@
+#!/bin/bash
+# Downloads and runs an analyst worker.
+# This will be run through Java MessageFormat, and the following items will be available by wrapping
+# the relevant index in curly braces
```
abyrd: I'm not sure someone reading this comment will understand "Available by wrapping the index in curly braces". How about this:
This shell script will undergo variable substitution via Java's MessageFormat class before it is run on newly started worker machines. MessageFormat will replace special tokens (consisting of numbers inside curly braces) with configuration information specific to the worker being started. These are:
0
1
2

src/main/java/com/conveyal/r5/analyst/broker/Broker.java
```diff
@@ -380,8 +379,27 @@ public void createWorkersInCategory (WorkerCategory category) {
             throw new RuntimeException(e);
         }
 
+        String workerConfigString = cfg.toString();
+        String scriptTemplate;
+
+        // Read in the startup script
```
abyrd: It would be good to have a more extensive comment here explaining why we do this (to avoid having a special Amazon Machine Image, so we can use a stock Amazon Linux image and have it configure itself).

src/main/java/com/conveyal/r5/profile/ProfileRequest.java
mattwigway: It's not. I'll cherry-pick into a fresh PR.

mattwigway:
@abyrd this should be good to go, let me know if there's anything else, and see also #267.

abyrd:
Looks good, thanks for the modifications @mattwigway

Issue #264, opened 2017-04-14
Workers fail on unknown profilerequest properties


mattwigway:
I have no idea why this is the case, but workers are still failing every time we add a new property to profilerequest, even though we are using a special ObjectMapper to deserialize profile requests that is not supposed to fail on unknown properties it still does.

PR #267, opened 2017-04-20
don't include reverse routing field if it is at its default value, to…

mattwigway:
… avoid confusing older workers.

Due to #264, adding fields to the profile request makes older workers blow up with recent brokers. Until we can fix this issue we need to avoid transmitting new values unless they have been modified.


mattwigway:
Split from #265

PR #273, opened 2017-05-11
Misc. Fixes for Arlington

demory:
This PR includes:
- Matt's `fix-modeify-time-estimates` branch
- A fix for `P2::equals()` in the above branch
- The `fix-fare-table` branch which allows for ignoring the AgencyID in comparisons


src/main/java/com/conveyal/r5/transit/fare/FareTable.java
```diff
@@ -47,7 +54,9 @@ public Fare lookup (String from, String to) {
 
     public Fare lookup (Stop from, Stop to) {
         //TODO: how does this works WRT different GTFS feeds having Stops with same Ids?
-        return lookup(from.stopId, to.stopId);
+        return this.ignoreAgencyId
+                ? lookup(from.stopId.split(":")[1], to.stopId.split(":")[1])
```
mattwigway: I'm wondering if we should use a regex here so we can catch everything after the first colon, in case a stop ID contains a colon (so you'd have a feed-qualified stop id of, say, `WMATA:GalleryPlace:LowerLevel`), where the stopId is `GalleryPlace:LowerLevel`.

abyrd: Absolutely, we do need to avoid splitting on colons in the ID. But we don't need to manually construct a regex. The split function has a limit parameter to limit the number of splits. stopId.split(":", 2) should do it.

abyrd:
I just fixed the split() call and will now merge this.

PR #274, opened 2017-05-23
New analysis interface

mattwigway:
This allows us to get travel times as Access Grids rather than only in Browsochrones format.

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/GridBounds.java
+++ b/src/main/java/com/conveyal/r5/analyst/GridBounds.java
@@ -1,0 +1,8 @@
+package com.conveyal.r5.analyst;
+
+/**
+ * We have a lot of classes which have zoom, west, north, width, and height attributes. We would like to be able to pass
+ * them into functions without breaking out the values of these.
+ */
+public class GridBounds {
+}
--- a/src/main/java/com/conveyal/r5/analyst/TravelTimeSurfaceComputer.java
+++ b/src/main/java/com/conveyal/r5/analyst/TravelTimeSurfaceComputer.java
@@ -2,7 +2,9 @@ package com.conveyal.r5.analyst;
 
 import com.conveyal.r5.analyst.cluster.AccessGridWriter;
 import com.conveyal.r5.analyst.cluster.TravelTimeSurfaceRequest;
+import com.conveyal.r5.analyst.error.TaskError;
 import com.conveyal.r5.api.util.LegMode;
+import com.conveyal.r5.common.JsonUtilities;
 import com.conveyal.r5.profile.FastRaptorWorker;
 import com.conveyal.r5.profile.PerTargetPropagater;
 import com.conveyal.r5.profile.RaptorWorker;
@@ -20,6 +22,7 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.stream.IntStream;
 
 /**
@@ -146,6 +149,16 @@ public class TravelTimeSurfaceComputer {
         }
 
         os.write(output.getBytes());
+
+        // Append scenario application warning JSON to result
+        ResultMetadata metadata = new ResultMetadata();
+        metadata.scenarioApplicationWarnings = network.scenarioApplicationWarnings;
+        JsonUtilities.objectMapper.writeValue(os, metadata);
+
         os.close();
+    }
+
+    private static class ResultMetadata {
+        public Collection<TaskError> scenarioApplicationWarnings;
     }
 }
--- a/src/main/java/com/conveyal/r5/analyst/TravelTimeSurfaceComputer.java
+++ b/src/main/java/com/conveyal/r5/analyst/TravelTimeSurfaceComputer.java
@@ -9,13 +9,11 @@ import com.conveyal.r5.profile.FastRaptorWorker;
 import com.conveyal.r5.profile.PerTargetPropagater;
 import com.conveyal.r5.profile.RaptorWorker;
 import com.conveyal.r5.profile.StreetMode;
-import com.conveyal.r5.publish.StaticComputer;
 import com.conveyal.r5.streets.LinkedPointSet;
 import com.conveyal.r5.streets.StreetRouter;
 import com.conveyal.r5.transit.TransportNetwork;
 import gnu.trove.iterator.TIntIntIterator;
 import gnu.trove.map.TIntIntMap;
-import org.apache.commons.math3.util.FastMath;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -26,13 +24,12 @@ import java.util.Collection;
 import java.util.stream.IntStream;
 
 /**
- * This computes a travel time surface and returns it in access grid format, with one travel time for each minute/MC draw.
+ * This computes a travel time surface and returns it in access grid format, with each pixel of the grid containing
+ * different percentiles of travel time requested by the frontend.
  */
 public class TravelTimeSurfaceComputer {
     private static final Logger LOG = LoggerFactory.getLogger(TravelTimeSurfaceComputer.class);
     private static final WebMercatorGridPointSetCache pointSetCache = new WebMercatorGridPointSetCache();
-
-    public static final int PERCENTILE_STEP = 5;
 
     public final TravelTimeSurfaceRequest request;
     public final TransportNetwork network;
@@ -43,16 +40,16 @@ public class TravelTimeSurfaceComputer {
     }
 
     public void write (OutputStream os) throws IOException {
-        StreetMode accessMode = LegMode.legModeSetToDominantStreetMode(request.request.accessModes);
-        StreetMode directMode = LegMode.legModeSetToDominantStreetMode(request.request.directModes);
+        StreetMode accessMode = LegMode.getDominantStreetMode(request.request.accessModes);
+        StreetMode directMode = LegMode.getDominantStreetMode(request.request.directModes);
 
-        int nSamples = 100 / PERCENTILE_STEP - 1;
+        int nPercentiles = request.percentiles.length;
 
         WebMercatorGridPointSet destinations = pointSetCache.get(request.zoom, request.west, request.north, request.width, request.height);
 
         AccessGridWriter output;
         try {
-            output = new AccessGridWriter(request.zoom, request.west, request.north, request.width, request.height, nSamples);
+            output = new AccessGridWriter(request.zoom, request.west, request.north, request.width, request.height, nPercentiles);
         } catch (IOException e) {
             // in memory, should not be able to throw this
             throw new RuntimeException(e);
@@ -82,7 +79,7 @@ public class TravelTimeSurfaceComputer {
                 // the frontend expects percentiles of travel time. There is no variation in nontransit travel time so
                 // just replicate the same number repeatedly. This could be improved, but at least it will compress well.
                 // int divide (floor) used below as well. TODO is this wise?
-                int[] results = IntStream.range(0, nSamples).map(i -> travelTimeMinutes).toArray();
+                int[] results = IntStream.range(0, nPercentiles).map(i -> travelTimeMinutes).toArray();
                 try {
                     output.writePixel(x, y, results);
                 } catch (IOException e) {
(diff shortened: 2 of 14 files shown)


mattwigway:
@abyrd I think this should be good to go! Let me know if you have any comments.

src/main/java/com/conveyal/r5/analyst/TravelTimeSurfaceComputer.java
```diff
@@ -0,0 +1,179 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.stream.IntStream;
+
+/**
+ * This computes a travel time surface and returns it in access grid format, with one travel time for each minute/MC draw.
```
abyrd: It looks like this Javadoc is incorrect - this class returns a grid with the third dimension representing percentiles of travel time at each destination, in increasing order.

src/main/java/com/conveyal/r5/analyst/TravelTimeSurfaceComputer.java
```diff
@@ -0,0 +1,179 @@
...
+    public TravelTimeSurfaceComputer (TravelTimeSurfaceRequest request, TransportNetwork network) {
+        this.request = request;
+        this.network = network;
+    }
+
+    public void write (OutputStream os) throws IOException {
+        StreetMode accessMode = LegMode.legModeSetToDominantStreetMode(request.request.accessModes);
+        StreetMode directMode = LegMode.legModeSetToDominantStreetMode(request.request.directModes);
+
+        int nSamples = 100 / PERCENTILE_STEP - 1;
```
abyrd: Since this always returns percentiles, I'd be in favor of changing this variable name to nPercentiles.

src/main/java/com/conveyal/r5/analyst/TravelTimeSurfaceComputer.java
```diff
@@ -0,0 +1,179 @@
...
+            for (int target = 0; target < travelTimesToTargets.length; target++) {
+                int x = target % request.width;
+                int y = target / request.width;
+
+                final int travelTimeMinutes =
+                        travelTimesToTargets[target] == RaptorWorker.UNREACHED ? RaptorWorker.UNREACHED : travelTimesToTargets[target] / 60;
+                // the frontend expects percentiles of travel time. There is no variation in nontransit travel time so
+                // just replicate the same number repeatedly. This could be improved, but at least it will compress well.
+                // int divide (floor) used below as well. TODO is this wise?
+                int[] results = IntStream.range(0, nSamples).map(i -> travelTimeMinutes).toArray();
```
abyrd: An access grid already announces its z-depth in the last integer in its header, so it should be pretty easy to flatten it to a depth of 1.

src/main/java/com/conveyal/r5/analyst/TravelTimeSurfaceComputer.java
```diff
@@ -0,0 +1,179 @@
...
+            sr.profileRequest = request.request;
+            sr.distanceLimitMeters = 2000; // TODO hardwired same as gridcomputer
+            sr.setOrigin(request.request.fromLat, request.request.fromLon);
+            sr.dominanceVariable = StreetRouter.State.RoutingVariable.DISTANCE_MILLIMETERS;
+            sr.route();
+
+            // Get the travel times to all stops reached in the initial on-street search. Convert distances to speeds.
+            int offstreetTravelSpeedMillimetersPerSecond = (int) (request.request.getSpeed(accessMode) * 1000);
+
+            TIntIntMap accessTimes = sr.getReachedStops();
```
abyrd: This map does not contain access times. The dominance variable is millimeters, so this contains access distances in millimeters.

src/main/java/com/conveyal/r5/analyst/TravelTimeSurfaceComputer.java
```diff
@@ -0,0 +1,179 @@
...
+
+            // find non-transit times
+            int[] nonTransitTravelTimesToDestinations = linkedDestinationsAccess
+                    .eval(sr::getTravelTimeToVertex, offstreetTravelSpeedMillimetersPerSecond)
+                    .travelTimes;
+
+            PerTargetPropagater perTargetPropagater = new PerTargetPropagater(transitTravelTimesToStops, nonTransitTravelTimesToDestinations, linkedDestinationsEgress, request.request, 120 * 60);
+            perTargetPropagater.propagateTimes((target, times) -> {
+                Arrays.sort(times);
+                int[] results = new int[nSamples];
```
abyrd: Again, I think the variable name should be changed and there should be some comment here explaining we're reading off percentiles and not reporting the entire list of travel times.

src/main/java/com/conveyal/r5/analyst/cluster/AccessGridWriter.java
```diff
@@ -0,0 +1,154 @@
...
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.RandomAccessFile;
+import java.nio.ByteBuffer;
+
+/**
+ * Random-access access grid writer - write arrays into offsets in an access grid format file (3-d array)
+ *  * Access grids look like this:
+ * Header (ASCII text "ACCESSGR") (note that this header is eight bytes, so the full grid can be mapped into a
```
abyrd: I would say this entire list of 4-byte ints is the header, and ACCESSGR is just the "magic numbers" in the header.

src/main/java/com/conveyal/r5/analyst/cluster/AccessGridWriter.java
```diff
@@ -0,0 +1,154 @@
...
+    /** The version of the origins we consume */
+    public static final int ORIGIN_VERSION = 0;
+
+    /** The offset to get to the data section of the access grid file */
+    public static final long HEADER_SIZE = 9 * 4;
+
+    private BufferAbstraction buffer;
+
+    // store as longs so we can use with impunity without fear of overflow
+    private final long zoom, west, north, width, height, nSamples;
```
abyrd: It seems like we should be calling this "depth" especially since now we're using for things other than samples of travel time.

src/main/java/com/conveyal/r5/analyst/cluster/TravelTimeSurfaceRequest.java
```diff
@@ -0,0 +1,33 @@
+package com.conveyal.r5.analyst.cluster;
+
+import com.conveyal.r5.profile.ProfileRequest;
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+/**
+ * Request a travel time surface containing travel times to all destinations for all iterations.
```
abyrd: Javadoc is no longer correct - this does not return travel times for all iterations, it returns a fixed number of percentiles.

src/main/java/com/conveyal/r5/analyst/cluster/TravelTimeSurfaceRequest.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+public class TravelTimeSurfaceRequest extends GenericClusterRequest {
+    /** The profile request to use */
+    public ProfileRequest request;
+
+    public int zoom;
+    public int west;
+    public int north;
+    public int width;
+    public int height;
```
abyrd: It might be useful to allow the requester to specify the number of percentiles, considering that we usually only show 5 of them.

src/main/java/com/conveyal/r5/api/util/LegMode.java
```diff
@@ -9,5 +13,12 @@
     //Renting a bicycle
     BICYCLE_RENT,
     //Park & Ride
-    CAR_PARK
+    CAR_PARK;
+
+    /** Return the heaviest/fastest StreetMode for use in stop finding */
+    public static StreetMode legModeSetToDominantStreetMode (Set<LegMode> modes) {
```
abyrd: the expression "set to" sounds like a value is being assigned to a variable ("I set this variable to X"). I don't think the name of the function needs to include the type of its parameter, since that is visible from the signature. I'd just call this "getDominantStreetMode".

src/main/java/com/conveyal/r5/profile/PerTargetPropagater.java
```diff
@@ -35,15 +35,16 @@
...
         this.travelTimesToStopsEachIteration = travelTimesToStopsEachIteration;
-        this.nonTransferTravelTimesToTargets = nonTransferTravelTimesToTargets;
+        this.nonTransitTravelTimesToTargets = nonTransitTravelTimesToTargets;
         this.targets = targets;
         this.request = request;
         this.cutoffSeconds = cutoffSeconds;
     }
 
-    public void propagate (Reducer reducer) {
+    private void propagate (Reducer reducer, TravelTimeReducer travelTimeReducer) {
```
abyrd: This signature is strange, where only one of the two parameters should be passed in and one overrides the other. If there's no other reasonable way to do this, there should at least be Javadoc explaining which parameter takes priority over the other, and why there are two different ones. The more standard way would be to overload the propagate() name with two different parameter types but I guess that just pushes the problem farther down the stack.

abyrd: Oh hey, you already did that below. Nice. But a line of Javadoc explaining this is the internal interface for those two external interfaces and which parameter takes priority would be good.

src/main/java/com/conveyal/r5/analyst/GridBounds.java
```diff
@@ -0,0 +1,8 @@
+package com.conveyal.r5.analyst;
+
+/**
+ * We have a lot of classes which have zoom, west, north, width, and height attributes. We would like to be able to pass
+ * them into functions without breaking out the values of these.
+ */
+public class GridBounds {
```
abyrd: It seems like this is supposed to be an interface allowing the same methods to be called on several classes, but it doesn't define any abstract methods or fields.

src/main/java/com/conveyal/r5/analyst/TravelTimeSurfaceComputer.java
mattwigway: Yes, but there is code in the frontend that depends on there being 20 percentiles, changing that is a pretty major project, I estimate it at at least a day. I can do it but probably not for at least a week and possibly not until I return from Australia.

src/main/java/com/conveyal/r5/analyst/TravelTimeSurfaceComputer.java
mattwigway: It does actually contain times, we get out distances and then convert them in place. I added a comment.

src/main/java/com/conveyal/r5/analyst/cluster/AccessGridWriter.java
mattwigway: You can look at it that way, but ACCESSGR was chosen for a reason, and it is actually written as ASCII text.

src/main/java/com/conveyal/r5/analyst/cluster/AccessGridWriter.java
mattwigway: I find depth misleading as well, so I used `nValuesPerPixel`

src/main/java/com/conveyal/r5/analyst/cluster/TravelTimeSurfaceRequest.java
mattwigway: Good point, fixed.

src/main/java/com/conveyal/r5/analyst/TravelTimeSurfaceComputer.java
mattwigway: Based on responses to your code below, the frontend now requests the number of percentiles it wants, and this code simply returns that number of percentiles. I'd rather do that than have special code on the frontend to handle the case where there is no variation in travel time.

mattwigway:
@abyrd I think I've addressed all the comments here.

src/main/java/com/conveyal/r5/analyst/cluster/AccessGridWriter.java
abyrd: I just meant that the term "header" (used in the javadoc) usually refers to the entire block of meta-info that precedes the main repeating block of data in a file. So those letters ACCESSGR are just the first element in the header (as I see it).

abyrd:
Great, thanks @mattwigway. I think we can merge this now.

PR #280, opened 2017-06-16
Park ride analyst support

buma:
I added changes but I can't test.

I'm not sure what to do with sr::getTravelTimeToVertex, since function returns time, but it seems that distance is expected.

ParkRideRouter returns travel time only to stops not to all vertices, since it doesn't actually route from P+R to stops. Because stops are found during graph build time.

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/GridComputer.java
+++ b/src/main/java/com/conveyal/r5/analyst/GridComputer.java
@@ -7,6 +7,7 @@ import com.amazonaws.services.sqs.model.SendMessageRequest;
 import com.conveyal.r5.analyst.cluster.GridRequest;
 import com.conveyal.r5.analyst.cluster.Origin;
 import com.conveyal.r5.api.util.LegMode;
+import com.conveyal.r5.point_to_point.builder.PointToPointQuery;
 import com.conveyal.r5.profile.FastRaptorWorker;
 import com.conveyal.r5.profile.PerTargetPropagater;
 import com.conveyal.r5.profile.StreetMode;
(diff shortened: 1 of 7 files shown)


src/main/java/com/conveyal/r5/analyst/GridComputer.java
```diff
@@ -188,20 +189,32 @@ public void run() throws IOException {
```
buma: LinkedPointSet linkedDestinationsDirect = destinations.link(network.streetLayer, accessMode);
I'm not sure if this uses correct accessMode since currently if CAR_PARK is used accessMode is WALK.

src/main/java/com/conveyal/r5/analyst/GridComputer.java
```diff
@@ -188,20 +189,32 @@ public void run() throws IOException {
...
+                // time to stop is time from CAR streetrouter to stop + CAR PARK time + time to walk to stop based on request walk speed
+                //by default 20 CAR PARKS are found it can be changed with sr.maxVertices variable
+                sr = PointToPointQuery.findParkRidePath(request.request, sr, network.transitLayer);
+            } else {
+                sr.distanceLimitMeters = 2000; // TODO hardwired same as traveltimesurfacecomputer
+                sr.streetMode = accessMode;
+                sr.setOrigin(request.request.fromLat, request.request.fromLon);
+                sr.dominanceVariable = StreetRouter.State.RoutingVariable.DISTANCE_MILLIMETERS;
+                sr.route();
+            }
```
buma: What to do if findParkRidePath returns null which happens if setOrigin on streetRouter fails?

mattwigway:
@abyrd @buma I added some commits here, I think this should be ready to go now.

src/main/java/com/conveyal/r5/analyst/GridComputer.java
```diff
@@ -180,31 +183,82 @@ public void run() throws IOException {
...
+                    // TODO this is ugly. we should have a way to break out of the search early (here and in other methods).
+                    accessTimes = new TIntIntHashMap();
+                } else {
+                    accessTimes = sr.getReachedStops();
+                }
+
+                // disallow non-transit access
+                // TODO should we allow non transit access with park and ride?
+                nonTransitTravelTimesToDestinations = new int[linkedDestinationsAccess.size()];
+                Arrays.fill(nonTransitTravelTimesToDestinations, RaptorWorker.UNREACHED);
```
buma: It seems that nonTransitTravelTimesToDestinations is filled with RaptorWorker.UNREACHED and then used in propagator where it is just read if P+R is used. In other modes it is filled with values based on states.

src/main/java/com/conveyal/r5/analyst/TravelTimeSurfaceComputer.java
```diff
@@ -102,22 +104,71 @@ public void write (OutputStream os) throws IOException {
...
-            // getReachedStops returns distances, not times, so convert to times in the loop below
-            TIntIntMap accessTimes = sr.getReachedStops();
-            for (TIntIntIterator it = accessTimes.iterator(); it.hasNext(); ) {
-                it.advance();
-                // TODO how to handle kiss and ride/park and ride? Clearly this is not right.
-                it.setValue(it.value() / offstreetTravelSpeedMillimetersPerSecond);
+                // disallow non-transit access
+                // TODO should we allow non transit access with park and ride?
+                nonTransitTravelTimesToDestinations = new int[linkedDestinationsAccess.size()];
+                Arrays.fill(nonTransitTravelTimesToDestinations, RaptorWorker.UNREACHED);
```
buma: It seems that nonTransitTravelTimesToDestinations is filled with RaptorWorker.UNREACHED and then used in propagator where it is just read if P+R is used. In other modes it is filled with values based on states.

src/main/java/com/conveyal/r5/analyst/GridComputer.java
mattwigway: I did this because when you have a park and ride search I don't think we ever want to return non-transit times (because the non-transit times would be from driving, not park-and-ride; I don't think people requesting a park-and-ride journey want to see some journeys that are just driving all the way to the destination).

buma: Then it's ok

Issue #279, opened 2017-06-15
Make number of transport networks in TransportNetworkCache configurable


landonreed:
Currently [TransportNetworkCache.java](https://github.com/conveyal/r5/blob/master/src/main/java/com/conveyal/r5/transit/TransportNetworkCache.java) supports only one transport network.  When using r5 as a library (e.g., conveyal/datatools-server) we need support for multiple transport networks.  Perhaps the behavior should follow that of [BaseGTFSCache](https://github.com/conveyal/gtfs-lib/blob/master/src/main/java/com/conveyal/gtfs/BaseGTFSCache.java) and just accept as many networks as memory allows.

abyrd:
Even within Analysis workers we'll eventually want it to cache multiple networks. This is why the network cache interface exists, even though the actual implementation just holds on to the last network used.

However, it seems to me that many tasks we want to perform should be performed in isolation, i.e. completely separate R5 instances should be created, then tasks performed in sequence to limit memory consumption, much as we do with other tasks that are fed through the worker thread pool. In that case each R5 worker would only see one graph.

But maybe I'm not aware of the use case you're building for.

abyrd:
Also, unfortunately it's basically impossible within a JVM to fully understand how much memory you've got, and what the full memory cost of an object tree is.

landonreed:
This was addressed in #281 with the use of a loading cache and a configurable number of networks to cache (defaults to 1).

PR #281, opened 2017-06-16
Support multiple networks in TransportNetworkCache

landonreed:
Addresses #279.

This moves `TransportNetworkCache.scenarioNetworkCache` -> `TransportNetwork.scenarios` and adds a Guava loading cache with a configurable number of transport networks (defaults to 1).

This also adds support for a bucketFolder, so that r5 networks (+ scenarios and manifest files) can be stored in a "subdirectory" on s3.

### DIFF
--- a/src/main/java/com/conveyal/r5/transit/TransportNetwork.java
+++ b/src/main/java/com/conveyal/r5/transit/TransportNetwork.java
@@ -41,6 +41,12 @@ public class TransportNetwork implements Serializable {
     public TransitLayer transitLayer;
 
     /**
+     * This stores any number of lightweight scenario networks built upon the current base network.
+     * FIXME that sounds like a memory leak, should be a WeighingCache or at least size-limited.
+     */
+    public Map<String, TransportNetwork> scenarios = new HashMap<>();
+
+    /**
      * A grid point set that covers the full extent of this transport network. The PointSet itself then caches linkages
      * to street networks (the baseline street network, or ones with various scenarios applied). If they have been
      * created, this point set and its linkage to the street network are serialized along with the network, which makes
--- a/src/main/java/com/conveyal/r5/transit/TransportNetwork.java
+++ b/src/main/java/com/conveyal/r5/transit/TransportNetwork.java
@@ -44,7 +44,7 @@ public class TransportNetwork implements Serializable {
      * This stores any number of lightweight scenario networks built upon the current base network.
      * FIXME that sounds like a memory leak, should be a WeighingCache or at least size-limited.
      */
-    public Map<String, TransportNetwork> scenarios = new HashMap<>();
+    public transient Map<String, TransportNetwork> scenarios = new HashMap<>();
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/com/conveyal/r5/transit/TransportNetwork.java
```diff
@@ -41,6 +41,12 @@
     public TransitLayer transitLayer;
 
     /**
+     * This stores any number of lightweight scenario networks built upon the current base network.
+     * FIXME that sounds like a memory leak, should be a WeighingCache or at least size-limited.
+     */
+    public Map<String, TransportNetwork> scenarios = new HashMap<>();
```
mattwigway: This should be transient so that it's not serialized with the TransportNetwork.

src/main/java/com/conveyal/r5/transit/TransportNetworkCache.java
```diff
@@ -44,11 +46,13 @@
     private final File cacheDir;
 
     private final String sourceBucket;
+    private final String bucketFolder;
 
-    String currentNetworkId = null;
+//    String currentNetworkId = null;
```
mattwigway: We should just delete this.

src/main/java/com/conveyal/r5/transit/TransportNetworkCache.java
```diff
@@ -374,12 +405,50 @@ private TransportNetwork buildNetworkFromManifest (String networkId) {
...
+                for (String type : extensions) {
+                    File file = new File(cacheDir, id + type);
+                    file.delete();
+                }
+            }
+        };
+        return CacheBuilder.newBuilder()
+                .maximumSize(size)
+                .removalListener(removalListener)
+                .build(new CacheLoader() {
```
mattwigway: Can we use a lambda function here?

src/main/java/com/conveyal/r5/transit/TransportNetworkCache.java
```diff
@@ -374,12 +405,50 @@ private TransportNetwork buildNetworkFromManifest (String networkId) {
...
+                        // due to type erasure it can't be sure I'm using types correctly.
+                        return getNetwork((String) s);
+                    }
+                });
+    }
+
     public Set<String> getLoadedNetworkIds() {
-        if (currentNetwork == null) return Collections.emptySet();
-        else return Sets.newHashSet(currentNetwork.scenarioId);
+        if (cachedNetworks == null) return Collections.emptySet();
```
mattwigway: Is this possible? `createCache` gets called in the constructor, no?

src/main/java/com/conveyal/r5/transit/TransportNetworkCache.java
landonreed: Hmm, I couldn't quite figure out what lambda should go there.

mattwigway: Oh, right, I tried putting a lambda here and java is too stupid to figure out the types.

PR #282, opened 2017-06-19
fix issue with PerTargetPropagater that caused incorrect results.

mattwigway:
We should merge this and cut 2.2.1, and probably 2.2.0 from production since it gives wrong results. However, this is quite a bit slower and takes a really long time to propagate in the Netherlands (40s for the whole country).


mattwigway:
@abyrd this should be good to go.

abyrd:
Great comments, merging now and I'll cut a release.

Issue #290, opened 2017-06-28
Exact times causes crash with add-trips


mattwigway:
@marcovanderlinden reports that he is experiencing this crash with exact-times enabled on an add-trips modification:

![image](https://user-images.githubusercontent.com/566958/27633110-7f1dc366-5c30-11e7-86d5-a0c780c1445b.png)

It appears that the exact-times mode is not implemented for add trips, we need to add it.

abyrd:
Now we are getting crashes any time exact-times is false!

ansoncfit:
This issue rears its head again.  In a worker log:
```
09:16:25
java.lang.RuntimeException: This pattern ID specified in a scenario does not exist: 5b837a3f32b98e0ee389fa3d at com.conveyal.r5.profile.FrequencyRandomOffsets.randomize(FrequencyRandomOffsets.java:96) at com.conveyal.r5.profile.FastRaptorWorker.runRaptorForMinute(FastRaptorWorker.java:333) at com.conveyal.r5.profile.FastRaptorWorker.route(FastRaptorWorker.java:181) at com.conveyal.r5.analyst.T
09:16:25
401763 [qtp11050232-16] WARN com.conveyal.r5.analyst.cluster.AnalystWorker - Reporting errors in response to single-point request:
09:16:25
[com.conveyal.r5.analyst.error.TaskError@75c78448]
09:16:25
401765 [qtp11050232-16] INFO com.conveyal.r5.analyst.cluster.TimeGrid - Writing travel time surface with uncompressed size 9351 kiB
09:16:25
401893 [qtp11050232-16] INFO com.conveyal.r5.analyst.cluster.AnalystWorker - Travel time surface written, appending metadata with scenario application 1 warnings
09:16:25
401895 [qtp11050232-16] INFO com.conveyal.r5.analyst.cluster.AnalystWorker - Done writing
09:16:25
401898 [qtp11050232-16] INFO spark.http.matching.MatcherFilter - The requested route [/single] has not been mapped in Spark for Accept: [null]
```

ansoncfit:
Here's the modification:
![image](https://user-images.githubusercontent.com/2173529/44672335-2570d280-a9f6-11e8-9199-a7a9f71dbb12.png)


Note that the key in the above exception (`5b837a3f32b98e0ee389fa3d`) is the _id of the modification.  But I think R5 is looking for the source trip.

abyrd:
[...]

PR #291, opened 2017-06-28
Add exact times support to add trips (#290).

mattwigway:
This still needs an automated test.


mattwigway:
@abyrd this is ready to go.

PR #296, opened 2017-07-05
ReverseSearch field is now on StreetRouter

buma:
Instead of ProfileRequest

Since I wasn't able to make tests for change. I tested with comparing two saved requests.
Times, distances and paths are the same also for Bike rental and P+R.

### DIFF
--- a/src/main/java/com/conveyal/r5/api/ProfileResponse.java
+++ b/src/main/java/com/conveyal/r5/api/ProfileResponse.java
@@ -93,8 +93,6 @@ public class ProfileResponse {
             if (accessPathIndex < 0) {
                 //Here accessRouter needs to have this access mode since stopModeAccessMap is filled from accessRouter
                 StreetRouter streetRouter = accessRouter.get(accessMode);
-                //FIXME: Must we really update this on every streetrouter?
-                streetRouter.profileRequest.reverseSearch = false;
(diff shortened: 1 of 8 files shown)


src/main/java/com/conveyal/r5/api/ProfileResponse.java
```diff
@@ -180,10 +176,9 @@ public void generateStreetTransfers(TransportNetwork transportNetwork, ProfileRe
         Map<Integer, List<Transfer>> transfersWithSameStart = transferToOption.keySet().stream()
             .collect(Collectors.groupingBy(Transfer::getAlightStop));
         //LOG.info("Filling middle paths");
-        boolean prevReverseSearch = request.reverseSearch;
-        request.reverseSearch = false;
         for (Map.Entry<Integer, List<Transfer>> entry: transfersWithSameStart.entrySet()) {
             StreetRouter streetRouter = new StreetRouter(transportNetwork.streetLayer);
+            //reverse search is false
```
abyrd, line 181: Again here upon first reading I don't understand the comments asserting that certain values are true or false.

abyrd (commented):
Revisiting this in October 2020, we're no longer using point to point routing much and I'm trying to understand what exactly this change does. It seems like this is making forward and backward searches a property of an individual street search, rather than something that is set across an entire request. This would facilitate doing several searches in different directions within a single request.

abyrd:

PR #298, opened 2017-07-09
Merge request types

mattwigway:
Don't have separate requests types and code paths for single and multipoint analysis.


mattwigway:
This should be good to go, lmk what you think @abyrd!

src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
```diff
@@ -327,84 +326,65 @@ private void handleOneRequest(GenericClusterRequest clusterRequest) {
                 e.printStackTrace();
             }
             if (random.nextInt(100) >= dryRunFailureRate) {
-                deleteRequest(clusterRequest);
+                deleteRequest(request);
             } else {
-                LOG.info("Intentionally failing to complete task for testing purposes {}", clusterRequest.taskId);
+                LOG.info("Intentionally failing to complete task for testing purposes {}", request.taskId);
```
abyrd: Seeing "request.taskId" reinforces my thought that we should just call them "tasks". `AnalysisTask task; task.taskId` makes sense to me.

src/main/java/com/conveyal/r5/analyst/cluster/GridResultAssembler.java
```diff
@@ -79,19 +79,19 @@
...
     public int nTotal;
 
-    private final String outputBucket;
+    public final String outputBucket;
 
     /** Number of iterations for this grid request */
     private int nIterations;
 
-    public GridResultAssembler (GridRequest request, String outputBucket) {
+    public GridResultAssembler (AnalysisRequest request, String outputBucket) {
```
abyrd: There are no longer subclasses, and instinctively I then think there should be a check that the parameter is the right kind of AnalysisRequest. But that amounts to writing our own type checking, rather than just using the Java type system. We've still got manually coded polymorphism in the sense that there are lots of `if (request.isHighPriority())`, and fields in the requests that are only used in one case or the other. I'm thinking we should still have two subclasses of request.

src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
```diff
@@ -327,84 +326,65 @@ private void handleOneRequest(GenericClusterRequest clusterRequest) {
...
+            TravelTimeComputer computer = new TravelTimeComputer(request, transportNetwork, gridCache);
+            if (request.isHighPriority()) {
+                PipedInputStream pis = new PipedInputStream();
+
+                // gzip the data before sending it to the broker. Compression ratios here are extreme (100x is not uncommon)
+                // so this will help avoid broken pipes, connection reset by peer, buffer overflows, etc. since these types
+                // of errors are more common on large files.
+                GZIPOutputStream pos = new GZIPOutputStream(new PipedOutputStream(pis));
+
+                finishPriorityTask(request, pis, "application/octet-stream", "gzip");
```
abyrd: This still strikes me as manually handled polymorphism. It's not ideal to have classes with a field that says what subtype they are (in this case AnalysisRequest.Type), then checking that subtype field (`if (request.type == X)`) and calling different functions based on the result. We should just have subtypes with different fields, and call a single method on those objects that behaves differently depending on what type they are. The finishPriorityTask method seems like something that should be on a SinglePointAnalysisRequest subclass of AnalysisRequest.

src/main/java/com/conveyal/r5/analyst/cluster/AnalysisRequest.java
```diff
@@ -0,0 +1,75 @@
...
+    public boolean isHighPriority() {
+        return type == Type.TRAVEL_TIME_SURFACE;
+    }
+
+    @JsonIgnore
+    public WorkerCategory getWorkerCategory () {
+        return new WorkerCategory(graphId, workerVersion);
+    }
+
+    public enum Type {
```
mattwigway: It does now, but once we reimplement static site support we will want to add another Type for that.

src/main/java/com/conveyal/r5/analyst/cluster/AnalysisRequest.java
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * Is this a request that should return a binary travel time surface or compute accessibility and return it via SQS
+     * to be saved in a regional analysis grid file?
+     */
+    public Type type;
+
+    /** Travel time surfaces (single point requests) are high priority. Regional analyses are not */
+    @JsonIgnore
+    public boolean isHighPriority() {
```
mattwigway: I don't think we want to do this because we will probably add an additional request type for static sites, and possibly even for non-gridded origins.

src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
mattwigway: Renamed  👍

src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
mattwigway: This is nontrivial to implement due to the architecture of the application at the moment, so I'm not going to have time to do it now.

mattwigway:
Okay @abyrd should be ready for a re-review.

abyrd:
Looks good. Has this been tested and shown to work in a complete backend-broker-worker system for single point and regional analysis?

mattwigway:
Yes. Should we merge?

Issue #306, opened 2017-07-14
FastRaptorRouter searches on schedule-only networks artificially deflated


landonreed:
For `GridComputer` searches on schedule-only networks, regional accessibility results appear to be significantly deflated relative to networks with frequency modifications.  For example, a network with a single 'add trip' and 'reroute' modification produced the following result:

![image](https://user-images.githubusercontent.com/2370911/28228186-63ec7556-68ab-11e7-815f-8eb94a843947.png)

@mattwigway suggests duplicating schedule search results (multiplying by number of Monte Carlo draws per minute used by frequency search) in order to bolster results relative to networks with frequencies.

mattwigway:
Fixed in v2.3.1

PR #308, opened 2017-07-17
Fix deflated access on schedule networks

landonreed:


src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
```diff
@@ -316,9 +316,16 @@ private void advanceScheduledSearchToPreviousMinute (int nextMinuteDepartureTime
 
             return result;
         } else {
-            // no frequencies, return result of scheduled search
-            if (saveAllStates) statesEachIteration.add(scheduleState[request.maxRides].deepCopy());
-            return new int[][] { scheduleState[request.maxRides].bestNonTransferTimes };
+            // No frequencies, return result of scheduled search, but duplicated by the number of
```
mattwigway: I'm not sure it will be clear why accessibility would be artificially deflated if we didn't do this. Maybe just say something about how this avoids potential bugs where assumptions are made about how many results will be returned from a search, and mention this has already happened, and refer to #306 as an example.

src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
```diff
@@ -316,9 +316,16 @@ private void advanceScheduledSearchToPreviousMinute (int nextMinuteDepartureTime
...
+            // MC draws so that the scheduled search accessibility isn't artificially deflated
+            // relative to any networks with frequencies.
+            int[][] result = new int[iterationsPerMinute][];
+            for (int i = 0; i < monteCarloDrawsPerMinute; i++) {
+                if (saveAllStates) statesEachIteration.add(scheduleState[request.maxRides].deepCopy());
+                result[i] = scheduleState[request.maxRides].bestNonTransferTimes;
+            }
+            if (saveAllStates) statesEachIteration.add(scheduleState[request.maxRides * monteCarloDrawsPerMinute].deepCopy());
```
mattwigway: I don't think we want to add it to statesEachIteration again outside the loop, this will result in statesEachIteration being longer than the travel time results.

src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
```diff
@@ -316,15 +316,15 @@ private void advanceScheduledSearchToPreviousMinute (int nextMinuteDepartureTime
...
-            // MC draws so that the scheduled search accessibility isn't artificially deflated
-            // relative to any networks with frequencies.
+            // No frequencies, return result of scheduled search, but multiplied by the number of
+            // MC draws so that the scheduled search accessibility avoids potential bugs where assumptions
+            // are made about how many results will be returned from a search, e.g., in
+            // https://github.com/conveyal/r5/issues/306
             int[][] result = new int[iterationsPerMinute][];
             for (int i = 0; i < monteCarloDrawsPerMinute; i++) {
-                if (saveAllStates) statesEachIteration.add(scheduleState[request.maxRides].deepCopy());
+                if (saveAllStates) statesEachIteration.add(scheduleState[request.maxRides * monteCarloDrawsPerMinute].deepCopy());
```
mattwigway: Why are we multiplying maxRides by monteCarloDrawsPerMinute here? scheduleState contains the result of the RAPTOR search after each ride.

PR #311, opened 2017-07-21
Merge master into remove cruft

mattwigway:
So it will merge cleanly back to master.


mattwigway:
This is now on staging.

I did test this to ensure that pointsets were not being relinked, but once I deployed it they started getting relinked on graph load again (tested in NYC). I'm not sure why that is.

Issue #157, opened 2016-05-10
Allow building of TransportNetwork without OSM network (GTFS only)


landonreed:
Allow building of TransportNetwork without OSM network (GTFS only) so that networks can be built quickly with GTFS feeds alone and allow comparison between feed versions without OSM data behaving as an interfering variable.

landonreed:
@mattwigway or @abyrd, what is the level of effort on this?  My guess is that this could make the use of r5 in datatools a little more feasible for (very rough) isochrone generation.

mattwigway:
I don't think this should be very hard. We need to make a transport network
builder that only builds a transit layer, and then you would call the
existing FastRaptorWorker access times derived from straight line distance.
They would return travel times to all stops, you'd then just need stop to
point distance tables derived from straight line distance so that the
propagater would work.

On May 10, 2016 9:17 AM, "Landon Reed" <notifications@github.com> wrote:

Allow building of TransportNetwork without OSM network (GTFS only) so that
networks can be built quickly with GTFS feeds alone and allow comparison
between feed versions without OSM data behaving as an interfering variable.

—
You are receiving this because you were assigned.
Reply to this email directly or view it on GitHub
<https://github.com/conveyal/r5/issues/157>

PR #339, opened 2017-09-20
Build transport network without OSM

landonreed:
We need to be able to build transport networks without OSM for the following use cases:

1. Quick, low resource building of transport networks when r5 is used as a library.
2. Network-based comparison of GTFS feeds without the influence of potential changes in the OSM networks used by each transport network

### DIFF
--- a/src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
+++ b/src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
@@ -142,7 +142,7 @@ public class LinkedPointSet implements Serializable {
         while (stopToPointDistanceTables.size() < nStops) stopToPointDistanceTables.add(null);
 
         /* First, link the points in this PointSet to specific street vertices. If there is no base linkage, link all streets. */
-        this.linkPointsToStreets(baseLinkage == null);
+        if (streetLayer.hasOSM) this.linkPointsToStreets(baseLinkage == null);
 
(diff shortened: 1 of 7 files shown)


landonreed:
@abyrd, this PR is by no means complete, but I wanted to put it on your radar.  The code currently compiles, but I believe there are currently issues with building the distance tables -- I wasn't quite sure how to approach this in the absence of an OSM network.

abyrd:
I need some more context here - why would we need distance tables if we don't have a street network?

landonreed:
Oh, I'm not sure that we would.  I was just doing my best to build the transport network without OSM and not change anything too dramatically.  But, of course, maybe that isn't the right approach.

I know very little about r5, so these commits may not represent the best path forward.  Additionally, I'm not sure if this PR should be made against `master` or `remove-cruft`.

landonreed:
@abyrd, I'm going to close this PR because a transport network without OSM is no longer needed by Data Tools and this code looks to be very out of date.

abyrd:
Thanks @landonreed. It may in fact be useful to build networks without OSM, e.g. in cases where the OSM data is huge but transit network is small (as would be the case with an Amtrak feed). 

But a change like this is going to require testing and maintenance that we don't currently have time for, so let's file this one away for possible future re-use.

landonreed:
For future reference, this item is being tracked in ticket #157.

PR #343, opened 2017-09-26
fix(pointsets) allow points outside base pointset

ansoncfit:
Addresses a small todo from https://github.com/conveyal/r5/pull/298/commits/9083021e19aeeb10fe8e16e7c38b41d51dacd701, by allowing sub-grids (based on user-defined project bounds) to extend beyond base super-grids (based on TransportNetwork extents).

Not sure if the distance arrays should be filled with `MAX_OFFSTREET_WALK_METERS` or something else.

It might be good to eventually send warnings like at LOC 171 to the front-end, as is currently done for some modification warnings.


abyrd:
@ansoncfit When closing a PR, please leave a comment about why it was closed. In this case, if I'm not mistaken, it's because it was superseded by another PR #344.

ansoncfit:
Superseded by #344

PR #344, opened 2017-09-26
fix(pointsets) allow points outside base pointset

ansoncfit:
Addresses a small todo from 9083021, by allowing sub-grids (based on user-defined project bounds) to extend beyond base super-grids (based on TransportNetwork extents).

Not sure if the distance arrays should be filled with MAX_OFFSTREET_WALK_METERS or something else.

It might be good to eventually send warnings like at LOC 171 to the front-end, as is currently done for some modification warnings.

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/WebMercatorGridPointSet.java
+++ b/src/main/java/com/conveyal/r5/analyst/WebMercatorGridPointSet.java
@@ -60,6 +60,10 @@ public class WebMercatorGridPointSet extends PointSet implements Serializable {
         // TODO don't copy, just have a cache that lazy loads and crops from base pointset
         if (base != null) {
             base.linkageCache.asMap().forEach((key, baseLinkage) -> {
+                // LinkedPointSet now handles the case where the new grid is not completely contained by the base grid.
+                // Since this generally happens when there are points beyond the transit network's extents, marking the points
+                // that are not contained by the base linkage as unlinked (and logging a warning if all points are beyond
+                // the network) is sufficient as you will not be able to reach these locations anyhow.
                 LinkedPointSet croppedLinkage = new LinkedPointSet(baseLinkage, this);
                 this.linkageCache.put(key, croppedLinkage);
             });
--- a/src/main/java/com/conveyal/r5/analyst/WebMercatorGridPointSet.java
+++ b/src/main/java/com/conveyal/r5/analyst/WebMercatorGridPointSet.java
@@ -60,10 +60,6 @@ public class WebMercatorGridPointSet extends PointSet implements Serializable {
         // TODO don't copy, just have a cache that lazy loads and crops from base pointset
         if (base != null) {
             base.linkageCache.asMap().forEach((key, baseLinkage) -> {
-                // TODO handle the case where the new grid is not completely contained by the base grid.
-                // Since this generally will happen when there are points beyond the transit network, just leaving the points
-                // that are not contained by the base linkage unlinked (and logging a warning, or perhaps even returning a
-                // warning to the UI) should be completely sufficient as you will not be able to reach these locations anyhow.
                 LinkedPointSet croppedLinkage = new LinkedPointSet(baseLinkage, this);
                 this.linkageCache.put(key, croppedLinkage);
             });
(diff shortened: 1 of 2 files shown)


src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
```diff
@@ -185,10 +187,17 @@ public LinkedPointSet(LinkedPointSet sourceLinkage, WebMercatorGridPointSet subG
...
-                edges[pixel] = sourceLinkage.edges[sourcePixel];
-                distances0_mm[pixel] = sourceLinkage.distances0_mm[sourcePixel];
-                distances1_mm[pixel] = sourceLinkage.distances1_mm[sourcePixel];
+                if (sourceColumn < 0 || sourceColumn >= superGrid.width || sourceRow < 0 || sourceRow >= superGrid.height) { //point is outside super-grid
+                    edges[pixel] = -1; //set the edge value to -1 to indicate no linkage.
+                    distances0_mm[pixel] = MAX_OFFSTREET_WALK_METERS;
+                    distances1_mm[pixel] = MAX_OFFSTREET_WALK_METERS;
+                } else { //point is inside super-grid
+                    int sourcePixel = sourceRow * superGrid.width + sourceColumn;
+                    LOG.info(Integer.toString(sourcePixel));
```
abyrd: This line is causing a lot of log noise, and is probably greatly slowing down the copy operation.

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
```diff
@@ -185,10 +187,17 @@ public LinkedPointSet(LinkedPointSet sourceLinkage, WebMercatorGridPointSet subG
             for (int x = 0; x < subGrid.width; x++, pixel++) {
                 int sourceColumn = subGrid.west + x - superGrid.west;
                 int sourceRow = subGrid.north + y - superGrid.north;
-                int sourcePixel = sourceRow * superGrid.width + sourceColumn;
-                edges[pixel] = sourceLinkage.edges[sourcePixel];
-                distances0_mm[pixel] = sourceLinkage.distances0_mm[sourcePixel];
-                distances1_mm[pixel] = sourceLinkage.distances1_mm[sourcePixel];
+                if (sourceColumn < 0 || sourceColumn >= superGrid.width || sourceRow < 0 || sourceRow >= superGrid.height) { //point is outside super-grid
+                    edges[pixel] = -1; //set the edge value to -1 to indicate no linkage.
+                    distances0_mm[pixel] = MAX_OFFSTREET_WALK_METERS;
```
abyrd: It shouldn't matter what the distances are if the edges[n] == -1. The code reading it should completely skip the point and not even look at the distances.

src/main/java/com/conveyal/r5/analyst/WebMercatorGridPointSet.java
```diff
@@ -60,10 +60,6 @@ public WebMercatorGridPointSet(int zoom, int west, int north, int width, int hei
         // TODO don't copy, just have a cache that lazy loads and crops from base pointset
         if (base != null) {
             base.linkageCache.asMap().forEach((key, baseLinkage) -> {
-                // TODO handle the case where the new grid is not completely contained by the base grid.
```
abyrd: Although this is no longer a TODO the comment should still be retained, explaining why it's OK to do what we're doing (the cells are unreachable by transit or streets so might as well be unlinked).

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
abyrd: Also there should probably be a comment explaining that.

ansoncfit: I'll remove these lines

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
ansoncfit: Stray log statement, will remove.

src/main/java/com/conveyal/r5/analyst/WebMercatorGridPointSet.java
ansoncfit: 👍

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
ansoncfit: and add a comment explaining why we don't need to set distances here.

PR #345, opened 2017-09-28
Coverage and point to point tests

evansiroky:
* Added coverage reporting
* Adding tests for PointToPointRouterServer


evansiroky:
Just updated this PR to be mergeable into the latest dev branch.

evansiroky:
@abyrd Just updated this PR with the latest changes in master.  One of my tests is having a hard time passing.   The test that checks if a network can be built is failing on this assertion: https://github.com/conveyal/r5/blob/a27d0cfe3d85d4aab898d150879fd383b7984da1/src/test/java/com/conveyal/r5/point_to_point/PointToPointRouterServerTest.java#L60-L63

The hash of the generated `network.dat` file isn't always the same.  Is there a reason why it's sometimes different?  Should I modify my test to simply check if the filesize is greater than an arbitrary amount of bytes?

abyrd:
@evansiroky it's not unexpected that the contents of the serialized graph are different each time. No special effort was made to make it deterministic and repeatable. There could be a timestamp or a UUID buried somewhere in there, or the items in a hashtable or set could somehow end up in a different order etc. I'd say this assertion is not checking an expected characteristic of the serialized network. Just to let you know, the serialization system has been completely replaced in the dev branch, so the characteristics of the existing serialization system on master will change radically when dev is merged.

As part of that new serialization system, dev contains tests that build networks, serialize them, and perform a pretty sophisticated recursive semantic equality check. I think it would be safe to just remove this particular test and rely on https://github.com/conveyal/r5/blob/dev/src/test/java/com/conveyal/r5/kryo/KryoNetworkSerializerTest.java

src/test/java/com/conveyal/r5/point_to_point/PointToPointRouterServerTest.java
```diff
@@ -0,0 +1,429 @@
...
+        String fromLat = "39.465659";
+        String fromLon = "-87.410839";
+        String fromTime = "2017-09-18T12:00:00.000Z";
+        String toLat = "39.4767";
+        String toLon = "-87.4052";
+        String toTime = "2017-09-18T14:00:00.000Z";
+        String transitModes = "BUS,RAIL,SUBWAY,TRAM";
+        String walkSpeed = "3";
+
+        return "{" +
```
abyrd: Should this be stored as a resource file rather than a string literal? I find things like this less readable/maintainable when inlined into the code.

src/test/java/com/conveyal/r5/point_to_point/PointToPointRouterServerTest.java
```diff
@@ -0,0 +1,429 @@
...
+public class PointToPointRouterServerTest {
+    private static final String resourcesDir = "./src/test/resources/com/conveyal/r5/point_to_point";
+
+    /**
+     * Start up a point to point route server that will be queried in various tests
+     */
+    @BeforeClass
+    public static void setUp() throws IOException {
+        // create a temporary director and copy needed files there
+        String[] files = {"network.dat", "osm.mapdb", "osm.mapdb.p"};
```
abyrd: These three binary files are included in the test resources. I tend to see them as intermediate results whose format is subject to change, with the original source data being the GTFS and OSM data. I'd like to avoid checking intermediate results in (especially the MapDB) if possible.

src/test/java/com/conveyal/r5/point_to_point/PointToPointRouterServerTest.java
evansiroky: Fixed in https://github.com/conveyal/r5/pull/345/commits/908d331ec291f2802cd484407d542022f5cd304d

src/test/java/com/conveyal/r5/point_to_point/PointToPointRouterServerTest.java
evansiroky: Fixed in https://github.com/conveyal/r5/pull/345/commits/908d331ec291f2802cd484407d542022f5cd304d

evansiroky:
@abyrd This PR should be good to go pending your approval.

PR #348, opened 2017-10-05
Misc. fixes for walk and bike

ansoncfit:
Work in progress on a few walk/bike/access mode bugs


abyrd:
Discussing with @ansoncfit the fix for the situation where `offset >= times.length`. This is happening in a travel time reducer function that extracts n percentiles from a list of travel times, on walk or bike modes where there is only one travel time at a given destination. The culprit line is `int offset = (int) Math.round(task.percentiles[i] / 100d * times.length);`. When there is only one element in the array, the index rounds up to 1 and runs off the end of the array. Of course this can happen with other combinations of array lengths and percentiles. The root problem is that we're multiplying by `times.length`; we're actually trying to scale the interval between the beginning and end elements of the array (the minimum and maximum values). In an array with N values the interval is N-1 elements. We should be scaling N-1, which makes the result always defined even with rounding up.

Issue #315, opened 2017-07-22
Faster algorithm for rasterizing polygons


mattwigway:
We currently intersect every pixel geometry (rectangular polygon) within the bounding box with each polygon when rasterizing polygons into a grid. This is too slow.

ansoncfit:
@abyrd and I discussed this again today. 

When loading a shapefile as large as the UK, the staging server does not have enough memory to cope.

There's [more work](https://github.com/conveyal/r5/blob/master/src/main/java/com/conveyal/r5/analyst/Grid.java#L138) to do to "avoid duplicating expensive geometric math."

ansoncfit:
Amending the above.  In July, the staging server showed high CPU utilization when three users attempted to set up Analysis regions and upload shapefiles, including one for the UK.  I'm not sure that there was "not enough memory to cope."

ansoncfit:
Addressed in https://github.com/conveyal/r5/pull/350

PR #350, opened 2017-10-17
Speedier rasterization

ansoncfit:
I've tried to follow some suggestions in the code related to #315, but it doesn't seem that this initial work actually speeds things up -- the geotools implementation of intersection may already do what I've tried to here.

Using Lists will require changes to AggregationAreaController and SeamlessCensusGridExtractor in analysis-backend (on the speedier-rasterization branch).

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/Grid.java
+++ b/src/main/java/com/conveyal/r5/analyst/Grid.java
@@ -49,6 +49,7 @@ import java.nio.charset.Charset;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.ArrayList;
+import java.util.List;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
@@ -169,7 +170,7 @@ public class Grid {
      * Version of getPixelWeights which returns the weights as relative to the total area of the input geometry (i.e.
      * the weight at a pixel is the proportion of the input geometry that falls within that pixel.
      */
-    public ArrayList<PixelWeight> getPixelWeights (Geometry geometry) {
+    public List<PixelWeight> getPixelWeights (Geometry geometry) {
         return getPixelWeights(geometry, false, null);
     }
 
@@ -185,13 +186,13 @@ public class Grid {
      * This used to return a map from int arrays containing the coordinates to the weight.
      *
      */
-    private ArrayList<PixelWeight> getPixelWeights (Geometry geometry, boolean relativeToPixels, PixelWeightCallback pixelWeightCallback) {
+    private List<PixelWeight> getPixelWeights (Geometry geometry, boolean relativeToPixels, PixelWeightCallback pixelWeightCallback) {
         // No need to convert to a local coordinate system
         // Both the supplied polygon and the web mercator pixel geometries are left in WGS84 geographic coordinates.
         // Both are distorted equally along the X axis at a given latitude so the proportion of the geometry within
         // each pixel is accurate, even though the surface area in WGS84 coordinates is not a usable value.
 
-        ArrayList<PixelWeight> weights = new ArrayList<>();
+        List<PixelWeight> weights = new ArrayList<>();
 
         double area = geometry.getArea();
         if (area < 1e-12) {
@@ -254,7 +255,7 @@ public class Grid {
     }
 
     /** Using a grid of weights produced by getPixelWeights, burn the value of a polygon into the grid */
-    public void incrementFromPixelWeights (ArrayList weights, double value) {
+    public void incrementFromPixelWeights (List weights, double value) {
         Iterator<PixelWeight> pixelWeightIterator = weights.iterator();
         while(pixelWeightIterator.hasNext()) {
             PixelWeight pix = pixelWeightIterator.next();
--- a/src/main/java/com/conveyal/r5/analyst/Grid.java
+++ b/src/main/java/com/conveyal/r5/analyst/Grid.java
@@ -131,7 +131,7 @@ public class Grid {
         int y;
         double weight;
 
-        public PixelWeight (int x, int y, double weight){
+        private PixelWeight (int x, int y, double weight){
             this.x = x;
             this.y = y;
             this.weight = weight;
@@ -148,13 +148,32 @@ public class Grid {
         }
     }
 
+    /**
+     * A functional interface (callback interface) that processes one pixel weight at a time.
+     * This avoids storing massive lists of intermediate results that have been known to eat memory on the server.
+     * This happens when processing geometries that contain a huge amount of pixels, which a user might include in
+     * a shapefile.
+     */
+    public interface PixelWeightCallback {
+        void handlePixelWeight (int x, int y, double weight);
+    }
+
+    /**
+     *
+     * @param geometry
+     * @param relativeToPixels
+     * @param pixelWeightCallback
+     */
+    public void streamPixelWeights (Geometry geometry, boolean relativeToPixels, PixelWeightCallback pixelWeightCallback) {
+        getPixelWeights(geometry, relativeToPixels, pixelWeightCallback);
+    }
 
     /**
      * Version of getPixelWeights which returns the weights as relative to the total area of the input geometry (i.e.
      * the weight at a pixel is the proportion of the input geometry that falls within that pixel.
      */
     public ArrayList<PixelWeight> getPixelWeights (Geometry geometry) {
-        return getPixelWeights(geometry, false);
+        return getPixelWeights(geometry, false, null);
     }
 
     /**
@@ -163,13 +182,13 @@ public class Grid {
      * grid.
      *
      * If relativeToPixels is true, the weights are the proportion of the pixel that is covered. Otherwise they are the
-     * portion of this polygon which is within the given grid cell. If using incrementPixelWeights, this should be set to
+     * portion of this polygon which is within the given pixel. If using incrementPixelWeights, this should be set to
      * false.
      *
      * This used to return a map from int arrays containing the coordinates to the weight.
      *
      */
-    public ArrayList<PixelWeight> getPixelWeights (Geometry geometry, boolean relativeToPixels) {
+    private ArrayList<PixelWeight> getPixelWeights (Geometry geometry, boolean relativeToPixels, PixelWeightCallback pixelWeightCallback) {
         // No need to convert to a local coordinate system
         // Both the supplied polygon and the web mercator pixel geometries are left in WGS84 geographic coordinates.
         // Both are distorted equally along the X axis at a given latitude so the proportion of the geometry within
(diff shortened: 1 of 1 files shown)


src/main/java/com/conveyal/r5/analyst/Grid.java
```diff
@@ -130,52 +128,90 @@ public Grid (int zoom, int width, int height, int north, int west) {
      * Version of getPixelWeights which returns the weights as relative to the total area of the input geometry (i.e.
      * the weight at a pixel is the proportion of the input geometry that falls within that pixel.
      */
-    public TObjectDoubleMap<int[]> getPixelWeights (Geometry geometry) {
+    public ArrayList<PixelWeight> getPixelWeights (Geometry geometry) {
         return getPixelWeights(geometry, false);
     }
 
+    public class PixelWeight {
```
ansoncfit: public static class

src/main/java/com/conveyal/r5/analyst/Grid.java
```diff
@@ -130,52 +128,90 @@ public Grid (int zoom, int width, int height, int north, int west) {
...
         }
 
         Envelope env = geometry.getEnvelopeInternal();
-        for (int worldx = lonToPixel(env.getMinX(), zoom); worldx <= lonToPixel(env.getMaxX(), zoom); worldx++) {
+
+        for (int worldy = latToPixel(env.getMaxY(), zoom); worldy <= latToPixel(env.getMinY(), zoom); worldy++) {
             // NB web mercator Y is reversed relative to latitude
-            for (int worldy = latToPixel(env.getMaxY(), zoom); worldy <= latToPixel(env.getMinY(), zoom); worldy++) {
+
+            double pixelAreaAtLat = 0;
```
ansoncfit: set to something "clearly impossible" e.g. -1

abyrd:
Discussing this now with @ansoncfit. The proposed new system is likely just causing complex intersection math to be done 2-3 times in a row. It also doesn't solve one core problem that we store all the intermediate results (pixel weights) in memory before operating on them one by one. We can actually operate on each pixel separately in a callback, without storing any results at all:


```
    /**
     * A functional interface (callback interface) that processes one pixel weight at a time. 
     * This avoids storing massive lists of intermediate results that have been known to eat memory on the server.
     * This happens when processing geometries that contain a huge amount of pixels, which a user might include in 
     * a shapefile.
     */
    interface PixelWeightCallback {
        public void handlePixelWeight (int x, int y, double weight);
    }

public void getPixelWeights (Geometry geometry, PixelWeightCallback pixelWeightCallback) {
        getPixelWeights(geometry, pixelWeightCallback, false);
    }

```

The getPixelWeights function is then used by passing in a callback function that operates on each pixel in turn:

```
getPixelWeights(geometry, (x, y, weight) -> {
            grid[x][y] += weight * value;
        });

```

ansoncfit:
For the case of many attributes (e.g. a shapefile with many columns), maybe we do want to keep the weights in memory.  See https://github.com/conveyal/analysis-backend/blob/master/src/main/java/com/conveyal/taui/grids/SeamlessCensusGridExtractor.java#L97 ("do geographic math once and cache for whole feature").  I'm also amending #315 -- I'm not certain that the issue I saw previously was hitting a memory limit.

ansoncfit:
This PR now:
1. Includes a streamPixelWeights method, to avoid writing an ArrayList of PixelWeights, then iterating through them.  We still want the iterable in-memory option for cases when we're repeating over many properties/shapefile columns.
2. Uses JTS PreparedGeometry to calculate intersection areas only when necessary, making rasterization ~20% faster.

src/main/java/com/conveyal/r5/analyst/Grid.java
```diff
@@ -126,96 +126,154 @@ public Grid (int zoom, int width, int height, int north, int west) {
...
+        double weight;
+
+        private PixelWeight (int x, int y, double weight){
+            this.x = x;
+            this.y = y;
+            this.weight = weight;
+        }
+
+        public int getX() {
+            return x;
```
abyrd: I would favor final fields, set in the constructor, without accessor methods ("getters"). This is the approach we've been using in most other Java code that doesn't need serious encapsulation of fields.

src/main/java/com/conveyal/r5/analyst/Grid.java
```diff
@@ -126,96 +126,154 @@ public Grid (int zoom, int width, int height, int north, int west) {
...
+        getPixelWeights(geometry, relativeToPixels, pixelWeightCallback);
+    }
+
     /**
      * Version of getPixelWeights which returns the weights as relative to the total area of the input geometry (i.e.
      * the weight at a pixel is the proportion of the input geometry that falls within that pixel.
      */
-    public TObjectDoubleMap<int[]> getPixelWeights (Geometry geometry) {
-        return getPixelWeights(geometry, false);
+    public ArrayList<PixelWeight> getPixelWeights (Geometry geometry) {
```
abyrd: Field and return types should usually be an interface type (List) rather than the implementation type (ArrayList).

src/main/java/com/conveyal/r5/analyst/Grid.java
```diff
@@ -126,96 +126,154 @@ public Grid (int zoom, int width, int height, int north, int west) {
...
+        return getPixelWeights(geometry, false, null);
     }
 
     /**
-     * Get the proportions of an input polygon feature that overlap each grid cell, in the format [x, y] => weight.
-     * This weight object can then be fed into the incrementFromPixelWeights function to actually burn a polygon into the
+     * Get the proportions of an input polygon feature that overlap each grid cell, for use in List(x, y, weight)
```
abyrd: It's not clear to me what `List(x, y, weight)` means.

src/main/java/com/conveyal/r5/analyst/Grid.java
```diff
@@ -599,7 +657,7 @@ public static Polygon getPixelGeometry (int x, int y, int zoom) {
 
             if (statusListener != null) statusListener.accept(currentCount, total);
 
-            if (currentCount % 10000 == 0) {
+            if (currentCount % 100 == 0) {
```
abyrd: This is going to produce large amounts of log output on large Shapefiles. Do we need anything more that a few log entries indicating that the file is being processed?

src/main/java/com/conveyal/r5/analyst/Grid.java
```diff
@@ -126,96 +126,154 @@ public Grid (int zoom, int width, int height, int north, int west) {
...
+    private void rasterize (Geometry geometry, double value) {
         incrementFromPixelWeights(getPixelWeights(geometry), value);
     }
 
     /** Using a grid of weights produced by getPixelWeights, burn the value of a polygon into the grid */
-    public void incrementFromPixelWeights (TObjectDoubleMap<int[]> weights, double value) {
-        for (TObjectDoubleIterator<int[]> it = weights.iterator(); it.hasNext();) {
-            it.advance();
-            grid[it.key()[0]][it.key()[1]] += it.value() * value;
+    public void incrementFromPixelWeights (ArrayList weights, double value) {
```
abyrd: Parameter type should be interface (List or Iterable or Collection) not a specific implementation of that interface (ArrayList).

src/main/java/com/conveyal/r5/analyst/Grid.java
```diff
@@ -126,96 +126,154 @@ public Grid (int zoom, int width, int height, int north, int west) {
...
+        Iterator<PixelWeight> pixelWeightIterator = weights.iterator();
+        while(pixelWeightIterator.hasNext()) {
```
abyrd: You can use the foreach syntax here: `for (PixelWeight pix : weights) { ... }`. You only need to create an iterator explicitly when you want to call methods on that iterator, e.g. removing elements.

src/main/java/com/conveyal/r5/analyst/Grid.java
```diff
@@ -126,96 +126,154 @@ public Grid (int zoom, int width, int height, int north, int west) {
...
 
         double area = geometry.getArea();
         if (area < 1e-12) {
             throw new IllegalArgumentException("Geometry is too small");
         }
 
+        // PreparedGeometry is often faster for small numbers of vertices;
+        // see https://github.com/chrisbennight/intersection-test
+
+        PreparedGeometryFactory pgFact = new PreparedGeometryFactory();
```
abyrd: I wonder if we're really intended to construct a new factory for each single prepared geometry we create. It might be sufficient to have one factory instance in a static field on this class, assuming that the factory is threadsafe.

src/main/java/com/conveyal/r5/analyst/Grid.java
```diff
@@ -126,96 +126,154 @@ public Grid (int zoom, int width, int height, int north, int west) {
...
      *
      * If relativeToPixels is true, the weights are the proportion of the pixel that is covered. Otherwise they are the
-     * portion of this polygon which is within the given grid cell. If using incrementPixelWeights, this should be set to
+     * portion of this polygon which is within the given pixel. If using incrementPixelWeights, this should be set to
      * false.
      *
-     * This returns a map from int arrays containing the coordinates to the weight. I believe int arrays in hashmaps are
-     * hashed based on memory location, meaning lookups will fail. However, we only use this in a streaming fashion, looping
-     * over all values, which is harmless.
+     * This used to return a map from int arrays containing the coordinates to the weight.
```
abyrd: It's pretty unconventional to have a function that returns an empty collection when it has been given a callback, but otherwise puts things in the collection. It would seem a little safer to me if the function returned null when it used the callback. It took me a moment to realize this is an internal private method that's called by two different public methods with different signatures, which is a good way to do it, but deserves a very clear explanation in the Javadoc.

src/main/java/com/conveyal/r5/analyst/Grid.java
ansoncfit: Oops, forgot to reset this from testing locally.

ansoncfit:
Thanks for the feedback, @abyrd.  Made the suggested changes here and in analysis-backend.

src/main/java/com/conveyal/r5/analyst/Grid.java
```diff
@@ -126,96 +127,154 @@ public Grid (int zoom, int width, int height, int north, int west) {
...
+        }
+    }
+
+    /**
+     * A functional interface (callback interface) that processes one pixel weight at a time.
+     * This avoids storing massive lists of intermediate results that have been known to eat memory on the server.
+     * This happens when processing geometries that contain a huge amount of pixels, which a user might include in
+     * a shapefile.
+     */
+    public interface PixelWeightCallback {
```
abyrd: If most of the speedup comes from use of PreparedGeometry, and these in-memory pixel weights are not one of our major memory consumers in this application, we should probably just keep the programming interface for this pixel weighting code as simple as possible, i.e. just do them all in memory since many of them need to be in memory.

src/main/java/com/conveyal/r5/analyst/Grid.java
```diff
@@ -516,7 +575,7 @@ public static Polygon getPixelGeometry (int x, int y, int zoom) {
 
         int i = 0;
         while (reader.readRecord()) {
-            if (++i % 10000 == 0) {
+            if (++i % 1000 == 0) {
```
abyrd: This will generate a lot of output, probably not useful outside development environment.

ansoncfit:
This PR now:

Includes a streamPixelWeights method, to avoid writing an ArrayList of PixelWeights, then iterating through them. We still want the iterable in-memory option for cases when we're repeating over many properties/shapefile columns.

It also uses Lists instead of Maps, requiring a change in analysis-backend: https://github.com/conveyal/analysis-backend/pull/70

PR #353, opened 2017-10-17
perf(logging): stop printing all zero-length hop warnings

ansoncfit:
Loading feeds like the MBTA or LA Metro takes an extra ~5 minutes because this warning is printed for all zero-duration hops, which are prevalent when feed maintainers round stop_times arrival/departure times to the nearest minute.

We could eventually build in more logic to check whether the implied speeds are reasonable.  But for now, this speeds things up, and still gives a warning that would raise a red flag for feeds that don't have rounded stop_times.


abyrd:
This looks like a good solution. Because of rounding to the nearest minute we're going to be seeing these teleportations forever. Because we know the stop locations we could calculate the minimum distance between the two stops and issue a warning when it's too far to realistically travel in 60 seconds, but we will be doing that kind of more thorough checking in gtfs-lib.

PR #357, opened 2017-10-30
Add 1 to grid height and width to prevent data loss

trevorgerhardt:
Grids could have a height of zero if all of the data was at the same latitude which would cause an
empty grid. Also, any data at the bounds of the grid would be left out since it would be equal to
the height or the width where our bounds check requires it to be less than those values ("within" the grid).

**NOTE**: A new release needs to be cut for this change since it is needed to complete https://github.com/conveyal/analysis-backend/pull/67


abyrd:
@trevorgerhardt have you actually encountered this causing failures? Of course I agree all grids should be at least one pixel wide and high, just wondering how this influences conveyal/analysis-backend#67. And do we really want to add one to the width and height, or just force them to be >= 1?

ansoncfit:
While we're looking at grid dimensions, is https://github.com/conveyal/analysis-backend/issues/53 a related off-by-one issue?

trevorgerhardt:
There are two problems that this solves @abyrd:

1. Important but uncommon: Allowing for points that are all in the same latitude or longitude. (height or width would = 0)
2. Very important and very common: Stop ignoring and actually create the points where the position of the point computes to the pixel edge of the boundary.

I assume there is more than one way to solve these problems. This is one of them. 

@ansoncfit not sure but I really doubt it.

ansoncfit:
I don't have a strong preference between always adding one vs. only adding one if the width/height would otherwise be zero.

How will the former affect the ability to use the bounds of one regional analysis to create another?  Will the proposed approach increment the bounds, preventing the completed analyses from being compared because their bounds are different?

ansoncfit:
Also related?

https://github.com/conveyal/analysis-ui/issues/349

trevorgerhardt:
@ansoncfit I don't believe this will have any effect on Project or Analysis bounds since it is confined to Opportunity Dataset grid creation and they are separate.

trevorgerhardt:
But @ansoncfit we should still be aware that this type of code problem could be an issue in those cases.

ansoncfit:
The following constructor (with ints) is used in SelectingGridReducer.  Maybe we should add a TODO to make that constructor behave similarly?

abyrd:
Oh right, if the grid extent is computed from the points themselves, if the cell number for the right edge of the grid is rounded down, some points will almost always fall outside the grid.

The key is that latToPixel and lonToPixel naturally round down - which is the correct behavior for binning points into cells but means the grid is always 1 row too narrow/short (except in the super rare case where the points with max coordinates fall exactly on a grid boundary). 

@ansoncfit we should be able to compare / intersect / etc. grids of different dimensions, assuming default values in all non-existing cells. We already do this to some extent, and should make it work anywhere it doesn't. So I wouldn't worry about not being able to compare the resulting grids at all. Besides this "leak" or grid expansion via copying should be pretty rare.

abyrd:
I would prefer a comment in the code though explaining the above.

trevorgerhardt:
I'll add a comment!

trevorgerhardt:
Done.

Issue #349, opened 2017-10-09
Bike has zero accessibility in most places


abyrd:
In places where walk+transit produces reasonable results, bike+transit does not. You get reasonable accessibility figures in a few scattered locations, but zero accessibility in most locations.

![image](https://user-images.githubusercontent.com/112871/31353083-2223bc5e-acff-11e7-85de-175a0b9f67a2.png)

![image](https://user-images.githubusercontent.com/112871/31353099-30a8b806-acff-11e7-9d8c-0b1aa7a26cfb.png)

ansoncfit:
Biking works as expected as a direct mode (i.e. with transit turned off):
![image](https://user-images.githubusercontent.com/2173529/32304007-8c9d3722-bf42-11e7-9614-42ee41c64b99.png)
(from the linking-fixes branch)

But not as an access mode:
![image](https://user-images.githubusercontent.com/2173529/32304074-ea96019c-bf42-11e7-88d8-e2dd1bf86c4c.png)

ansoncfit:
This expects a biking access limit in seconds:
https://github.com/conveyal/r5/blob/master/src/main/java/com/conveyal/r5/analyst/TravelTimeComputer.java#L157

But our default request has `maxBikeTime: 20` (presumably minutes, or we've got some lazy bike riders)

Multiplying by 60 at the line above should fix the issue.

Issue #347, opened 2017-10-05
Walk/bike with no transit is too fast


ansoncfit:
Isochrone on a day with no transit service (i.e. walking only, up to walk limit):
![image](https://user-images.githubusercontent.com/2173529/31248118-c9b7ac18-a9e0-11e7-8522-b0206ab29233.png)

Same settings, but turning off the transit mode selector:
![image](https://user-images.githubusercontent.com/2173529/31248152-e5790a46-a9e0-11e7-82a5-bddf555b261c.png)

Related to https://github.com/conveyal/r5/pull/295?

ansoncfit:
I think the culprit is here: https://github.com/conveyal/r5/blob/master/src/main/java/com/conveyal/r5/analyst/TravelTimeComputer.java#L78

If the front-end is expecting results in seconds, this is off by a factor of 60.

Issue #337, opened 2017-09-18
do not link water to roads with zero cost


abyrd:
It seems like we're still linking faraway cells to roads with minimal or no cost. This should be changed with a shorter linking distance and lower speed to traverse any >1 cell links.

abyrd:
We appear to still have this problem wiith driving searches. There are still big horizontal jumps in isochrones at the coastline (see image) which imply zero travel time. ![image](https://user-images.githubusercontent.com/112871/39625287-0c59af78-4f9d-11e8-9409-c3e4b8718e19.png)

abyrd:
This might just be an illusion caused by the car being relatively fast. Still should be checked in the code to make sure speed is being properly applied.

ansoncfit:
We are applying a cost: https://github.com/conveyal/r5/blob/dev/src/main/java/com/conveyal/r5/streets/LinkedPointSet.java#L320

For cars, the assumed speed is 11 m/s (https://github.com/conveyal/r5/blob/dev/src/main/java/com/conveyal/r5/profile/ProfileRequest.java#L66), which takes one pretty far into the water in 60 seconds. 🌊 

Other than linking pointset points to streets, it looks like the only other place `carSpeed` might be used is the StreetRouter's A* heuristic.  So we could consider reducing it to something closer to real driveway/parking lot speeds.

ansoncfit:
:tada: This issue has been resolved in version 4.2.0 :tada:

Your **[semantic-release](https://github.com/semantic-release/semantic-release)** bot :package::rocket:

PR #358, opened 2017-10-31
Linking fixes

ansoncfit:
Fixes a couple problems:
1. non-transit analyses too fast by a factor of 60 (#347)
2. biking access too slow by a factor of 60 (#349)
3. no cost to access network, leading to long swims (#337)

Adds steps to set egress mode (but this is not yet fully implemented)

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/TravelTimeComputer.java
+++ b/src/main/java/com/conveyal/r5/analyst/TravelTimeComputer.java
@@ -75,10 +75,10 @@ public class TravelTimeComputer {
                 int x = target % request.width;
                 int y = target / request.width;
 
-                final int travelTimeMinutes =
-                        travelTimesToTargets[target] == FastRaptorWorker.UNREACHED ? FastRaptorWorker.UNREACHED : travelTimesToTargets[target] / 60;
+                final int travelTimeSeconds =
+                        travelTimesToTargets[target] == FastRaptorWorker.UNREACHED ? FastRaptorWorker.UNREACHED : travelTimesToTargets[target];
 
-                output.accept(y * request.width + x, new int[] { travelTimeMinutes });
+                output.accept(y * request.width + x, new int[] { travelTimeSeconds });
             }
 
             output.finish();
--- a/src/main/java/com/conveyal/r5/analyst/TravelTimeComputer.java
+++ b/src/main/java/com/conveyal/r5/analyst/TravelTimeComputer.java
@@ -44,6 +44,7 @@ public class TravelTimeComputer {
     public void write (OutputStream os) throws IOException {
         StreetMode accessMode = LegMode.getDominantStreetMode(request.accessModes);
         StreetMode directMode = LegMode.getDominantStreetMode(request.directModes);
+        StreetMode egressMode = LegMode.getDominantStreetMode(request.egressModes);
 
         double fromLat = request.fromLat;
         double fromLon = request.fromLon;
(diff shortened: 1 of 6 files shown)


ansoncfit:
@abyrd we talked about this a bit last month.  The isochrones don't extend far into the water now, but it still may be worth it to put a water tile layer above the isochrone layer in the UI.

image

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
```diff
@@ -307,13 +326,31 @@ public PointSetTimes eval (TravelTimeFunction travelTimeForVertex, int offstreet
...
             int time1 = travelTimeForVertex.getTravelTime(edge.getToVertex());
 
-            // TODO apply walk speed
+            // An "off-roading" penalty is applied to limit extending isochrones into water bodies, etc.
+            // We may want to keep the MAX_OFFSTREET_WALK_METERS relatively high to avoid holes in the isochrones,
+            // but make it costly to walk long distances where there aren't streets.  The approach below
+            // accomplishes that, applying a penalty to off-street distances greater than the typical grid cell size.
+            // We could use a distance threshold more closely tied to pointset resolution/coverage
+            int scaledLinkDist = distancesToEdge_mm[i] > CELL_SIZE*1000 ?
+                    distancesToEdge_mm[i] + OFFROAD_PENALTY * (distancesToEdge_mm[i] - CELL_SIZE * 1000) : distancesToEdge_mm[i];
```
abyrd: I'm hesitant to use empirically tuned unitless multipliers to keep the isochrone edges from extending out into the water. They would probably work in one place and not another, and seem to be prone to overflow. Also, because they are pre-multiplied into the distance we lose information that could be useful for dynamically applying speed or distance parameters during routing. It's also hard to justify why speed would change at some predetermined distance - if we consider a cell within reachable distance at all, how do we know where the hypothetical walking path to that cell ends? Instead I think we should just limit the linking distance, then apply the same walking/biking speed across the full length.

src/main/java/com/conveyal/r5/streets/Split.java
```diff
@@ -48,6 +53,8 @@ public void setFrom (Split other) {
         distSquared = other.distSquared;
```
abyrd: We should be able to find a fairly accurate distance with `VertexStore.fixedDegreesToFloating(FastMath.sqrt(distSquared)) * metersPerDegreeLat * 1000;` and such a line deserves a comment about what's going on.

src/main/java/com/conveyal/r5/analyst/TravelTimeComputer.java
```diff
@@ -154,7 +155,7 @@ public void write (OutputStream os) throws IOException {
                 // Other modes are already asymmetric with the egress/stop trees, so just do a time-based on street
                 // search and don't worry about distance limiting.
                 sr.streetMode = accessMode;
-                sr.timeLimitSeconds = request.getMaxAccessTimeForMode(accessMode);
+                sr.timeLimitSeconds = request.getMaxAccessTimeForMode(accessMode)*60;
```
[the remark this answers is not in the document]
ansoncfit: Addressed in https://github.com/conveyal/r5/pull/358/commits/f1dca6bf4811211c224d12315d1e5d7a2ff7f02f

src/main/java/com/conveyal/r5/streets/Split.java
ansoncfit: Addressed in https://github.com/conveyal/r5/pull/358/commits/440e107de872429dd5c0da2461cbcd7bd0a7d945

src/main/java/com/conveyal/r5/analyst/TravelTimeComputer.java
```diff
@@ -75,10 +76,10 @@ public void write (OutputStream os) throws IOException {
                 int x = target % request.width;
                 int y = target / request.width;
 
-                final int travelTimeMinutes =
-                        travelTimesToTargets[target] == FastRaptorWorker.UNREACHED ? FastRaptorWorker.UNREACHED : travelTimesToTargets[target] / 60;
+                final int travelTimeSeconds =
+                        travelTimesToTargets[target] == FastRaptorWorker.UNREACHED ? FastRaptorWorker.UNREACHED : travelTimesToTargets[target];
 
-                output.accept(y * request.width + x, new int[] { travelTimeMinutes });
+                output.accept(y * request.width + x, new int[] { travelTimeSeconds });
```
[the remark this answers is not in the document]
ansoncfit: Addressed in https://github.com/conveyal/r5/pull/358/commits/f1dca6bf4811211c224d12315d1e5d7a2ff7f02f

PR #361, opened 2017-11-07
Allow exact-times to be false

ansoncfit:
Small bugfix (for #290).

Also bump version number (unrelated to this bugfix; should have been done on the previously merged speedier-rasterization branch)

### DIFF
--- a/src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
+++ b/src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
@@ -445,6 +445,10 @@ public class FastRaptorWorker {
     /** Do a frequency search. If computeDeterministicUpperBound is true, worst-case frequency boarding time will be used
      * so that the output of this function can be used in a range-RAPTOR search. Otherwise Monte Carlo schedules will be
      * used to improve upon the output of the range-RAPTOR bounds search.
+     *
+     * @param computeDeterministicUpperBound specifies whether to compute a deterministic upper bound, which helps speed up
+     *                                       subsequent frequency searches. If false, a bona fide frequency search is conducted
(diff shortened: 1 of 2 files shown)


src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
```diff
@@ -461,7 +461,9 @@ private void doFrequencySearchForRound(RaptorState inputState, RaptorState outpu
 
                 for (int frequencyEntryIdx = 0; frequencyEntryIdx < schedule.headwaySeconds.length; frequencyEntryIdx++) {
                     int originalPatternIndex = originalPatternIndexForFrequencyIndex[patternIndex];
-                    int offset = offsets.offsets.get(originalPatternIndex)[tripScheduleIndex][frequencyEntryIdx];
+
+                    int offset = computeDeterministicUpperBound ? -1 : offsets.offsets.get(originalPatternIndex)[tripScheduleIndex][frequencyEntryIdx];
```
abyrd: Here we are giving a variable what appears to be a special or sentinel value of -1, but in reality that value will never be read. The variable `offset` is only read in one place, and only when we are computing true randomized schedules rather than the "deterministic upper bound" before the actual randomized monte carlo searches. So a ternary operator is not appropriate here, we want to make assignment to this variable `offset` and ideally even the existence of the variable conditional upon it needing to be used at all. So I think a better solution is to change the lines a bit farther down where we use another ternary operator `int newBoardingDepartureTimeAtStop = computeDeterministicUpperBound ?`, performing the declaration and assignment of `offset` right before `getRandomFrequencyDepartureTime` is called, inside a conditional.

PR #364, opened 2017-11-10
Analysis Worker Optimizations

abyrd:
These changes are all optimizations that skip unnecessary calculation, especially useful for large regional analyses in places with a lot of water.


abyrd:
We've been using this for a while in large regional analyses and it works great. I think it's ready to merge.

PR #366, opened 2017-11-15
A callback for the street router to look up dynamic travel times

michaz:


src/main/java/com/conveyal/r5/point_to_point/builder/PointToPointQuery.java
```diff
@@ -5,10 +5,7 @@
 import com.conveyal.r5.api.util.ProfileOption;
 import com.conveyal.r5.api.util.StreetSegment;
 import com.conveyal.r5.profile.*;
-import com.conveyal.r5.streets.ParkRideRouter;
-import com.conveyal.r5.streets.Split;
-import com.conveyal.r5.streets.StreetRouter;
-import com.conveyal.r5.streets.VertexStore;
+import com.conveyal.r5.streets.*;
```
ansoncfit: We prefer to avoid wildcard imports.  See https://github.com/conveyal/JavaStyle

src/main/java/com/conveyal/r5/streets/EdgeStore.java
```diff
@@ -1122,4 +1121,13 @@ public void forEachTemporarilyAddedOrDeletedEdge (IntConsumer consumer) {
         }
     }
 
+    public static class DefaultTravelTimeCalculator implements TravelTimeCalculator {
+
+        @Override
+        public float getTravelTimems(Edge edge, int durationSeconds, StreetMode streetMode, ProfileRequest req) {
```
ansoncfit: At first I thought this was getTravelTimes with a typo.  Maybe rename to getTravelTimeMilliseconds, to avoid confusion and stay consistent with naming elsewhere (e.g. https://github.com/conveyal/r5/blob/dev/src/main/java/com/conveyal/r5/streets/ParkRideRouter.java#L54)?

abyrd:
The checks are failing because it's trying to import signing certificates using encrypted variables that are not available when building PRs from outside our organization. We need to add conditionals in travis.yml around the commands that import the keys, so that occurs in exactly the same situations where we do a deploy instead of a simple build.

abyrd:
Now it's making it all the way through the build but still attempting to sign the artifacts even through it's not deploying them.

Issue #335, opened 2017-09-04
Unexpected Island Pruning for SF Bay Bridge


colinsheppard:
Hi there, we are using R5 on a subset of the San Francisco OSM network that includes the financial district, and some of the Bay Bridge including Treasure Island.

When we build the R5 network, the TarjanIslandPruner removes the Bay Bridge along with the motorway links through the city all the way up until the surface streets. See attached image where red indicates the links that have been pruned.

![image](https://user-images.githubusercontent.com/333206/30036835-5d0e872c-916b-11e7-8e6e-a9e6d687b2d7.png)

I've confirmed that during execution of the Tarjan algorithm, the links in red appear to be identified as 1-vertex strong components. 

I understand that 1-way dead ends get pruned, so it would make sense if the bridge where to be pruned on the edge of the area of detail, but not consistently throughout the entire bridge/highway system as there are multiple on-ramps and off-ramps which would form a graph cycle from any interior point on the highway to any other.

In fact, because we used completeRelations=true during clipping of the network, the actual network being proceed by R5 extends far beyond the area of detail (see second attached image) so there are even more opportunities for cycles.

![image](https://user-images.githubusercontent.com/333206/30036842-6c7ca950-916b-11e7-994e-b52efa008d4d.png)

Do you have any idea how these links could be getting pruned and how we can fix the issue?

Thanks.

colinsheppard:
In case you want to recreate our situation, here is the data we are using:

https://www.dropbox.com/s/44fnvl9n25la74w/sf-detail.osm.pbf?dl=0
https://www.dropbox.com/s/undir2ox84wv6ka/SF.gtfs.zip?dl=0

ansoncfit:
Part of this issue relates to `minComponentSize`.  Setting `MIN_SUBGRAPH_SIZE = 3` in r5/streets/StreetLayer allows pedestrian routing on Treasure Island and car routing across to SF:

![image](https://user-images.githubusercontent.com/2173529/31829217-a2ef3924-b58a-11e7-86aa-2c20038f89af.png)

![image](https://user-images.githubusercontent.com/2173529/31829256-bc029244-b58a-11e7-90c2-ebed06b90884.png)

But there are still problems.  Note the "island" around 9th/Folsom in the above, and the fact that car routing does not work from SF to Treasure Island:

![image](https://user-images.githubusercontent.com/2173529/31829410-28f43f24-b58b-11e7-9b84-cdd7100b6675.png)

[...]

PR #370, opened 2017-11-21
Island pruning

ansoncfit:
This PR addresses the original problem identified in #335 (though subsequent posts in that thread may be related to other issues).

Strong components ("islands") were being falsely identified for cars, resulting in edges being removed even though they were well connected to the whole network.

When looping through vertices, a boolean `everySuccessorExplored` is initialized as true.  Then, `if (discoveryIndex[toVertex] == -1) everySuccessorExplored[0] = false` resets this boolean to false if there are more successors to explore.  Previously, this was nested inside `if (discoveryIndex[toVertex] > discoveryIndex[vertex])`, which is often false because the `discoveryIndex` for each vertex is initialized as -1.  So a given vertex _v_ could be left with `everySuccessorExplored = true`, even if not every successor had actually been explored.  If an unexplored successor to _v_ was the only link to the sourceVertex (root), then _v_ itself would be considered the root, leading a strongComponent to be identified.

Pulling the `if (discoveryIndex[toVertex] == -1) everySuccessorExplored[0] = false` out of the if clause it had been in leads to the expected network, without edges spuriously removed:

image

### DIFF
--- a/src/main/java/com/conveyal/r5/streets/EdgeStore.java
+++ b/src/main/java/com/conveyal/r5/streets/EdgeStore.java
@@ -213,7 +213,7 @@ public class EdgeStore implements Serializable {
         NO_THRU_TRAFFIC_BIKE (11),
         NO_THRU_TRAFFIC_CAR (12),
         SLOPE_OVERRIDE (13),
-        /** Link edge, two should not be traversed one-after-another FIXME comment seems incorrect */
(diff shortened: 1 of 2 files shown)


ansoncfit:
We should add a test for this.  We can make use of a toy OSM network I made that has two very clear islands.

Issue #377, opened 2017-12-03
Holes in regional results


abyrd:
We are now accounting for the distance of a linked point/cell away from the road it is linked to. Our linking distance on the destination (opportunity grid) end is still quite high (around 1km) to account for businesses in the middle of big parking lots etc. However our linking distance on the origin end is only 300 meters (see com.conveyal.r5.streets.StreetLayer#LINK_RADIUS_METERS). The distance should be identical to that on the destination end. Currently we get jagged holes in our output that are visually disruptive and require a place to have an access road in OSM to have nonzero accessibility.
 
We don't want this linking distance to be too huge though, or we'll be calculating useless accessibility numbers out into the ocean.

We want `StreetLayer#LINK_RADIUS_METERS` == `LinkedPointSet#MAX_OFFSTREET_WALK_METERS` and probably to even use the same symbolic constant.

Example:
![image](https://user-images.githubusercontent.com/112871/33525063-f6192078-d827-11e7-9d10-cb103b82340a.png)

PR #379, opened 2017-12-05
Fix car linking

abyrd:
This pull request fixes some problems with car routing in Analysis that were encountered by Alex on the SBB project. The biggest one is that we weren't providing the on-street mode of travel to the linking method at the origin point, so it could link to roads that were not traversable by the chosen mode.

A more subtle problem for which we do not yet have a complete solution is that ways can have different permissions in different directions, the most obvious example being parallel one-way OSM ways making up "dual carriageways" and boulevards.

There are multiple possible approaches, e.g. linking only to ways that are traversable in both directions, or linking to ways that are traversable in at least one direction. The approach in this PR works better but could still use refinement over time.

### DIFF
--- a/src/main/java/com/conveyal/r5/streets/Split.java
+++ b/src/main/java/com/conveyal/r5/streets/Split.java
@@ -95,34 +95,36 @@ public class Split {
         Split curr = new Split();
         Split best = new Split();
         candidateEdges.forEach(e -> {
-
             curr.edge = e;
             edge.seek(e);
-            //LOG.info("Considering OSM way {}", edge.getOSMID());
-            //Skip Link edges those are links between transit stops/P+R/Bike share vertices and graph
-            //Without this origin or destination point can link to those edges because they have ALL permissions
-            //and route is never found since point is inaccessible because edges leading to it don't have required permission
+
+            // Do not consider linking to edges that are links to streets from transit stops, P+Rs, and bike shares.
+            // These edges allow all modes to traverse, but may be connected to roads with more restrictive permissions.
+            // On a given edge pair both directions will have the same flag.
             if (edge.getFlag(EdgeStore.EdgeFlag.LINK)) return true;
 
-            if (!edge.allowsStreetMode(streetMode)) {
-                //LOG.info("rejecting edge {} (OSM way {} forward)", edge.edgeIndex, edge.getOSMID());
+            // If either direction of the current edge doesn't allow the specified mode of travel, skip it.
+            // It is arguably better to skip it only if BOTH directions forbid the specified mode (see commented block
+            // below). This system has odd effects in areas with lots of one-way streets or divided roads.
+            // TODO Really, we want to allow linking to two different edge-pairs in such cases but that is more complex.
+            // Do not consider linking to edges that are not marked "linkable". This excludes e.g. tunnels and motorways.
+            if (!edge.allowsStreetMode(streetMode) || !edge.getFlag(EdgeStore.EdgeFlag.LINKABLE)) {
                 return true;
             }
             edge.advance();
-            if (!edge.allowsStreetMode(streetMode)) {
-                //LOG.info("rejecting edge {} (OSM way {} backward)", edge.edgeIndex, edge.getOSMID());
+            if (!edge.allowsStreetMode(streetMode) || !edge.getFlag(EdgeStore.EdgeFlag.LINKABLE)) {
                 return true;
             }
             edge.retreat();
 
-//            if (!edge.allowsStreetMode(streetMode)) {
-//                // The edge does not allow forward traversal with the specified mode, try the backward edge.
-//                edge.advance();
-//                // If backward traversal is also not allowed, skip this edge and try the next one.
-//                if (!edge.allowsStreetMode(streetMode)) return true;
-//            }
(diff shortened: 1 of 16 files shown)


src/main/java/com/conveyal/r5/streets/Split.java
```diff
@@ -98,16 +98,31 @@ public static Split find (double lat, double lon, double searchRadiusMeters,
...
+            edge.retreat();
+
+//            if (!edge.allowsStreetMode(streetMode)) {
+//                // The edge does not allow forward traversal with the specified mode, try the backward edge.
+//                edge.advance();
+//                // If backward traversal is also not allowed, skip this edge and try the next one.
+//                if (!edge.allowsStreetMode(streetMode)) return true;
+//            }
+            // FIXME we are splitting only one direction, we actually want to split the two directions separately, so either of distance 1 or distance 2 can be infinite.
+            // FIXME we should probably entirely skip over motorways and link only surface streets.
```
ansoncfit: Using the `LINKABLE` flag?

src/main/java/com/conveyal/r5/streets/Split.java
```diff
@@ -98,16 +98,31 @@ public static Split find (double lat, double lon, double searchRadiusMeters,
...
+                return true;
+            }
+            edge.advance();
+            if (!edge.allowsStreetMode(streetMode)) {
+                //LOG.info("rejecting edge {} (OSM way {} backward)", edge.edgeIndex, edge.getOSMID());
+                return true;
+            }
+            edge.retreat();
+
+//            if (!edge.allowsStreetMode(streetMode)) {
```
ansoncfit: Is this commented out code duplicating the above?

src/main/java/com/conveyal/r5/streets/StreetLayer.java
```diff
@@ -94,6 +94,8 @@
      */
     public static final double LINK_RADIUS_METERS = 300;
```
ansoncfit: Consider increasing?

https://github.com/conveyal/r5/issues/377

src/main/java/com/conveyal/r5/streets/StreetLayer.java
```diff
@@ -1308,8 +1310,8 @@ public int createAndLinkVertex (double lat, double lon) {
      */
     public Split findSplit(double lat, double lon, double radiusMeters, StreetMode streetMode) {
         Split split = null;
-        if (radiusMeters > 300) {
-            split = Split.find(lat, lon, 150, this, streetMode);
+        if (radiusMeters > MAX_LINK_RADIUS_METERS) {
```
ansoncfit: I might rename `MAX_LINK_RADIUS_METERS` to `FIRST_SEARCH_RADIUS` or something like that.  If `radiusMeters` exceeds this value, a split can still be found below, which means it's not really the maximum.

The comment above could also be clarified, to require less thinking about what "The search radius can be specified freely here because we use this function to link transit stops to streets but also to link pointsets to streets, and currently we use different distances for these two things" means.

ansoncfit:
Oh, the relevant change was committed to dev, and I didn't notice looking at the diff.

https://github.com/conveyal/r5/commit/aebf4fbcb6609950d14ae3f3cf90d726b45ff447

src/main/java/com/conveyal/r5/streets/Split.java
abyrd: It's not exactly duplicating the logic, the commented out section is providing the other alternative implementation, which accepts an edge if it is car/bike traversable in either direction. The "real" solution is a more complicated set of heuristics to link to multiple roads in the event that they are one-way or dual-carriageway.

src/main/java/com/conveyal/r5/streets/Split.java
abyrd: `LINKABLE` seems to be currently used just for linking park-and-ride lots, but yes I'd say its use should be extended to linking origins and destinations. The set of tags checked seems correct: `if (!(way.hasTag("tunnel", "yes") || way.hasTag("covered", "yes") || way.hasTag("highway", "motorway")))`. Let's make another ticket for this because we need to test the results thoroughly before releasing such a change.

src/main/java/com/conveyal/r5/streets/StreetLayer.java
abyrd: Agreed, as mentioned in #377 we should just use one symbolic constant for linking distance on both ends of routing. We've decided to make this change later though so we can get a release out, and produce multiple worker releases with different characteristics in case a client needs them.

src/main/java/com/conveyal/r5/streets/StreetLayer.java
abyrd: Definitely agree, this needs to be clarified. It took me 5 minutes to even understand why I did this.

abyrd:
Correct, I apparently pushed the main fix directly to the dev branch.

abyrd:
@ansoncfit as agreed, I have pushed a fix for the identifier naming issue you mentioned in the review. If you approve and merge this PR to dev, I will then merge to master and perform a release. Subsequently I will address the other PR comments in other commits.

PR #395, opened 2018-01-10
Enable static sites

abyrd:
This PR makes some additional changes so that we can actually signal to the workers that they should produce travel time surfaces and save them to s3 when calculating regional analysis tasks. This provides the files that will be used by the next version of the static sites UI.

Requires a few changes in analysis-backend (in a branch with the same name) to actually trigger the behavior.


ansoncfit:
I believe this was going to be superseded by https://github.com/conveyal/r5/tree/eliminate-broker.  Is that branch ready for review?

abyrd:
Correct @ansoncfit. I've got another branch/PR that supersedes this one: #399

PR #397, opened 2018-01-25
Change method name to correct unit.

michaz:
Those are seconds, not milliseconds. My mistake, probably due to proximity to a variable ending in "ms", which means meters per second. (Code is correct.)


abyrd:
Thanks for the fix!

PR #399, opened 2018-02-02
Analysis Simplifications: Broker and Worker

abyrd:
This PR greatly simplifies several parts of Conveyal Analysis. The goal is to make the system more stable and maintainable, while paving the way for long-requested features. We want to eliminate deadlocks or other connection failures in the broker and make its memory consumption more predictable (near-constant if possible). The broker had mutated into a very tangled collection of code that matched up incoming HTTP connections from workers and the server, then had to deal with all sorts of problematic states.

These changes also make major progress toward allowing deployment of Analysis locally or in alternate hosting environments.

This PR makes breaking changes to the interface between workers and the server. It will make all past worker versions unusable. We should make sure to stabilize that new interface on staging before releasing this to production, unless we want to just release the server and worker in lockstep and not allow custom version selection.

This PR is tightly integrated with changes to analysis-backend in https://github.com/conveyal/analysis-backend/pull/113.

1. The Broker:

1a. The broker no longer stores separate objects for each task in a job. It stores one template task and produces the other tasks by cloning and modifying the template. This could eventually be taken even further by retaining the template tasks on the worker and cloning them there, with the broker only handing out the initial job description and task numbers to the workers. But the current system is an incremental change that targets the most significant improvements on the central server, where memory is often eaten up by the presence of multiple jobs. I don't really think the time spent doing the same thing on the workers would be very beneficial.

1b. The broker is no longer a separate process, and not even a separate thread. It's just an object within the Analysis server that is written to and read from using synchronized methods called from HTTP API handler methods, or directly by the analysis-backend itself. Therefore *the broker is not even in the R5 project anymore*. After it was refactored I moved it over to analysis-backend. But of course that broker requires a bunch of other worker changes here in R5. This change removes all overhead and version-matching complexity of serializing jobs to JSON and deserializing them back into the broker, then serializing them again to send them to workers.

2. Batch job tasks:

2a. Workers still pull batch tasks from the broker, but they short-poll and wait instead of long polling. The long polling system was originally created to allow single point requests to be seen immediately by the workers, and is no longer relevant. This eliminates the need to hold connections open and detect connection closure, which eliminates all sorts of fragile custom code. Every HTTP request returns a result immediately.

2b. Workers now return their regional results directly to the broker when polling for more work. This eliminates the need for an SQS queue to hold the results, decoupling us from AWS. The workers asynchronously poll for more work before they run out of tasks to work on, so even if there is contention for broker resources the worker should not block and will continue doing its job. The SQS queue appears to have been a way to work around the IOPS limits on our analysis server, whose IO would completely lock up after regional jobs created large files. We are no longer creating such big files and should shift to a server with provisioned IOPS anyway.

3. Single-point tasks: We now push single-point tasks to an appropriate worker (rather than pull them via long-poll), and single-point requests don't go through the broker at all. The analysis-backend just proxies them to a worker that has the right graph loaded. This completely eliminates our most lockup-prone code and means that our API is expressed in a very classic way as a collection of simple imperative HTTP handler methods.

4. Workers:

4a. Producing static sites. Anson's work on producing static site files from within regional workers is fully enabled here. There is a new boolean variable `makeStaticSite` in the analysisRequest that will cause travel time surfaces and job metadata to be saved on S3 when enabled on a regional job.

4b. Bootstrapping support is removed. The bootstrapping process estimates variation in the accessibility indicator due to the fact that we're using only a sample of possible schedules rather than testing them exhaustively. However it is only relevant when frequency-based lines are in use (which is often not the case for many of our users) and produces very voluminous results, whose storage had taken priority over the very useful feature of calculating indicators for many travel time cutoffs, percentiles, and destination sets at once. We have decided to use these statistical scenario comparison concepts, including the bootstrapping process, as a way to determine appropriate numbers of Monte Carlo draws and thresholds for significant change caused by scenarios, but not as something we will expose to customers. This calibration and sensitivity testing will be carried out by us as needed, and not on every analysis. Eliminating bootstrapping from this code base drastically simplifies it and increases maintainability.

4c. Calculation methods have been repeatedly refactored to produce more conventional functions that transform an input type into an output type, rather than tail-calling layers onto the stack. The previous structure seems to have grown out of a streaming approach intended to handle the very large data sets produced by the bootstrapping process.


src/main/java/com/conveyal/r5/analyst/TravelTimeReducer.java
```diff
@@ -0,0 +1,141 @@
...
+                    // Int divide will floor; this is correct because value 0 has travel times of up to one minute, etc.
+                    // This means that anything less than a cutoff of (say) 60 minutes (in seconds) will have value 59,
+                    // which is what we want. But maybe converting to minutes before we actually export a binary format is tying
+                    // the backend and frontend (which makes use of UInt8 typed arrays) too closely.
+                    int timeMinutes = timeSeconds / 60;
+                    percentileTravelTimesMinutes[p] = timeMinutes;
+                }
+            }
+        } else {
+            throw new ParameterException("Must supply expected number of times or only one time.");
```
ansoncfit: I might use "travel time values" instead of "times".  As written, it could seem like the error message is missing a word "Must supply X expected number of times..."

src/main/java/com/conveyal/r5/analyst/cluster/AnalysisTask.java
```diff
@@ -48,26 +46,18 @@
     /** A unique identifier for this task assigned by the queue/broker system. */
     public int taskId;
 
-    /** Save results to a bucket.  If blank, the response to this task will be via the default channel (broker for
-     *  single-point requests, queue for regional requests).  Not yet compatible with UI.
+    /**
+     * Whether to save results on S3. If null or empty, the response to this task will be via the
```
ansoncfit: Is this comment still correct?  Results will be via the default channel if `makeStaticSite` is false, rather than null or empty.

src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
```diff
@@ -84,32 +88,44 @@
...
     public int dryRunFailureRate = -1;
 
-    /** How long (minimum, in milliseconds) should this worker stay alive after receiving a single point task? */
-    public static final int SINGLE_POINT_KEEPALIVE_MSEC = 15 * 60 * 1000;
+    /** How long (minimum, in milliseconds) should this worker stay alive after processing a single-point task. */
+    public static final int SINGLE_KEEPALIVE_MINUTES = 30;
 
-    /** should this worker shut down automatically */
+    /** How long (minimum, in milliseconds) should this worker stay alive after processing a regional job task. */
+    public static final int REGIONAL_KEEPALIVE_MINUTES = 1;
```
ansoncfit: milliseconds?

src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
```diff
@@ -84,32 +88,44 @@
      */
     public int dryRunFailureRate = -1;
 
-    /** How long (minimum, in milliseconds) should this worker stay alive after receiving a single point task? */
-    public static final int SINGLE_POINT_KEEPALIVE_MSEC = 15 * 60 * 1000;
+    /** How long (minimum, in milliseconds) should this worker stay alive after processing a single-point task. */
+    public static final int SINGLE_KEEPALIVE_MINUTES = 30;
```
ansoncfit: milliseconds?

src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
```diff
@@ -212,25 +229,50 @@ public AnalystWorker(Properties config, TransportNetworkCache cache) {
...
+     * We only need to consider shutting them down once an hour. Leaving them alive for the rest of the hour provides
+     * immediate compute capacity in case someone starts another job.
+     * So a few minutes before each hour of uptime has elapsed, we check how long the machine has been idle.
+     */
+    public void considerShuttingDown() {
+        long now = System.currentTimeMillis();
+        if (now > nextShutdownCheckTime && autoShutdown) {
+            // Check again exactly one hour later (assumes billing in one-hour increments)
+            nextShutdownCheckTime += 60 * 60 * 1000;
+            if (now > lastSinglePointTime + (SINGLE_KEEPALIVE_MINUTES * 60 * 1000) &&
```
ansoncfit: conversion to ms happens here, so above comments should be revised.

src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
```diff
@@ -212,25 +229,50 @@ public AnalystWorker(Properties config, TransportNetworkCache cache) {
...
+                        "shutting down.", SINGLE_KEEPALIVE_MINUTES, REGIONAL_KEEPALIVE_MINUTES);
+                // Stop accepting any new single-point requests while shutdown is happening.
+                // TODO maybe actively tell the broker this worker is shutting down.
+                sparkHttpService.stop();
+                try {
+                    Process process = new ProcessBuilder("sudo", "/sbin/shutdown", "-h", "now").start();
+                    process.waitFor();
+                } catch (Exception ex) {
+                    LOG.error("Unable to terminate worker", ex);
+                    // TODO email us or something
```
ansoncfit: ☎️

src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
```diff
@@ -212,25 +229,50 @@ public AnalystWorker(Properties config, TransportNetworkCache cache) {
...
+     */
+    public void considerShuttingDown() {
+        long now = System.currentTimeMillis();
+        if (now > nextShutdownCheckTime && autoShutdown) {
+            // Check again exactly one hour later (assumes billing in one-hour increments)
+            nextShutdownCheckTime += 60 * 60 * 1000;
+            if (now > lastSinglePointTime + (SINGLE_KEEPALIVE_MINUTES * 60 * 1000) &&
+                now > lastRegionalTaskTime + (REGIONAL_KEEPALIVE_MINUTES * 60 * 1000)) {
+                LOG.info("Machine has been idle for at least {} minutes (single point) and {} minutes (regional), " +
+                        "shutting down.", SINGLE_KEEPALIVE_MINUTES, REGIONAL_KEEPALIVE_MINUTES);
```
ansoncfit: Nice.  This will be a welcome improvement from shutting down if the worker happens to be idle at :55 past startup.

src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
```diff
@@ -524,75 +538,39 @@ public void reportTaskErrors(int taskId, int httpStatusCode, List<TaskError> tas
...
-     * pointsets-bucket             S3 bucket in which pointsets are stored
-     * auto-shutdown                Should this worker shut down its machine if it is idle (e.g. on throwaway cloud instances)
-     * statistics-queue             SQS queue to which to send statistics (optional)
-     * initial-graph-id             The graph ID for this worker to start on
+     * broker-address     address of the broker, without protocol or port
+     * broker-port        port broker is running on, default 80.
+     * graphs-bucket      S3 bucket in which graphs are stored.
+     * pointsets-bucket   S3 bucket in which pointsets are stored
+     * auto-shutdown      Should this worker shut down its machine if it is idle (e.g. on throwaway cloud instances)
+     * statistics-queue   SQS queue to which to send statistics (optional)
```
ansoncfit: is this still used?

src/main/java/com/conveyal/r5/profile/PerTargetPropagater.java
```diff
@@ -9,230 +12,204 @@
...
  */
 public class PerTargetPropagater {
 
     private static final Logger LOG = LoggerFactory.getLogger(PerTargetPropagater.class);
 
-    /** The travel time cutoff in this regional analysis */
+    // FIXME why are these fields public?
+
+    /** The travel time cutoff in this regional analysis FIXME why would this have a default value? */
     public int cutoffSeconds = 120 * 60;
```
ansoncfit: Is this the cutoff specified in the regional request (i.e. `task.maxTripDurationMinutes`), or the maximum possible cutoff used as an optimization here?

src/main/java/com/conveyal/r5/analyst/TravelTimeReducer.java
trevorgerhardt: Should we have a company wide "definitions" document for variable naming and general discussion? We've gotten caught up with this in the past ("grid"s, "scenarios", etc.)

abyrd: Fixed. @trevorgerhardt we should at least add a glossary to the Analysis documentation.

src/main/java/com/conveyal/r5/analyst/cluster/AnalysisTask.java
abyrd: The comment is definitely wrong and I'll update it. This one is tricky though because I'm aiming for a non-dichotomous situation where you can set this flag to "true" and whatever results are produced by each task, which may be travel time or accessibility results or both, will be saved to S3.

src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
abyrd: Changed to minutes and reworded.

src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
abyrd: Also changed to minutes and reworded.

src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
abyrd: Noted. Thanks.

src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
abyrd: We need some cloudwatch dashboard magic for tracking quantities of workers.

src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
abyrd: Unused options removed.

src/main/java/com/conveyal/r5/profile/PerTargetPropagater.java
abyrd: Thanks for pointing that out. I always call the threshold travel time in the indicator travel time the "cutoff". Another place where an official glossary would be helpful. I have changed the comment.

PR #401, opened 2018-02-08
Correct write ints to OutputStreams

trevorgerhardt:
Previously we were using the base `write` method which only wrote the first 8 bits of an integer
when we were expecting the full 32. This PR corrects the writing of those ints and adds two more items to the path file headers for clarity.

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/TravelTimeComputer.java
+++ b/src/main/java/com/conveyal/r5/analyst/TravelTimeComputer.java
@@ -289,7 +289,7 @@ public class TravelTimeComputer {
                         waitTimesToStops, pathsToStops);
                 // TODO factor out identical lines in else clause
                 perTargetPropagater.travelTimeReducer = travelTimeReducer;
-                perTargetPropagater.pathWriter = new PathWriter(request, network, pathList);
+                perTargetPropagater.pathWriter = new PathWriter(request, network, pathList, egressModeLinkedDestinations.size(), transitTravelTimesToStops.length);
             } else {
                 perTargetPropagater = new PerTargetPropagater(egressModeLinkedDestinations, request,
                         transitTravelTimesToStops, nonTransitTravelTimesToDestinations, null, null, null);
--- a/src/main/java/com/conveyal/r5/analyst/TravelTimeComputer.java
+++ b/src/main/java/com/conveyal/r5/analyst/TravelTimeComputer.java
@@ -289,7 +289,7 @@ public class TravelTimeComputer {
                         waitTimesToStops, pathsToStops);
                 // TODO factor out identical lines in else clause
                 perTargetPropagater.travelTimeReducer = travelTimeReducer;
-                perTargetPropagater.pathWriter = new PathWriter(request, network, pathList, egressModeLinkedDestinations.size(), transitTravelTimesToStops.length);
+                perTargetPropagater.pathWriter = new PathWriter(request, network, pathList, egressModeLinkedDestinations.size(), iterations);
             } else {
                 perTargetPropagater = new PerTargetPropagater(egressModeLinkedDestinations, request,
                         transitTravelTimesToStops, nonTransitTravelTimesToDestinations, null, null, null);
--- a/src/main/java/com/conveyal/r5/analyst/cluster/PathWriter.java
+++ b/src/main/java/com/conveyal/r5/analyst/cluster/PathWriter.java
@@ -3,11 +3,11 @@ package com.conveyal.r5.analyst.cluster;
 import com.conveyal.r5.multipoint.MultipointDataStore;
 import com.conveyal.r5.profile.Path;
 import com.conveyal.r5.transit.TransportNetwork;
+import com.google.common.io.LittleEndianDataOutputStream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
-import java.io.OutputStream;
 import java.util.List;
 
 /**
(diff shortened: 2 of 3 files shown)


src/main/java/com/conveyal/r5/analyst/cluster/PathWriter.java
```diff
@@ -33,22 +32,26 @@
 
     OutputStream outputStream;
 
-    public PathWriter (AnalysisTask task, TransportNetwork network, List<Path> pathList) {
+    public PathWriter (AnalysisTask task, TransportNetwork network, List<Path> pathList, int destinations, int iterations) {
```
abyrd: I would call these `nDestinations` and `nIterations` to emphasize that the variables don't contain the destinations and iterations themselves, but are just counts/sizes.

src/main/java/com/conveyal/r5/analyst/cluster/PathWriter.java
```diff
@@ -61,14 +64,27 @@ public void recordPathsForTarget (int[] paths) {
...
             throw new RuntimeException(e);
         }
     }
 
+    /**
+     *
+     * @param value
+     * @throws IOException
+     */
+    private void writeInt (int value) throws IOException {
```
abyrd: This method definitely needs a comment explaining what it does and why. It seems strange to re-implement a fundamental method to write out an integer (more code we have to maintain ourselves). It also looks to me like it's writing out a big-endian representation of an int, which is the opposite of what we want. I think what you're looking for is DataOutputStream, which wraps an OutputStream to write more complex data types than bytes. But it's permanently big-endian, so for these cases we use LittleEndianDataOutputStream.

Also style-wise we avoid javadoc `@param` etc. unless we are writing a full description. Also just style-wise, this function should be declared static because its work does not depend on instance state.

src/main/java/com/conveyal/r5/analyst/cluster/PathWriter.java
```diff
@@ -33,22 +32,26 @@
...
-    public PathWriter (AnalysisTask task, TransportNetwork network, List<Path> pathList) {
+    public PathWriter (AnalysisTask task, TransportNetwork network, List<Path> pathList, int destinations, int iterations) {
         this.task = task;
         this.network = network;
         this.pathList = pathList;
         try {
             outputStream = MultipointDataStore.getOutputStream(task, task.taskId + "_paths.dat", "application/octet-stream");
             outputStream.write("PATHGRID".getBytes());
+            // In the header store the number of destinations stored and iterations used
+            writeInt(destinations);
```
abyrd: Rather than calling our own method I think we should just wrap the OutputStream in a LittleEndianDataOutputStream to write full-width ints to it. Probably 75% of this file will be zeros but we will worry about that only if it actually proves to be a problem. I'm wondering whether we should just change MultipointDataStore.getOutputStream to always return a stream that has LittleEndianDataOutput capabilities.

src/main/java/com/conveyal/r5/analyst/cluster/PathWriter.java
```diff
@@ -33,22 +32,26 @@
...
 
-    public PathWriter (AnalysisTask task, TransportNetwork network, List<Path> pathList) {
+    public PathWriter (AnalysisTask task, TransportNetwork network, List<Path> pathList, int destinations, int iterations) {
         this.task = task;
         this.network = network;
         this.pathList = pathList;
         try {
             outputStream = MultipointDataStore.getOutputStream(task, task.taskId + "_paths.dat", "application/octet-stream");
             outputStream.write("PATHGRID".getBytes());
+            // In the header store the number of destinations stored and iterations used
```
abyrd: I would mention that the iterations are "inside" the destinations, e.g. `the number of destinations and the number of iterations at each destination`.

src/main/java/com/conveyal/r5/analyst/TravelTimeComputer.java
```diff
@@ -289,7 +289,7 @@ public OneOriginResult computeTravelTimes() throws IOException {
                         waitTimesToStops, pathsToStops);
                 // TODO factor out identical lines in else clause
                 perTargetPropagater.travelTimeReducer = travelTimeReducer;
-                perTargetPropagater.pathWriter = new PathWriter(request, network, pathList);
+                perTargetPropagater.pathWriter = new PathWriter(request, network, pathList, egressModeLinkedDestinations.size(), transitTravelTimesToStops.length);
```
abyrd: There should already be a variable in this context called `iterations`, which should be clearer to read than transitTravelTimesToStops.length. Actually I'd prefer it be called `iterations` but that's another refactor.

ansoncfit:
Thanks for figuring this out, @trevorgerhardt.

Just saw that @abyrd added some comments in-line, which may render my notes below irrelevant.

The comment here asks why we need a homemade implementation for writing to a buffer:
https://github.com/conveyal/r5/blob/227f2fe86c7199cfc76ba3306e4134afa714a091/src/main/java/com/conveyal/r5/analyst/cluster/AccessGridWriter.java#L85

I think that's what prompted me to move away from the original path writing approach and introduce this bug.

What you've found for writing to an OutputStream wouldn't be relevant for writing to a buffer, right (e.g. at https://github.com/conveyal/r5/blob/5f636210619a4309664f0ebc89bca314cd4ba78d/src/main/java/com/conveyal/r5/analyst/cluster/TimeGrid.java#L108)?  

That larger question of stream vs. buffer doesn't need to be resolved here, but I wanted to note some places where I've seen related comments (e.g. on Guava LittleEndianDataOutputStream being slow: https://github.com/conveyal/r5/blob/1bf05eed8873925d2ddcf75974979efbf5a9c12c/src/main/java/com/conveyal/r5/analyst/LittleEndianIntOutputStream.java#L9) before forgetting.

abyrd:
@ansoncfit I remember Matt mentioning that the Guava LittleEndianOutputStream was ultra slow. I don't understand why that would be the case. We could use his `LittleEndianIntOutputStream` but I'd like to actually observe it being faster. The code is almost character for character the same thing as in the Guava class. I suspect he misinterpreted EC2 IOPS throttling as this method being slow, because the IOPS throttling was blocking the writes when this method was being used to write a large file.

The problems with writing ints would not apply to the buffer you mentioned because we call
`buffer.order(ByteOrder.LITTLE_ENDIAN);` and then use `writeInt`.

trevorgerhardt:
Made changes based on the feedback, let me know what you think @abyrd

PR #405, opened 2018-02-27
Updates Google's Java libs from v18 (released 2014) to 24 (released 2018)

trevorgerhardt:

Issue #404, opened 2018-02-26
Invert TimeGrid storage loop


trevorgerhardt:
We're likely to just be using a single percentile for static sites and imagine we could achieve a better compression from delta coding within an entire grid rather than within the percentiles in each pixel. One way to do this would be to write out an entire percentile's grid before moving to the next saved percentile.

Current implementation (from https://github.com/conveyal/r5/blob/dev/src/main/java/com/conveyal/r5/analyst/cluster/TimeGrid.java#L131):

```java
int i = 0;
while (i < nValues) {
  int prev = 0;
  for (int v = 0; v < nValuesPerPixel; v++, i++) {
    int curr = values[i];
    int delta = curr - prev;
    buffer.putInt(delta);
    prev = curr;
  }
}
```

Proposed implementation:

```java
for (int i = 0; i < nValuesPerPixel; i++) {
  int prev = 0; // delta code within each percentile grid
  for (int j = 0; j < width * height; j++) {
    int curr = values[j * nValuesPerPixel + i]; // assuming we use the same values array as above
    int delta = curr - prev;
    buffer.putInt(delta);
    prev = curr;
  }
}
```

PR #406, opened 2018-02-27
Invert time grid storage loop and explicitly use LittleEndianDataOutputStreams

trevorgerhardt:
Utilizes the changes made in #405 and addresses #404. Will require front-end changes in the Analysis UI and Taui to work with the new storage format.

Intellij automatically expanded the `import java.io.*;` on save...is that the style we use?

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
+++ b/src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
@@ -1,9 +1,9 @@
 package com.conveyal.r5.analyst.cluster;
 
-import com.amazonaws.regions.Region;
 import com.amazonaws.regions.Regions;
 import com.amazonaws.services.s3.AmazonS3;
 import com.amazonaws.services.s3.AmazonS3Client;
+import com.amazonaws.services.s3.AmazonS3ClientBuilder;
 import com.conveyal.r5.OneOriginResult;
 import com.conveyal.r5.analyst.GridCache;
 import com.conveyal.r5.analyst.TravelTimeComputer;
@@ -13,7 +13,10 @@ import com.conveyal.r5.common.JsonUtilities;
 import com.conveyal.r5.common.R5Version;
 import com.conveyal.r5.multipoint.MultipointDataStore;
 import com.conveyal.r5.multipoint.MultipointMetadata;
+import com.conveyal.r5.transit.TransportNetwork;
+import com.conveyal.r5.transit.TransportNetworkCache;
 import com.fasterxml.jackson.core.type.TypeReference;
+import com.google.common.io.LittleEndianDataOutputStream;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpResponse;
 import org.apache.http.client.HttpClient;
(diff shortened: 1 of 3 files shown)


abyrd:
On `import x.*`, see https://github.com/conveyal/javastyle#do-not-use-wildcard-imports

src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
```diff
@@ -433,7 +448,8 @@ public void sleepSeconds (int seconds) {
                 TravelTimeSurfaceTask timeSurfaceTask = (TravelTimeSurfaceTask) request;
                 if (timeSurfaceTask.getFormat() == TravelTimeSurfaceTask.Format.GRID) {
                     // Return raw byte array representing grid to caller, for return to client over HTTP.
-                    byteArrayOutputStream.write(oneOriginResult.timeGrid.writeGrid());
+                    oneOriginResult.timeGrid.writeGridToStream(new LittleEndianDataOutputStream(byteArrayOutputStream));
```
abyrd: The grid format itself is an array of little-endian ints, so shouldn't the LittleEndianDataOutputStream usage be encapsulated entirely within this method that writes the grid format?

src/main/java/com/conveyal/r5/analyst/cluster/AnalystWorker.java
```diff
@@ -449,8 +465,8 @@ public void sleepSeconds (int seconds) {
                 // This is a single task within a regional analysis with many origins.
                 if (request.makeStaticSite) {
                     // This is actually a time grid, because we're generating a bunch of those for a static site.
-                    OutputStream s3stream = MultipointDataStore.getOutputStream(request, request.taskId + "_times.dat", "application/octet-stream");
-                    s3stream.write(oneOriginResult.timeGrid.writeGrid());
+                    LittleEndianDataOutputStream s3stream = MultipointDataStore.getOutputStream(request, request.taskId + "_times.dat", "application/octet-stream");
```
abyrd: I guess MultipointDataStore is always returning a LittleEndianDataOutputStream because all our regional results stored o S3 are always gzipped arrays of little endian ints. I would more expect it to return a general OutputStream, and for TimeGrid to be able to write to any old stream (by wrapping it with its own LittleEndianDataOutputStream internally). But I guess it already worked the way I'm seeing here and you were just maintaining the existing functionality.

src/main/java/com/conveyal/r5/analyst/cluster/TimeGrid.java
```diff
@@ -105,40 +99,33 @@ public void setTarget(int targetIndex, int[] pixelValues) {
...
+        outputStream.write(gridType.getBytes());
+        outputStream.writeInt(version);
+        outputStream.writeInt(zoom);
+        outputStream.writeInt(west);
+        outputStream.writeInt(north);
+        outputStream.writeInt(width);
+        outputStream.writeInt(height);
+        outputStream.writeInt(nValuesPerPixel);
+
+        // Write values, delta coded
```
abyrd: This nested loop could use some clearer comments, to the effect of: If we have several different values per pixel (unually N different percentiles), we write out one flattened 2D array for the first percentile, then one flattened 2D array for the second plane and so on, delta-coding within each of these flattened 2D arrays, on the principle that neighboring pixels have similar values within a given percentile.

abyrd:
@trevorgerhardt I'm going to merge this and then start another PR that will use the more advanced S3 upload manager as well as the few tweaks I mentioned in this review.

Issue #407, opened 2018-02-28
FastRaptorWorker allows boarding/alighting when pickup/drop_off disallowed in GTFS


ansoncfit:
The Staten Island bus feed includes the middle of the Brooklyn-Battery Tunnel (stop_id 901171, stop_name B.B.T./Andes Road (N.B.)) on express bus patterns, with pickup and drop_off type set to 1, presumably as a timepoint.

Routing results with no modifications (in red) allow alighting there and at other similar stops along the Gowanus Expressway (isochrone in blue is with remove-stop modifications for the express buses).  
![image](https://user-images.githubusercontent.com/2173529/36818296-697323a6-1cb2-11e8-996d-58693bafa806.png)

It looks like we correctly import pickup/drop_off type (https://github.com/conveyal/r5/blob/1bf05eed8873925d2ddcf75974979efbf5a9c12c/src/main/java/com/conveyal/r5/transit/TripPattern.java#L95), but then do not check them in FastRaptorWorker (e.g. https://github.com/conveyal/r5/blob/1bf05eed8873925d2ddcf75974979efbf5a9c12c/src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java#L371)

PR #409, opened 2018-03-01
Disallow boarding/alighting when pickup/drop_off type is none

ansoncfit:
I tested this using the example described in #407, and it worked.

We should probably do a bit more testing to make sure there are not unintended side effects.


ansoncfit:
I just added the suggested clarifications to gtfs-lib: https://github.com/conveyal/gtfs-lib/commit/e1fd0652b96c5943f423778e81f8f02d0c1025c9

PR #413, opened 2018-03-20
Hardcode percentiles (DO NOT MERGE)

ansoncfit:
A user needs accessibility values given 15th percentile travel times.  This hack will display the right values in the UI.


abyrd:
For reference: note that this hack only changes the percentiles for "high priority" (AKA single point) tasks and does not affect regional analyses.

PR #420, opened 2018-04-20
Limit shapefile size

ansoncfit:
We've had multiple instances of users locking up the server by uploading geometries that are too large.  These large geometries usually indicate data errors (e.g. features at null island) or incorrect projections.

This PR should help avoid locking up by throwing explicit exceptions if:
- The shapefile projection is unrecognized
- The area of the overall shapefile bounding box exceeds 250,000 sq. km.
- The area of any feature exceeds 0.01 sq. degrees.  The individual features are processed as WGS84, so to avoid doing a large number of transformations, I'm proposing to use an area limit in square degrees.  This should be fine unless we start doing accessibility analyses in Antarctica.

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/Grid.java
+++ b/src/main/java/com/conveyal/r5/analyst/Grid.java
@@ -103,7 +103,7 @@ public class Grid {
     public final double[][] grid;
 
     /** Maximum area allowed for the bounding box of an uploaded shapefile -- large enough for New York State.  */
-    private static final double MAX_BOUNDING_BOX_AREA_SQ_KM = 250000;
+    private static final double MAX_BOUNDING_BOX_AREA_SQ_KM = 250 000;
 
     /** Maximum area allowed for features in a shapefile upload */
     double MAX_FEATURE_AREA_SQ_DEG = 0.01;
--- a/src/main/java/com/conveyal/r5/analyst/Grid.java
+++ b/src/main/java/com/conveyal/r5/analyst/Grid.java
@@ -105,6 +105,9 @@ public class Grid {
     /** Maximum area allowed for the bounding box of an uploaded shapefile -- large enough for New York State.  */
     private static final double MAX_BOUNDING_BOX_AREA_SQ_KM = 250000;
 
+    /** Maximum area allowed for features in a shapefile upload */
+    double MAX_FEATURE_AREA_SQ_DEG = 0.01;
+
     /**
      * @param zoom web mercator zoom level for the grid.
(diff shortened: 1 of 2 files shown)


src/main/java/com/conveyal/r5/analyst/Grid.java
```diff
@@ -102,6 +102,12 @@
     /** The data values for each pixel within this grid. */
     public final double[][] grid;
 
+    /** Maximum area allowed for the bounding box of an uploaded shapefile -- large enough for New York State.  */
+    private static final double MAX_BOUNDING_BOX_AREA_SQ_KM = 250000;
```
abyrd: Both the Netherlands and Switzerland are around 40k km2, and their bounding boxes could be up to twice that big. So this is big enough to contain shapefiles for those locations. I did misread this at first at 25k km2. Tip: Java lets you put spaces in number, so you can write `MAX_BOUNDING_BOX_AREA_SQ_KM = 250 000;`. We should make this standard for constants >= 10k in our Java code. I'll update the style guide.

src/main/java/com/conveyal/r5/util/ShapefileReader.java
```diff
@@ -77,6 +82,14 @@ public ReferencedEnvelope getBounds () throws IOException {
         return source.getBounds();
     }
 
+    public double getAreaSqKm () throws IOException, TransformException, FactoryException {
+        CoordinateReferenceSystem webMercatorCRS = CRS.decode("EPSG:3857");
+        MathTransform webMercatorTransform = CRS.findMathTransform(crs, webMercatorCRS, true);
+        Envelope mercatorEnvelope = JTS.transform(getBounds(), webMercatorTransform);
```
abyrd: This is going to grossly inflate the size in m2 of bounding boxes at high latitudes right? I guess we can tolerate that if the upper limit on size is permissive enough.

src/main/java/com/conveyal/r5/analyst/Grid.java
ansoncfit: 👍

abyrd: Just curious @ansoncfit, didn't your IDE show an error with the space in the constant? I guess you would expect Travis CI to tell you if there was a compile error but it's failing due to a misconfiguration - I'll fix that.

ansoncfit: It did show an error.  I thought it was just a style setting I hadn't changed, but I should have paid more attention to it.

Issue #422, opened 2018-04-25
Unexpected decreases in accessibility


ansoncfit:
Using R5 v3.4.1, comparing a scenario with only add-trip modifications to a baseline:
![image](https://user-images.githubusercontent.com/2173529/39262187-af8ebfa2-488c-11e8-9d7a-3b4aeca166cc.png)

![image](https://user-images.githubusercontent.com/2173529/39262206-be837912-488c-11e8-85eb-864a33b54516.png)

Do the red areas (reachable in 20 minutes in the baseline, but not in the scenario with added service) suggest a bug related to linking the newly added stops?   

Another observation, using v3.2.0-70-g84bcb76 and a different scenario with only add-trip modifications:

There are 3 grid cells that show a decrease of 31 jobs reachable:

![image](https://user-images.githubusercontent.com/2173529/39265967-ad54fc2e-4896-11e8-9bd7-1cf765bcf13e.png)

![image](https://user-images.githubusercontent.com/2173529/39266434-ff5d53ee-4897-11e8-8084-4f17b27a7b08.png)

abyrd:
My first thought @ansoncfit, does the baseline include any frequency routes, and what is the number of Monte Carlo draws? Not sure if this cause has already been excluded, but when frequency routes are handled via Monte Carlo selection of randomized schedules, there is no mechanism to ensure that the same randomized schedules are used for the two different scenarios being compared. This means that the baseline could get a slightly advantageous selection of schedules relative to the comparison scenario. The travel times, and therefore the accessibility indicator value, contain a certain amount of noise which is inversely related to the Monte Carlo sample size. Small magnitude changes must be filtered out (hence the whole bootstrap process and probability of improvement computations).

However, if the baseline doesn't contain any frequency routes then adding a line should only cause improvements. In that case the linking process is indeed a likely suspect.

abyrd:
The isochrones are also depending on interpolation between grid cells. The position of the isochrone at a scale smaller than one cell is not necessarily meaningful. However, if I'm not mistaken this is linear  interpolation with none of the overshoot problems that can occur in higher-degree interpolation. Adding a line to a non-frequency network should cause all travel times to be <= their baseline values, and intuitively it seems to me that linearly interpolated isochones for the baseline grid should necessarily fall entirely within those for a grid of >= values (not convinced of this at proof-level though).

ansoncfit:
The baseline network has no frequency routes, so Monte Carlo noise should not be a factor here.

[...]

PR #427, opened 2018-05-09
Give link edges a length

ansoncfit:
This PR addresses a previously identified TODO: giving a length to the edges that link newly created stops to the street network.  

Before this PR, these links were given a length of 1 mm.  In edge cases, this could shift the starting vertex of the street search to an unexpected location (see #422 )

### DIFF
--- a/src/main/java/com/conveyal/r5/streets/StreetLayer.java
+++ b/src/main/java/com/conveyal/r5/streets/StreetLayer.java
@@ -1290,15 +1290,9 @@ public class StreetLayer implements Serializable, Cloneable {
             return -1; // Unlinked
         }
 
-        double streetVertexLat = vertexStore.fixedLats.get(streetVertex)/FIXED_FACTOR;
-        double streetVertexLon = vertexStore.fixedLons.get(streetVertex)/FIXED_FACTOR;
-
-        ArrayList<Node> edgeNodes = new ArrayList<>();
-        edgeNodes.add(new Node(lat, lon));
-        edgeNodes.add(new Node(streetVertexLat,streetVertexLon));
-        int length = getEdgeLengthMillimeters(edgeNodes);
+        int length_mm = (int) (GeometryUtils.distance(lat,lon, vertexStore.getCursor(streetVertex).getLat(), vertexStore.getCursor(streetVertex).getLon())*1000);
         // Set OSM way ID is -1 because this edge is not derived from any OSM way.
-        Edge e = edgeStore.addStreetPair(stopVertex, streetVertex, length, -1);
+        Edge e = edgeStore.addStreetPair(stopVertex, streetVertex, length_mm, -1);
 
         // Allow all modes to traverse street-to-transit link edges.
         // In practice, mode permissions will be controlled by whatever street edges lead up to these link edges.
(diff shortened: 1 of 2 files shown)


src/main/java/com/conveyal/r5/streets/StreetLayer.java
```diff
@@ -1289,9 +1290,15 @@ public int createAndLinkVertex (double lat, double lon) {
             return -1; // Unlinked
         }
 
-        // TODO give link edges a length.
+        double streetVertexLat = vertexStore.fixedLats.get(streetVertex)/FIXED_FACTOR;
+        double streetVertexLon = vertexStore.fixedLons.get(streetVertex)/FIXED_FACTOR;
```
abyrd: There's a function defined for this operation: com.conveyal.r5.streets.VertexStore#fixedDegreesToFloating(int). We should use the symbolic function name as much as possible to increase readability and decrease the chance of a typo performing the calculation differently in one place. I see that is not the case throughout the code but we should aim for it. 

There is an even more encapsulated way of retrieving these values: `new VertexStore.Vertex(streetVertex).getLat()` and `getLon()`.

src/main/java/com/conveyal/r5/streets/StreetLayer.java
```diff
@@ -1289,9 +1290,15 @@ public int createAndLinkVertex (double lat, double lon) {
...
+
+        ArrayList<Node> edgeNodes = new ArrayList<>();
+        edgeNodes.add(new Node(lat, lon));
+        edgeNodes.add(new Node(streetVertexLat,streetVertexLon));
+        int length = getEdgeLengthMillimeters(edgeNodes);
```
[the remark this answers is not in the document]
abyrd: So in the end that would give something like: `GeometryUtils.distance(lat, lon, streetVertex.getLat(), streetVertex.getLon());`

src/main/java/com/conveyal/r5/streets/StreetLayer.java
abyrd: I will say though that while fixed-point coordinates are a good optimization under certain circumstances, R5 may have outgrown them. Using ints instead of doubles halves the storage space and speeds up certain operations (as well as needing to pull less things in from memory) but our edges now have a lot more data attached to them than they originally did, so the savings may be proportionately smaller. We can address that another time...

src/main/java/com/conveyal/r5/streets/StreetLayer.java
```diff
@@ -1289,9 +1290,9 @@ public int createAndLinkVertex (double lat, double lon) {
             return -1; // Unlinked
         }
 
-        // TODO give link edges a length.
+        int length_mm = (int) (GeometryUtils.distance(lat,lon, vertexStore.getCursor(streetVertex).getLat(), vertexStore.getCursor(streetVertex).getLon())*1000);
```
abyrd: This is creating two vertex cursor instances - again this is an arguably minor optimization, but it's generally better to avoid creating excess objects in heavily used functions.

My take is that it's worth constructing the Vertex cursor because that's our project's standard way of accessing the details of vertices, but we don't want to make extra ones. Who knows, the JVM might be able to optimize this away, but nonetheless I think the meaning is clearer if you make one such object representing the vertex, then read off the lat and lon from the same object.

PR #428, opened 2018-05-14
Add ability to route to a set of discrete points and enable bikeshare as access + egress mode.

samaratrilling:
* Add ability to route to a set of discrete points with IDs in addition to pixels on a grid.
* Add BikeShare importer for GBFS.
* Implement bikeshare as an access and egress option in one-to-many routing.
* Add information about transit edges to response to enable better front-end visualization.

This code powers the backend of totx.sidewalklabs.com.

Very open to suggestions for splitting this up into multiple PRs or only merging part of it back to master - whatever is most useful to the community.


pom.xml
```diff
@@ -367,7 +374,7 @@
         <dependency>
             <groupId>com.conveyal</groupId>
             <artifactId>osm-lib</artifactId>
-            <version>1.3.0</version>
+            <version>1.1.1</version>
```
abyrd: Why has osm-lib been changed to an older version? We see that this is somehow related to tests failing, but do you have any additional information?

pom.xml
```diff
@@ -358,6 +360,11 @@
             <version>24.0-jre</version>
         </dependency>
         <!-- Conveyal library for loading and storing GTFS transit data. -->
+        <dependency>
+            <groupId>com.google.protobuf</groupId>
```
abyrd: Can you clarify what protobuf is being used for?

pom.xml
```diff
@@ -415,10 +437,20 @@
             <artifactId>jackson2-geojson</artifactId>
             <version>0.8</version>
         </dependency>
+        <dependency>
+            <groupId>org.mapdb</groupId>
+            <artifactId>elsa</artifactId>
```
abyrd: I don't see elsa being used anywhere in this pull request. I believe I was experimenting with it sometime in the past. Is it possible that this PR is reverting some changes we have made since your branch diverged?

pom.xml
```diff
@@ -415,10 +437,20 @@
...
+            <version>3.0.0-M5</version>
+        </dependency>
         <dependency>
             <groupId>com.axiomalaska</groupId>
             <artifactId>polyline-encoder</artifactId>
             <version>0.1</version>
         </dependency>
+        <dependency>
+          <groupId>com.google.cloud</groupId>
+          <artifactId>google-cloud-storage</artifactId>
```
abyrd: What is Google cloud storage being used for? I don't see any references to it in the diff.

src/main/java/com/conveyal/r5/GraphQLSchema.java
```diff
@@ -697,6 +699,65 @@ public GraphQLSchema(PointToPointQuery profileResponse) {
                 .build())
             .build();
 
+        transitEdgeInfoType = GraphQLObjectType.newObject()
+            .name("TransitEdgeInfo")
+            .description("Represents one transit network link")
+            .field(GraphQLFieldDefinition.newFieldDefinition()
+                .name("id")
+                .type(new GraphQLNonNull(Scalars.GraphQLString))
+                .description("transit edge id composed of stop IDs and route ID")
```
abyrd: If this needs to be unique, note that two stop IDs and a route ID will not uniquely identify a transit link - you'd probably need to use a trip ID.

src/main/java/com/conveyal/r5/analyst/TravelTimeComputer.java
```diff
@@ -42,18 +42,22 @@ public TravelTimeComputer(AnalysisTask request, TransportNetwork network, GridCa
...
-
         // The mode of travel that will be used to reach transit stations from the origin point.
         StreetMode accessMode = LegMode.getDominantStreetMode(request.accessModes);
         // The mode of travel that will be used to reach destinations from transit stations.
         StreetMode egressMode = LegMode.getDominantStreetMode(request.egressModes);
         // The mode of travel that would be used to reach the destination directly without using transit.
         StreetMode directMode = LegMode.getDominantStreetMode(request.directModes);
 
-        // The set of destinations in the one-to-many travel time calculations, not yet linked to the street network.
+        // The set of destinations in the one-to-many travel time calculations, already linked to the street network.
```
abyrd: The comment has been changed, but looking at our code, the pointset is indeed not linked to the street network at this line. We might just not have a shared base of terminology to communicate here. Our definition of linking is associating a point with the closest edge in one specific network. Here we are just fetching the pointset but not yet ensuring that it is linked.

pom.xml
```diff
@@ -381,6 +388,21 @@
             <artifactId>slf4j-api</artifactId>
             <version>1.7.25</version>
         </dependency>
+        <dependency>
```
ansoncfit: We don't see the protobuf, aws, and google cloud dependencies used in R5 (at least searching through the diff).  If they're being used transitively by another library, could they be included there instead?

src/main/java/com/conveyal/r5/analyst/TravelTimeComputer.java
```diff
@@ -81,18 +85,47 @@ public OneOriginResult computeTravelTimes() throws IOException {
...
+                    sr,
+                    true,
+                    false);
+
+                if (sr == null) {
+                    // Origin not found. Return an empty access times map, as is done by the other
+                    // conditions for other modes.
+                    // FIXME this is ugly. we should have a way to break out of the search early
+                    // (here and in other methods).
+                    // It causes regional analyses to be very slow when there are a large number
```
abyrd: This comment appears to be copied from somewhere else, it should be revised. The log message also refers to destination coordinates, while this is probably in fact an origin coordinate.

src/main/java/com/conveyal/r5/analyst/TravelTimeComputer.java
```diff
@@ -81,18 +85,47 @@ public OneOriginResult computeTravelTimes() throws IOException {
...
-            sr.quantityToMinimize = StreetRouter.State.RoutingVariable.DURATION_SECONDS;
-            sr.route();
+            if (request.directModes.contains(LegMode.BICYCLE_RENT)) {
+                if (!this.network.streetLayer.bikeSharing) {
+                    LOG.warn("Bike sharing trip requested but no bike sharing stations in the " +
+                        "streetlayer");
+                    travelTimeReducer.finish();
+                    return null;
+                }
+                PointToPointQuery bikeQuery = new PointToPointQuery(network);
```
abyrd: This does look like it would work for your use case and potentially for ours if we wanted to enable bike rental within our spatial analysis system. However it breaks a certain hierarchy that we have in this project, where point to point routing is built on top of the routing components and data structures that were developed for the analysis system. We're hesitant to call "up" to functions in the point-to-point routing system from within the one-to-many code it is built upon. The point-to-point code is expected to be a continuous work in progress to respond to end users' expectations, and we are planning to allow it to become rather ad-hoc and messy. The one-to-many code on the other hand is intended to remain relatively minimalist and "principled" since we're using it in specialist tools that need to run headless in clusters for example.

src/main/java/com/conveyal/r5/analyst/PointSetWithIds.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * A pointset that represents a list of points with IDs.
+ * For example, this is used to represent interior points of census dissemination areas, where
+ * the ID is the census dissemination area ID.
+ *
+ * This allows for doing one-to-many routing to an arbitrary list of points, rather than isochrone
+ * generation that calculates times to every single map pixel.
+ */
+public class PointSetWithIds extends PointSet implements Serializable {
```
abyrd: Our original design, and the one we planned to eventually go back to, was to have two subclasses of PointSet: a WebMercatorGridPointset (or more general GridPointset for arbitrary projections) and a FreeformPointSet which corresponds to your PointSetWithIds class here. The idea is for the code that intersects the pointset with the travel time surface to not care which type it was using. FreeformPointSets were removed in https://github.com/conveyal/r5/pull/338/files#diff-9556fc113307c09aae33464b22cedc6c because we decided to concentrate on only grids for the time being. 

The intent was to have general superclass methods for linking both types of pointsets to street networks, and store all of them (gridded or not) in the same cache of abstract PointSets.

src/main/java/com/conveyal/r5/analyst/PointSetCache.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.Map;
+
+/**
+ * A cache for a geographic PointSet, including grid-based and custom pointset-based sets.
+ * This allows pointsets to not be recreated and relinked on every analysis task.
+ * The cache does not have expiration.
+ */
+public class PointSetCache {
```
abyrd: See comment on PointSetWithIds below. In our eventual design we intend to have a single abstract PointSet class or interface and the fact that it's gridded or free-form would be an implementation detail, with gridded points having implicit coordinates.

pom.xml
ansoncfit: I noticed compatibility issues with osmosis-osm-binary v0.46 via osm-lib v1.3.0.  The newly released osm-lib v1.4.0 downgrades osmosis-osm-binary to v0.45 and may resolve issues.

abyrd:
Closing this PR because we have since reintroduced our original FreeformPointSet design. Thanks for contributing the PR though!

Issue #373, opened 2017-12-01
Some points not linked to streets


abyrd:
Alex has reported holes in a regional analysis of Switzerland. These appear in both regional accessibility analyses and single-point analyses.

![image](https://user-images.githubusercontent.com/112871/33475537-0bf2ad7e-d67f-11e7-826b-c0520ba90eea.png)

![image](https://user-images.githubusercontent.com/112871/33475520-01f8b35e-d67f-11e7-8882-fdf2f381e2e1.png)

His analyses are with the car mode. In the regional analyses, these are probably errors linking the origin to the network. In the single-point analyses they are probably errors linking destinations.

abyrd:
Alex has confirmed that the accessibility value at all these places is zero.

Not all unconnected areas are near motorways though. if connections to motorways are not allowed, they should still be connected to something. I suspect it’s finding the motorway, realizing it’s a motorway, and then not re-trying to find a nearby surface street. The same thing could be happening in the city center: it’s in car mode, and linking to a bicycle path or pedestrian path, realizing it’s not appropriate for cars, and then not retrying.

abyrd:
Looking at com.conveyal.r5.analyst.TravelTimeComputer, only walking can work correctly 100% of the time, because we're linking to the street network before setting the routing mode. I'm tempted to make the default mode null so it will fail if we do this anywhere.

abyrd:
After making the obvious fix and telling the origin linker what mode will be used, things get much better. But there are still some holes at the origin and destination ends. Another regional analysis with the new worker:
![image](https://user-images.githubusercontent.com/112871/33482282-90d3ff8e-d698-11e7-8ac9-61813361ae32.png)
Note that there are still tiny holes around some roads, and strangely an area of low accessibility around the highway. The latter may be due to directionality issues (cells linked to only one direction of the highway).

The first part may have something to do with dead ends or service roads. I can reproduce these holes around service roads, and very interestingly the holes appear even when I drop the origin inside the hole (of course it may just be linking to a point outside the hole, and the hole is still inaccessible).

In Portland: 
![image](https://user-images.githubusercontent.com/112871/33482444-672e6358-d699-11e7-972c-26bbf9baa212.png)

abyrd:
[...]

Issue #360, opened 2017-11-06
Teleportation for walk-only trips


ansoncfit:
Note the small area of reachability near Little Nahant (northeast quadrant):

![image](https://user-images.githubusercontent.com/2173529/32462126-47936ea4-c306-11e7-8462-df824b8a29d1.png)

![image](https://user-images.githubusercontent.com/2173529/32462140-579228c2-c306-11e7-945d-891be8d8b76d.png)

Is an int overflowing somewhere?

ansoncfit:
The erroneously reachable areas seem to be near the borders of graphs.

A possibly related example, west of Portland:

![image](https://user-images.githubusercontent.com/2173529/34902070-0db6d368-f7e2-11e7-8ba9-8cde03e9d652.png)


![image](https://user-images.githubusercontent.com/2173529/34902069-00bf944c-f7e2-11e7-8fe6-4d5da82a1364.png)

ansoncfit:
Linking also seems suspect.  

Reasonable:
![image](https://user-images.githubusercontent.com/2173529/34902159-da37c77a-f7e3-11e7-81be-c09f95df0260.png)

Questionable:
![image](https://user-images.githubusercontent.com/2173529/34902165-02d8209e-f7e4-11e7-91c9-7cd37bfb28ee.png)

ansoncfit:
Starting from the middle of a racetrack:
![image](https://user-images.githubusercontent.com/2173529/34902187-52737662-f7e4-11e7-8716-8abb13a8296b.png)

ansoncfit:
These latest examples may be unrelated to the original issue.  But Marco has reported something resembling the original issue for bikes:

![image](https://user-images.githubusercontent.com/2173529/34902522-1e4fd64e-f7eb-11e7-9982-c0559370ba54.png)

abyrd:
@ansoncfit your original image and Marco's do look like they could be overflow. Since in the UI we use single bytes for travel times in minutes, anything over about 2 hours is subject to overflow so maybe we should add some additional checks.

ansoncfit:
The Portland International Raceway examples above (not int overflow issues) can be addressed with the FIXME here: https://github.com/conveyal/r5/blob/701916ee71d9f51f7115118569c648370e52bff2/src/main/java/com/conveyal/r5/streets/StreetRouter.java#L324.  The time to reach the two vertices of an origin split edge is not set, so both ends can be reached in 0 travel time (teleportation!)

PR #436, opened 2018-07-10
Splitting and linking fixes

ansoncfit:
Relevant issues: #373, #422, #360 

The main change here is adding the time and distance required to reach the first vertex of the streetLayer.

This change will make a comment in Split more salient: 
>  We don't yet handle initial and final Splits on the same edge. This can be a problem for Analysis if your starting point is on a long road without intersections. Travel times to other points along that road would be measured by going to the end of the road and back.

We could consider using something like StreetLayer.getOrCreateVertexNear() to actually perform a split at the origin.

### DIFF
--- a/src/main/java/com/conveyal/r5/profile/ProfileRequest.java
+++ b/src/main/java/com/conveyal/r5/profile/ProfileRequest.java
@@ -63,7 +63,7 @@ public class ProfileRequest implements Serializable, Cloneable {
     public int bikeTrafficStress = 4;
     
     /** The speed of driving, in meters per second. Roads from OSM use the speed limit, this is the speed used when propagating from the street network to a pointset. */
-    public float carSpeed = 11; // ~40 km/h
+    public float carSpeed = 2.22f; // ~8 km/h
 
     /** Maximum time to reach the destination without using transit in minutes */
     public int    streetTime = 60;
--- a/src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
+++ b/src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
@@ -56,10 +56,12 @@ public class LinkedPointSet implements Serializable {
      */
     public int[] edges;
 
-    /** For each point, distance from the initial vertex of the edge to the split point. */
+    /** For each point, distance from the beginning vertex of the edge geometry up to the split point for a link point,
+     * plus the distance from the link point to the split point */
     public int[] distances0_mm;
 
-    /** For each point, distance from the final vertex of the edge to the split point. */
+    /** For each point, distance from the end vertex of the edge geometry up to the split point for a link point,
+     * plus the distance from the link point to the split point*/
     public int[] distances1_mm;
 
     /** For each transit stop, the distances to nearby PointSet points as packed (point_index, distance) pairs. */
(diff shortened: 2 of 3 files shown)


ansoncfit:
Just discussed with @abyrd:
- Make it clear in user-facing documentation that we assume origin and destination are not on the same edge.  For typical travel time cutoffs (>=15 minutes) and block sizes (i.e. where it takes less than the cutoff to walk to the end of the block and back down the street), this noise will disappear.
- If we wanted to add a post-processing step, we could calculate streetMode times along the origin edge and overwrite the travel time values for the destination grid points along that edge.  But with our current grid cell resolution, this would be adding complexity that in most cases would not change results.

abyrd:
Further details of the discussion referenced by @ansoncfit above: not supporting initial and final splits on the same edge will only affect accessibility results with travel time thresholds less that the time it takes to traverse that initial edge. This system is not really designed for such cases - it is for analyzing multimodal transport to places of employment or other urban amenities, for which the usual travel time tolerance is upwards of 15 minutes, and often more like 30, 45, or 60 minutes. While it would be nice to produce perfect results in all circumstances, it's probably not worth the additional complexity to support lower travel time thresholds.

src/main/java/com/conveyal/r5/profile/ProfileRequest.java
```diff
@@ -63,7 +63,7 @@
     public int bikeTrafficStress = 4;
     
     /** The speed of driving, in meters per second. Roads from OSM use the speed limit, this is the speed used when propagating from the street network to a pointset. */
-    public float carSpeed = 11; // ~40 km/h
+    public float carSpeed = 2.22f; // ~8 km/h
```
abyrd: I believe this is the off-road driving speed, which is usually a access road / parking lot speed, so 8km/h makes sense. But is this change intentionally bundled with this batch of linking fixes?

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
```diff
@@ -56,10 +56,12 @@
      */
     public int[] edges;
 
-    /** For each point, distance from the initial vertex of the edge to the split point. */
+    /** For each point, distance from the beginning vertex of the edge geometry up to the split point for a link point,
+     * plus the distance from the link point to the split point */
```
abyrd: I'm not sure the different meanings of "link point" and "split point" will be clear to a future reader.

src/main/java/com/conveyal/r5/streets/StreetRouter.java
```diff
@@ -317,11 +317,14 @@ public boolean setOrigin (double lat, double lon) {
         // Uses weight based on distance from end vertices, and speed on edge which depends on transport mode
         float speedMetersPerSecond = edge.calculateSpeed(profileRequest, streetMode);
         startState1.weight = (int) ((split.distance1_mm / 1000) / speedMetersPerSecond);
+        startState1.durationSeconds = startState1.weight;
+        startState1.distance = split.distance1_mm;
         edge.advance();
         // Speed can be different on opposite sides of the same street
         speedMetersPerSecond = edge.calculateSpeed(profileRequest, streetMode);
         startState0.weight = (int) ((split.distance0_mm / 1000) / speedMetersPerSecond);
-        // FIXME we're setting weight but not time and distance on these states above!
```
abyrd: Ack, looks like the kind of thing that should have just been resolved at the time the FIXME comment was added. But often those FIXMEs are added while in the midst of another change or bugfix.

ansoncfit:
:tada: This PR is included in version 4.2.0 :tada:

Your **[semantic-release](https://github.com/semantic-release/semantic-release)** bot :package::rocket:

PR #442, opened 2018-08-11
Semantic release fix

evansiroky:
Sorry for the extra commits...  I branched off of master.

This PR updates maven-semantic-release to the latest version.  I refactored maven-semantic-release to act as a plugin to semantic-release since other semantic-release plugins also behave this way.  Also, I moved some stuff around in .travis.yml.  The installation of aws and semantic release now happen in before_script so that these items are cached.  Also, semantic-release is now ran before deploying to s3.  This way, when a new release is generated, that release will get uploaded to s3.  Finally, I also added an entry to the git-commit-id-plugin telling it to use lightweight tags because it looks like semantic release does not create annotated tags.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,6 +57,7 @@ after_success:
   # builds: one for the tag and one for the commit to the branch.
   - JAVADOC_TARGET=${TRAVIS_TAG:-$TRAVIS_BRANCH}
   - if [[ "$TRAVIS_PULL_REQUEST" = false ]]; then mvn javadoc:javadoc; aws s3 cp --cache-control max-age=300 --recursive /home/travis/build/conveyal/r5/target/site/apidocs s3://javadoc.conveyal.com/r5/${TRAVIS_BRANCH}/; fi
+  - yarn global add @conveyal/maven-semantic-release && maven-semantic-release
 
 sudo: true # sudo: true gets us more memory, which we need, at the expense of slower builds
 
@@ -68,6 +69,7 @@ git:
 cache:
   directories:
   - $HOME/.m2/repository
+  - $HOME/.yarn-cache
 
 notifications:
   slack: conveyal:vqrkN7708qU1OTL9XkbMqQFV
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,11 @@ before_install: |
     gpg --import --batch maven-artifact-signing-key.asc
   fi
 
+# install dependencies used in after_success here because they don't get cached if called in after_success
+before_script:
+  - pip install --user awscli && export PATH=$PATH:~/.local/bin/
+  - yarn global add @conveyal/maven-semantic-release semantic-release
+
 # Replace Travis's default build step.
 # Run all Maven phases at once up through verify, install, and deploy.
 script: |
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -107,6 +107,9 @@
                 </executions>
                 <configuration>
                     <dotGitDirectory>${project.basedir}/.git</dotGitDirectory>
+                    <gitDescribe>
+                        <tags>true</tags>
```
abyrd: Please add a comment here in the XML explaining what this setting is for. I see in the PR description that this is a response to the fact that semantic release doesn't make annotated tags, but am not sure how this option resolves the problem.

abyrd: Might it be possible to instead get semantic release to create annotated tags? I think other tools including GitHub only consider annotated tags to be releases.

.travis.yml
```diff
@@ -47,9 +52,9 @@ after_success:
   # Upload shaded JAR to S3. The un-shaded jar was deployed to our Maven repo during the main build.
   # Ideally we'd use the Travis S3 deployer here instead of installing the AWS CLI software every time.
   # See the analyst-server travis.yml for how to copy specific files to a subdirectory and deploy those.
-  - pip install --user awscli && export PATH=$PATH:~/.local/bin/
   # Upload the shaded jar. It's named with git describe, so begins with the last tag name, so begins with the letter v.
   # On the other hand, the main un-shaded artifact will begin with the project name 'r5'.
+  - semantic-release --prepare @conveyal/maven-semantic-release --publish @semantic-release/github,@conveyal/maven-semantic-release --verify-conditions @semantic-release/github --verify-release @conveyal/maven-semantic-release
```
abyrd: The comments above this line no longer apply to it - please reorder the lines and add a comment saying what this line does.

pom.xml
evansiroky: Without further going a rabit hole I can't figure out how I'd make semantic release create releases with annotated tags.  I'll add a comment saying why this was added for now.

abyrd:
I wouldn't say the non-annotated tag thing is blocking, I just want to make sure we understand it. Does github auto-create releases from tags? I've always noticed things appearing in the list of releases and tags on Github projects, even when I don't add them:
https://github.com/conveyal/r5/releases

Will it do this for all tags or only annotated tags? Or are these releases being created some other way?

It's also worth noting that our use of `git describe` to identify JARs won't work with our current system (merge branches to dev, merge dev to master to release) unless we systematically merge master back into dev after every release (because the release tags won't be ancestors of development branch commits).

evansiroky:
> I wouldn't say the non-annotated tag thing is blocking, I just want to make sure we understand it. Does github auto-create releases from tags? I've always noticed things appearing in the list of releases and tags on Github projects, even when I don't add them:
> https://github.com/conveyal/r5/releases
> 
> Will it do this for all tags or only annotated tags? Or are these releases being created some other way?
> 
> It's also worth noting that our use of `git describe` to identify JARs won't work with our current system (merge branches to dev, merge dev to master to release) unless we systematically merge master back into dev after every release (because the release tags won't be ancestors of development branch commits).

Semantic release creates a release using the github rest API.  It'll specify a commit and also upload the html/markdown that is shown on the releases page.

Regarding merging `master` into `dev` to have the ancestor show up properly, this seems to have been the standard operating procedure in the past for this repo.

abyrd:
Thanks for the clarification @evansiroky. I was wondering if it made the Github releases using the API, that's great. In that case I don't really care what kind of tags they are. We might want to change our way of identifying R5 JARs though to just the commit hash without the "last version" because it's so easy to get that wrong.

Then the master -> dev merge after a release won't be so important, unless it's just to increase the SNAPSHOT version number.

evansiroky:
So, is this ready to be merged then?

PR #452, opened 2018-10-01
doc(log): add log statement

ansoncfit:


ansoncfit:
Commit `2e7082d` makes the R5 worker log received analysis requests to Cloudwatch.  This is useful for debugging, but we don't plan to merge this into R5.

PR #453, opened 2018-10-22
Fare limits

ansoncfit:
This PR enables multiobjective searches that result in full journeys that are Pareto optimal on time or cost.  

At each McRAPTOR state, the cumulative fare paid to reach that state and the transfer allowance redeemable in future journey stages are tracked.  This approach allows for safe pruning; if we did not track transfer allowance, journeys that were ultimately optimal but, without considering transfer privileges, more expensive at intermediate points, might be prematurely pruned.

Depends on https://github.com/conveyal/analysis-backend/pull/154, which sets `maxFare` and which `inRoutingFareCalculator` to use.  As many special cases are not easy to handle with GTFS fare_attributes and fare_rules tables, we handle many of them in custom fare calculator classes (e.g. BostonInRoutingFareCalculator).

If no `inRoutingFareCalculator` is specified in the request, TravelTimeComputer uses FastRaptorWorker (as per usual).  If one is specified, a McRaptorSuboptimalPathProfileRouter is used.

Other small changes:
- Modifies TransitLayer to save GTFS fare information.
- Renames GreedyFareCalculator to InRoutingFareCalculator, to emphasize that the fare calculations happen within the McRAPTOR search.


src/main/java/com/conveyal/r5/transit/TransitLayer.java
```diff
@@ -184,6 +192,9 @@ public void loadFromGtfs (GTFSFeed gtfs, LoadLevel level) throws DuplicateFeedEx
             // This is only used while building the TransitNetwork to look up StopTimes from the same feed.
             indexForUnscopedStopId.put(stop.stop_id, stopIndex);
             stopIdForIndex.add(scopedStopId);
+            // intern zone IDs to save memory
```
abyrd: gtfs-lib might already intern every string field it reads - I guess this doesn't hurt but might not be necessary.

ansoncfit: When reading zone_id and parent_station, loadOneRow of stops.txt, gtfs-lib indeed uses a Deduplicator which "does the same thing as String.intern"

src/main/java/com/conveyal/r5/analyst/TravelTimeComputer.java
```diff
@@ -202,26 +208,46 @@ public OneOriginResult computeTravelTimes() throws IOException {
...
             PerTargetPropagater perTargetPropagater = new PerTargetPropagater(egressModeLinkedDestinations, request,
                     transitTravelTimesToStops, nonTransitTravelTimesToDestinations);
 
             // We cannot yet merge the functionality of the TravelTimeReducer into the PerTargetPropagator
             // because in the non-transit case we call the reducer directly (see above).
-            perTargetPropagater.travelTimeReducer = travelTimeReducer;
+            if (request.inRoutingFareCalculator == null) {
+                perTargetPropagater.travelTimeReducer = travelTimeReducer;
+            } else {
+                perTargetPropagater.travelTimeReducer = new TravelTimeReducer(request, transitTravelTimesToStops
```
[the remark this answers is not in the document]
ansoncfit: I think these tortuous conditionals were to handle a non-deterministic number of samples generated by McRaptorSuboptimalPathProfileRouter.java:route().  But as of https://github.com/conveyal/r5/pull/453/commits/95d076896e0a076207ced83742deb8fb620127df, the McRaptor router returns exactly the requested number of samples, so we should be able to clean this up.

src/main/java/com/conveyal/r5/analyst/TravelTimeReducer.java
```diff
@@ -40,31 +40,24 @@
...
+     *             run Monte Carlo draws within departure minutes, so it uses the monteCarloDraws parameter in a way
+     *             that's consistent with its name.
      */
-    public TravelTimeReducer (AnalysisTask task, int timesPerDestination) {
+    public TravelTimeReducer (AnalysisTask task) {
 
         this.maxTripDurationMinutes = task.maxTripDurationMinutes;
-        this.timesPerDestination = timesPerDestination;
+        this.timesPerDestination = task.inRoutingFareCalculator == null ? task.getMonteCarloDrawsPerMinute
+                () * task.getTimeWindowLengthMinutes() : task.monteCarloDraws;
```
abyrd: Nice, this is a lot easier to read.

Issue #426, opened 2018-05-07
Allow non-walk egress modes


ravi9884:
Actual
In multimodal transport r5 uses bicycle only for the first leg of the journey. The last leg always uses foot.

Expected
If there is a bike route at the final leg of the journey, it should be used instead of walking.

ansoncfit:
Our current implementation of LinkedPointSet means some effort is needed for non-walk egress modes. `makeStopToPointDistanceTables` uses `stopToVertexDistanceTables` (distances for WALK computed when the TransportNetwork is built and serialized with it).

Linking a pointset by bicycle would require recalculating the `stopToVertexDistanceTables` (which will be different from the ones serialized with the network if cyclists and pedestrians use different ways), extending these new distances to targets, and revising PerTargetPropagator to use the requested bikeSpeed instead of walkSpeed.  

Linking a pointset by car would require computing stop-to-vertex and vertex-to-target _times_, as total distance won't work (given varying speeds). To minimize other changes needed, we could use the convention of saving the times in the distance tables, then setting `PerTargetPropgator.speedMillimetersPerSecond` to 1 when `request.egressMode = CAR`.

ansoncfit:
See also https://github.com/conveyal/r5/issues/450

Issue #410, opened 2018-03-01
Check walk and bike speeds to avoid divide by 0 errors


ansoncfit:
If a user enters 0 for walk speed or bike speed, bad things happen (especially when running a regional analysis):

```
java.lang.ArithmeticException: / by zero
at com.conveyal.r5.profile.PerTargetPropagater.lambda$propagate$0(PerTargetPropagater.java:146)
at gnu.trove.map.hash.TIntIntHashMap.forEachEntry(TIntIntHashMap.java:422)
at com.conveyal.r5.profile.PerTargetPropagater.propagate(PerTargetPropagater.java:140)
at com.conveyal.r5.analyst.TravelTimeComputer.computeTravelTimes(TravelTimeComputer.java:313)
at com.conveyal.r5.analyst.cluster.AnalystWorker.handleOneRequest(AnalystWorker.java:382)
at com.conveyal.r5.analyst.cluster.AnalystWorker.lambda$null$0(AnalystWorker.java:283)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
at java.lang.Thread.run(Thread.java:748)
```

ansoncfit:
R5 times out when walk speed = 0 and access mode = bike.

Issue #285, opened 2017-06-23
use maxWalkTime to filter egress propagation


mattwigway:
When doing propagation from stops to the street network, we should be using maxWalkTime to filter the stop to point distance tables since we are using it for the initial stop search and for transfers. The maxWalkTime should apply to the length of any individual walking leg, not all of them together, which would be a resource-limiting problem.

PR #463, opened 2018-11-14
Access/egress changes

ansoncfit:
Our intent for `maxWalkTime` and `maxBikeTime` is that they are applied separately to access, transfer, and egress legs.  For example, if `maxWalkTime = 20 min`, itineraries can involve a 20 min walk from origin to first stop, a 20 min walk between transfer stops, and a 20 min walk from egress stop to final target.  This PR adds that intended behavior for access by walking (#460) and egress by all modes (#285). See also https://github.com/conveyal/analysis-ui/issues/396

It also includes small refactors and patches (e.g. #410)

In the UI, we should set reasonable limits on maxWalkTime and maxBikeTime.  We may also be close to enabling biking as an egress mode in the UI (#426).


src/main/java/com/conveyal/r5/profile/PerTargetPropagater.java
```diff
@@ -105,7 +110,9 @@ public PerTargetPropagater(LinkedPointSet targets, AnalysisTask task,
         // If we're making a static site we'll break travel times down into components and make paths.
         // This expects the pathsToStopsForIteration and pathWriter fields to be set separately by the caller.
         this.calculateComponents = task.makeStaticSite;
-        speedMillimetersPerSecond = (int) (request.walkSpeed * 1000);
+        StreetMode egressMode = LegMode.getDominantStreetMode(task.egressModes);
+        speedMillimetersPerSecond = (int) (request.getSpeedForMode(egressMode) * MM_PER_METER);
+        egressLegTimeLimitSeconds = (int) request.getMaxAccessTimeForMode(egressMode) * SECONDS_PER_MINUTE;
```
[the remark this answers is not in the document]
abyrd: If both multiplication operands are ints the result will be an int. If at least one of them is floating point the result will be floating point. request.getSpeedForMode returns a float, the result of the multiplication is a float, and you need an explicit cast to truncate it for storage in an integer variable. request.getMaxAccessTimeForMode returns an int though, so the result of the multiplication is an int and the cast is redundant. An IDE should show you that it's redundant. However also be aware that the operator precedence of casting is higher than multiplication or division, so here you're actually casting the output of getMaxAccessTimeForMode (which is already an int) to an int before multiplying it. If getMaxAccessTimeForMode was a method returning a floating point value, this would truncate the value before the multiplication, which would be bad.

src/main/java/com/conveyal/r5/analyst/TravelTimeComputer.java
```diff
@@ -152,10 +156,12 @@ public OneOriginResult computeTravelTimes() throws IOException {
...
                 Arrays.fill(nonTransitTravelTimesToDestinations, FastRaptorWorker.UNREACHED);
             } else if (accessMode == StreetMode.WALK) {
-                // Special handling for walk search, find distance in seconds and divide to match behavior at egress
-                // (in stop trees). For bike/car searches this is immaterial as the access searches are already asymmetric.
+                // Special handling for walk search: find distance in millimeters and divide by speed to match behavior
+                // at egress (in stop trees). For bike/car searches this is immaterial as the access searches are
+                // already asymmetric.
                 // TODO clarify - I think this is referring to the fact that the egress trees are pre-calculated for a standard speed and must be adjusted.
-                sr.distanceLimitMeters = 2000; // TODO hardwired same as gridcomputer, at least use a symbolic constant
+                sr.distanceLimitMeters = (int) (request.getSpeedForMode(accessMode) *
```
abyrd: Maybe we should hard-wire this to request.walkSpeed and request.maxWalkTime since it occurs in a block where accessMode always is StreetMode.WALK.

src/main/java/com/conveyal/r5/analyst/TravelTimeComputer.java
```diff
@@ -152,10 +156,12 @@ public OneOriginResult computeTravelTimes() throws IOException {
...
+                        request.getMaxAccessTimeForMode(accessMode) * 60);
```
abyrd: If we're using SECONDS_PER_MINUTE elsewhere, we could also use it here (importing the constant from FastRaptorWorker).

src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
```diff
@@ -628,7 +630,7 @@ public int getWorstCaseFrequencyDepartureTime (TripSchedule schedule, int stopPo
     private void doTransfers (RaptorState state) {
         // avoid integer casts in tight loop below
         int walkSpeedMillimetersPerSecond = (int) (request.walkSpeed * 1000);
-        int maxWalkMillimeters = (int) (request.walkSpeed * request.maxWalkTime * 60 * 1000);
+        int maxWalkMillimeters = walkSpeedMillimetersPerSecond * (request.maxWalkTime * SECONDS_PER_MINUTE);
```
abyrd: So it looks like this one is a pure refactor, we were already applying a distance limit in the transfers.

src/main/java/com/conveyal/r5/profile/PerTargetPropagater.java
```diff
@@ -214,23 +222,25 @@ private void propagateTransit (int targetIndex) {
...
-                    // Propagate from the current stop out to the target.
-                    int timeAtTarget = timeAtStop + distanceMillimeters / speedMillimetersPerSecond;
-                    if (timeAtTarget < cutoffSeconds &&
-                        timeAtTarget < perIterationTravelTimes[iteration]) {
-                        // To reach this target, alighting at this stop is faster than any previously checked stop.
-                        perIterationTravelTimes[iteration] = timeAtTarget;
-                        if (calculateComponents) {
-                            Path[] pathsToStops = pathsToStopsForIteration.get(iteration);
-                            perIterationPaths[iteration] = pathsToStops[stop];
+                if (distanceMillimeters / speedMillimetersPerSecond <= egressLegTimeLimitSeconds) {
```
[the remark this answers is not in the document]
abyrd: Thanks for pointing out the core change. So if I understand right we were not applying distance limits to walking access legs, but are now; we were already applying distance limits to transfers but it's been refactored for clarity; and were not applying distance limits to egress legs but are now. These distance limits can only shorten the egress distances, making them less than the maximum length used when building the distance tables. That might surprise people in some cases but I guess it's better than not applying any limits at all.

We should probably refactor to pull this division outside the foreach loop - division is a slow operation and I don't know if the JVM can figure out that these variables won't change on each loop iteration, especially since the loop body is a lambda function call. That would mean applying an egress distance limit rather than an egress time limit.

src/main/java/com/conveyal/r5/profile/ProfileRequest.java
```diff
@@ -135,10 +135,9 @@
     public int suboptimalMinutes = 5;
 
     /**
-     * The maximum duration of any transit trip found by this search. Results used to be very sensitive to this
-     * when we were using mean travel times, now we use medians so we could set this to 2 hours.
+     * The maximum duration of any transit trip found by this search.
      */
-    public int maxTripDurationMinutes = 4 * 60;
+    public int maxTripDurationMinutes = 2 * 60;
```
abyrd: We should verify that as expected the results are no longer super sensitive to this value.

ansoncfit:
I've addressed the comments above and added a few other changes. As revised, this PR:

**Allows egress by car/bike**

More specifically, this PR builds and uses proper linkage cost tables for CAR and BICYCLE StreetModes, fixing #426 (and allowing us to address https://github.com/conveyal/analysis-ui/issues/850)

Previously, we computed `stopToPointDistanceTables` and `pointToStopDistanceTables` for WALK by extending the `stopToVertexDistanceTables` saved with the network.  This PR renames the distance tables to linkage cost tables and adds `linkageCostUnit` as a field to `LinkedPointSet` to record whether the tables are storing distance or time values. As explained in a comment, 
> The speeds for Walk and Bicycle can be specified in an analysis request, so it makes sense above to store distances and apply the requested speed. In contrast, car speeds vary by link and cannot be set in analysis requests, so it makes sense to use seconds directly as the linkage cost.

Linking stops to points is a slow process. For this round of testing, I included hardwired limits of `BICYCLE_DISTANCE_LINKING_LIMIT_METERS = 2000` and `CAR_TIME_LINKING_LIMIT_SECONDS = 20 * 60`.  With a 20 minute limit for a test case in Boston, average time to link a stop by CAR was ~.18 s -- at that rate, linking 40,000 stops would take 2 hours, and that's probably for a linking limit that's too low. 

Let's discuss optimizations and reasonable limits to use when building these tables.

My sense is that CAR egress will be used infrequently, so it's not worth serializing stopToVertexLinkageCostTables for all networks at build time. We could consider saving the linkages separately, and providing more detailed progress updates when returning status messages to the UI.

**Limits access and egress legs**

As summarized above, we were not applying distance limits to walking access legs, but are now (fixing #460); we were already applying distance limits to transfers but it's been refactored for clarity; and were not applying distance limits to egress legs but are now (fixing #285).

With these limits in place, we should update documentation to clarify that the max walk and max bike times in the UI are per leg, and we should address recent comments at https://github.com/conveyal/analysis-ui/issues/396.

**Checks for non-positive speeds**
#410 

**Includes a few refactors and comment updates**
e.g. using symbolic constants like SECONDS_PER_MINUTE

ansoncfit:
@abyrd could you please re-review this when you have the chance?

It may be worth considering in conjunction with your investigations into #503

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
```diff
@@ -411,44 +430,87 @@ public void makeStopToPointDistanceTables (Geometry treeRebuildZone) {
...
+                // The speeds for Walk and Bicycle can be specified in an analysis request, so it makes sense above to
+                // store distances and apply the requested speed. In contrast, car speeds vary by link and cannot be
+                // set in analysis requests, so it makes sense to use seconds directly as the linkage cost.
+                linkageCostUnit = StreetRouter.State.RoutingVariable.DURATION_SECONDS;
+                // TODO confirm this works as expected when modifications can affect street layer.
+                StreetRouter sr = new StreetRouter(transitLayer.parentNetwork.streetLayer);
+                sr.streetMode = StreetMode.CAR;
+                sr.timeLimitSeconds = CAR_TIME_LINKING_LIMIT_SECONDS;
+                sr.quantityToMinimize = linkageCostUnit;
+                sr.setOrigin(stopPoint.getY() / FIXED_FACTOR, stopPoint.getX() / FIXED_FACTOR);
```
abyrd: Should this setOrigin call be the same as the one in the above `StreetMode.BICYCLE` clause? It seems to just use the index number of the street vertex corresponding to the stop.

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
```diff
@@ -411,44 +430,87 @@ public void makeStopToPointDistanceTables (Geometry treeRebuildZone) {
...
+                // Walking distances from stops to street vertices are saved in the transitLayer.
+                // Get the pre-computed distance table from the stop to the street vertices,
+                // then extend that table out from the street vertices to the points in this PointSet.
+                TIntIntMap distanceTableToVertices = transitLayer.stopToVertexDistanceTables.get(stopIndex); // walk!
+                Envelope distanceTableZone = stopPoint.getEnvelopeInternal();
+                GeometryUtils.expandEnvelopeFixed(distanceTableZone, TransitLayer.WALK_DISTANCE_LIMIT_METERS);
+                linkageCostToPoints = distanceTableToVertices == null ? null :
+                        extendDistanceTableToPoints(distanceTableToVertices, distanceTableZone);
+
+            } else if (streetMode == StreetMode.BICYCLE) {
```
abyrd: A lot of code can be factored out of these mode-specific clauses. It wouldn't make the code shorter, but would serve to communicate that the same basic process is happening in both cases, just with some variables changed. For example, `StreetRouter sr = new StreetRouter(transitLayer.parentNetwork.streetLayer); sr.streetMode = streetMode; sr.quantityToMinimize = linkageCostUnit;` is the same in both cases.

src/main/java/com/conveyal/r5/analyst/TravelTimeComputer.java
```diff
@@ -167,10 +171,13 @@ public OneOriginResult computeTravelTimes() {
...
-                // Special handling for walk search, find distance in seconds and divide to match behavior at egress
-                // (in stop trees). For bike/car searches this is immaterial as the access searches are already asymmetric.
-                // TODO clarify - I think this is referring to the fact that the egress trees are pre-calculated for a standard speed and must be adjusted.
-                sr.distanceLimitMeters = 2000; // TODO hardwired same as gridcomputer, at least use a symbolic constant
+                // Special handling for walk search: find distance in millimeters and divide by speed to match behavior
+                // at egress (in stop to point linkage cost tables). For bike/car searches this is immaterial as the
+                // access searches are already asymmetric (i.e. bike/car linkage distances precomputed for egress
+                // can't be used for access, because there are one-way streets; we ignore the possibility of one-way
+                // pedestrian ways).
+                sr.distanceLimitMeters = (int) (request.walkSpeed * request.maxWalkTime * // in StreetMode.WALK block
```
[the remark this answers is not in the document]
ansoncfit: This had been looking up the speed for the request mode, and I wanted to note that we changed it to walkSpeed. But the comment directly above makes this redundant.

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
ansoncfit: Yes, forgot to commit this line.

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
```diff
@@ -134,11 +133,11 @@ public LinkedPointSet (PointSet pointSet, StreetLayer streetLayer, StreetMode st
         // Null means relink and rebuild everything, but this will be constrained below if a base linkage was supplied.
         Geometry linkageCostRebuildZone = null;
 
+        // TODO general purpose method to check compatability
         if (baseLinkage != null && (
-                !(baseLinkage.pointSet == ((WebMercatorGridPointSet) pointSet).base ||
-                 baseLinkage.pointSet == pointSet) ||
+                baseLinkage.pointSet != pointSet) ||
```
abyrd: A test failed after this commit - maybe a stray paren on this line? Looks like the closing paren should be just before the opening curly brace of the conditional body.

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
```diff
@@ -142,19 +160,20 @@ public LinkedPointSet (PointSet pointSet, StreetLayer streetLayer, StreetMode st
...
         }
 
         // If dealing with a scenario, pad out the stop trees list from the base linkage to match the new stop count.
         // If dealing with a base network linkage, fill the stop trees list entirely with nulls.
-        while (stopToPointDistanceTables.size() < nStops) stopToPointDistanceTables.add(null);
+        while (stopToPointLinkageCostTables.size() < nStops) stopToPointLinkageCostTables.add(null);
 
         // First, link the points in this PointSet to specific street vertices.
-        // If there is no base linkage, link all streets.
+        // If there is no base linkage, link all points.
```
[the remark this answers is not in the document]
abyrd: Oh never mind, that's actually a boolean expression in there, which is true when base linkage is null.

abyrd:
Merged after discussion with @ansoncfit and a few small cleanup commits. Will probably need some changes to the constraint checking in the LinkedPointSet constructor when solving #503.

PR #467, opened 2018-11-21
Set route type from modifications

abyrd:
The mode of travel for added trips (routes) still can't be set in the UI, but this change at least copies over the default value (bus) from the AddTrips modification class. This has been confirmed to work - when bus is deselected in the UI, added routes are not considered. It should suffice to include a mode property in the modification JSON.

I don't want to merge this right away because it will change behavior in a way that may surprise users - if they are used to turning off buses and still seeing their added routes.


ansoncfit:
An upcoming project does plan to use this existing idiosyncracy.  They want to turn off buses in the baseline network, but include bus routes that are added via add-trip modifications.

Let's not deploy this to production until the UI has components to set the mode for add-trip modifications (unless the route_type is set in the worker, in which case the users in question could just stick with the current version of R5).

PR #468, opened 2018-11-24
EC2 startup and shutdown behavior

ansoncfit:
Together with https://github.com/conveyal/analysis-backend/pull/191, addresses https://github.com/conveyal/analysis-backend/issues/182 by making workers tag themselves in the userData startup script.

Also updates shutdown behavior (in light of https://aws.amazon.com/blogs/aws/new-per-second-billing-for-ec2-instances-and-ebs-volumes/).  We should go over the initial shutdown check carefully to ensure workers won't shut down prematurely:

- If workers start polling before they have completed network building, they may consider shutting down before actually completely building a network.  
- Also, is it a problem that the UI now stops retrying requests after 5 minutes?  For example, a worker may take 6 minutes to build a graph, but if a user walks away from their computer, the worker might eventually shut down without ever providing results that are displayed.

### DIFF
--- a/src/main/resources/worker.sh
+++ b/src/main/resources/worker.sh
@@ -6,6 +6,10 @@
 # 0: the URL to grab the worker JAR from
 # 1: the AWS log group to use
 # 2: the worker configuration to use
+# 3: the (Auth0) accessGroup (useful for billing)
+# 4: the (Auth0) user who made the request that started the worker
+# 5: the UUID of the TransportNetwork the worker will start analyzing
+# 6: the worker version to use
 # If you are reading this comment inside the EC2 user data field, this variable substitution has already happened.
 # The string instance_id in curly brackets is substituted by EC2 at startup, not by our Java code. It and any shell
 # variable references that contain brackets are single-quoted to tell MessageFormat not to substitute them.
@@ -61,6 +65,17 @@ cat /etc/aws
 
 echo AWS Log agent logs:
 cat /var/log/awslogs.log
+
+# Set up aws cli
+mkdir ~/.aws
+cat > ~/.aws/config << EOF
+[default]
+region = $REGION
+EOF
+
+# Tag the instance (so we can identify it in the EC2 console)
+aws ec2 create-tags --resources '{instance_id}' --tags Key=Name,Value=AnalysisWorker,Key=Project,Value=Analysis&&
+Key=group,Value={3},Key=user,Value={4},Key=networkId,Value={5},Key=workerVersion,Value={6}
 
 # Download the worker
 sudo -u ec2-user wget -O ~ec2-user/r5.jar {0} >> $LOGFILE 2>&1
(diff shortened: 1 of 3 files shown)


abyrd:
Discussing this now with @ansoncfit, we have made these refinements to the requirements.

There should be an initial long period where the worker will not shut down (over an hour) in case it needs to build a graph. This time will be shortened as soon as the worker actually handles a task.

Instead of considering shutting down at infrequent intervals of one hour or one half hour, we will eliminate the nextShutdownCheckTime variable and consider shutting down every time the worker polls the backend.

We'll get rid of the lastSinglePointTime and lastRegionalTaskTime and just keep a single shutdownAfter variable, which keeps getting bumped into the future every time a task is handled.

This should be done before the calculations start for a task, but after we're sure the network is already built.

ansoncfit:
I decided to keep separate shutdown behavior based on whether single point activity has occurred (which we take to indicate interactive use).  Most spot instances will never see single-point use, so they should shutdown quickly.

abyrd:
@ansoncfit it should be possible to have separate preload, single-point, and regional keepalive constants, but with only one "keep alive until" variable (rather than three). I think this would make the code more readable. Basically:

```
boolean inPreloading = true;
long keepAliveUntil = System.currentTimeMillis() + PRELOAD_KEEP_ALIVE_MINUTES * 60 * 1000;
...
adjustTimeout(SINGLE_KEEP_ALIVE_MINUTES);
...
adjustTimeout(REGIONAL_KEEP_ALIVE_MINUTES);
...
if (now > keepAliveUntil) shutdown;
...
void adjustTimeout (int keepAliveMinutes) {
    long t = System.currentTimeMillis() + keepAliveMinutes * 60 * 1000);
    if (inPreloading) {
        inPreloading = false;
        keepAliveUntil = t;
    } else {
        keepAliveUntil = Math.max(keepAliveUntil, t);
    }
}
```

src/main/resources/worker.sh
```diff
@@ -62,6 +66,17 @@ cat /etc/aws
 echo AWS Log agent logs:
 cat /var/log/awslogs.log
 
+# Set up aws cli
```
abyrd: This comment made me expect at first to see it installing the CLI software (though that's already available on Amazon Linux machines). It might be better to say something like "Create a config file to tell the AWS CLI which region to operate in."

ansoncfit:
I agree your proposed approach is much more readable @abyrd.  Is there any performance concern if `adjustTimeout` with a Math.max() check is called for every regional task?  I was trying to keep those checks to the polling.

src/main/resources/worker.sh
ansoncfit: 👍

abyrd:
There's certainly a cost but it should be so small as to be negligible. Even a regional task that's skipped because it's outside the street network already performs many times this much calculation to determine it should be skipped. And even those requests are effectively rate-limited by the workers' internal queue size.

ansoncfit:
Great, I'll go ahead and revise to use the cleaner approach.

abyrd:
Feel free to rename the variables and methods in my pseudocode - my names aren't very good.

PR #470, opened 2018-11-27
Linkage cache weighing

abyrd:
This builds on #469 but is more experimental. It provides a basic mechanism for estimating the size of linkages and distance tables in kB, and evicts from the cache based on estimated memory consumption instead of the number of linkages.


ansoncfit:
#463 will build `pointToStopLinkageCostTables` for CAR egress, potentially making linkages much heavier.  This may be worth revisiting at the same time.

abyrd:
In recent discussions we realized that cache weighing might not be effective because a lot of elements in linkages should be shared between several linkages (baseline, scenarios, sub-regions etc.) Currently they are not for non-walk modes, but making those elements shared would be a much greater optimization than kicking out voluminous un-shared linkages.

abyrd:
Closing for the reason cited above - the largest elements of a linkage may be shared among several linkages, so it's not straightforward to determine how much memory will be freed by removing one.

Issue #483, opened 2018-12-15
FileAlreadyExistsException when worker tries to save scenario .json from S3


ansoncfit:
```
15:30:38 2781 [pool-2-thread-2] INFO com.conveyal.r5.transit.TransportNetworkCache - Applying scenario to base network...
15:30:38 2782 [pool-2-thread-2] INFO com.conveyal.r5.transit.TransportNetworkCache - Retrieving scenario stored separately on S3 rather than in the ProfileRequest.
15:30:38 2811 [pool-2-thread-2] ERROR com.conveyal.r5.transit.TransportNetworkCache - Error retrieving scenario 5c15608c32b98e5d92c601f9-0-3147522539 from S3: java.nio.file.FileAlreadyExistsException: cache/graphs/5c15607432b98e5d92c601f6_5c15608c32b98e5d92c601f9-0-3147522539.json
15:30:38 2811 [pool-2-thread-2] INFO com.conveyal.r5.transit.TransportNetworkCache - Loading scenario from disk file.
15:30:38 2811 [pool-2-thread-2] ERROR com.conveyal.r5.transit.TransportNetworkCache - Could not read scenario 5c15608c32b98e5d92c601f9-0-3147522539 from disk: com.fasterxml.jackson.databind.JsonMappingException: No content to map due to end-of-input at [Source: cache/graphs/5c15607432b98e5d92c601f6_5c15608c32b98e5d92c601f9-0-3147522539.json; line: 1, column: 1]
15:30:38 2811 [pool-2-thread-2] WARN com.conveyal.r5.transit.TransportNetworkCache - No scenario provided or loaded. Replacing with empty scenario.
15:30:38 2812 [pool-2-thread-2] INFO com.conveyal.r5.analyst.scenario.Scenario - Applying scenario null
15:30:38 2812 [pool-2-thread-2] WARN com.conveyal.r5.analyst.scenario.Scenario - Scenario does not have feed checksums, not checking to ensure it is applicable to transport network.
15:30:38 2814 [pool-2-thread-2] INFO com.conveyal.r5.analyst.scenario.Scenario - Resolving modifications against TransportNetwork and sanity checking.
15:30:38 2814 [pool-2-thread-2] INFO com.conveyal.r5.analyst.scenario.Scenario - Applying modifications to TransportNetwork.
15:30:38 2815 [pool-2-thread-2] INFO com.conveyal.r5.transit.TransitLayer - Rebuilding transient indices.
[...]

Issue #480, opened 2018-12-10
Not all EC2 instances tag themselves


ansoncfit:
After https://github.com/conveyal/r5/pull/468, workers are supposed to tag themselves.  On a static site analysis, I used the EC2 console to "launch more like this" in two rounds.  In the first round, I requested 99 instances and all tagged themselves.  In the second round, a few minutes later, I requested 200 instances and only ~2/3 of them did.

abyrd:
That's strange. Maybe EC2 API rate limiting?

ansoncfit:
Good idea, I'll look into it and consider adding a small random delay.  The batches of 150 just started are exhibiting similar inconsistent behavior.

ansoncfit:
Starting many batches of 100 yesterday, it appeared that all instances successfully tagged themselves.  I didn't see limits on tagging following the instructions at https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-resource-limits.html, but rate limiting above ~100 tagging requests at a time does appear to be a likely suspect.

abyrd:
Sounds promising. I think adding some random jitter of say 0-20 seconds (as we do with the worker polling delay) would do the job.

ansoncfit:
Starting a batch of 200 with a random delay of up to 20 seconds did not seem to slow things down enough.  If we move the tagging operation after the R5 startup script so we don't delay users, we could try a longer delay (maybe ~4 minutes)?

ansoncfit:
Fixed in https://github.com/conveyal/r5/pull/488, then moved to analysis-backend in https://github.com/conveyal/analysis-backend/pull/203

PR #484, opened 2018-12-15
Regional patches

ansoncfit:
https://github.com/conveyal/analysis-backend/issues/200 was a specific case of a more general problem -- scenarios not being applied for regional analyses.  Fixes:
- [x] Request scenario-applied network from cache
- [x] Apply scenario when downloading from S3 (#483)

Separately,
- [x] Add jitter to auto-tagging (#480 )


abyrd:
I did a test of this PR running one baseline and one modified regional analysis, starting two workers. One of them was used for single point analysis and the other cloned in EC2 management. Results look correct - no banding, improvements at the expected locations. Worker logs show the expected sequence of actions when loading the scenario from S3. I think this is fixed.

PR #486, opened 2019-01-22
Move serializers to shared kryo-tools module

abyrd:
Both R5 and OTP defined custom serializers, and some of those defined in R5 were used by OTP. We don't want OTP to depend on R5 just for a couple of serializers. So I moved all the serializers into the same project as our object-differ that enables serialization tests. 

That kryo-tools project now transitively brings in versions of Trove, Guava, and Kryo that will be identical across both projects.


abyrd:
Note that based on what I've seen after this transition in OTP, this may yield incompatible network files. This should not be a problem in Analysis as long as it is used in a new version of R5.

PR #488, opened 2019-02-07
Retry EC2 instance tagging

ansoncfit:
...every ~15 seconds until tags are successfully applied.

An outstanding question is whether worker.sh belongs in R5 or analysis-backend. Given the Broker code now resides in analysis-backend, I think it should be moved as well (see https://github.com/conveyal/analysis-backend/blob/dev/src/main/java/com/conveyal/taui/analysis/broker/EC2RequestConfiguration.java#L73, and the `resources` of the R5 pom)


abyrd:
For future reference: note that this PR contains commits that change the script, but then the script is deleted in a subsequent commit because we're moving it from R5 to analysis-backend.

PR #491, opened 2019-03-21
feat(worker): log request

ansoncfit:


ansoncfit:
Needed for https://github.com/conveyal/analysis-backend/pull/216

src/main/java/com/conveyal/r5/analyst/cluster/AnalysisTask.java
```diff
@@ -81,6 +81,8 @@
      */
     public int nPathsPerTarget = 3;
 
+    public boolean logRequest = false;
```
abyrd: Please add Javadoc explaining what the parameter is for and how it is used in practice.

PR #496, opened 2019-03-31
Parallelize finding transfers

abyrd:
@t2gran pointed out that building a Norway network takes 11 minutes, of which 8 are computing transfers (over 70% of network build time). It is relatively easy to parallelize this process. This could also very greatly improve the network preparation times for large networks like the Netherlands or Switzerland.

The process is a little more complicated / error prone where Analysis scenarios are concerned, so we'll want to carefully review this new transfer finder before using it in Analysis, but for the initial network build case, I'm fairly certain the new version is correct. Routing results seem reasonable, and all tests pass (except an unrelated failure caused by today's summer time shift, which @t2gran is fixing as an unrelated issue).


t2gran:
I have tested this with the Norwegian GTFS file and OSM data and the transfer calculation speedup on my 4 core CPU is almost 4 times as fast - which is great. 

I also run my test suite with 29 travel samples on the OTP2 routing and I get back the EXACT same itineraries as I did before fore this change.

PR #497, opened 2019-04-01
Release: road congestion, custom modifications, OTP2 range raptor

abyrd:
This merge will produce a new release, which can then be imported by other projects depending on R5. 

This release will enable analysis-backend to store and convert custom modifications, so our production analysis system can pass experimental modification types through to workers without having a model class for them. It will also be usable as a worker which understands the new experimental road congestion modification. So watch for a pull request on analysis-backend as soon as this R5 release happens.

This release will also contain a new range raptor implementation which is very cleanly decoupled from the existing R5 code and lives in its own package. This should have no effect at all on Analysis. It will be used by the OTP 2.x branch for its transit routing.


src/main/java/com/conveyal/r5/otp2/rangeraptor/PerformTransitStrategy.java
```diff
@@ -0,0 +1,34 @@
+package com.conveyal.r5.otp2.rangeraptor;
+
+import com.conveyal.r5.otp2.api.transit.TripPatternInfo;
+import com.conveyal.r5.otp2.api.transit.TripScheduleInfo;
+import com.conveyal.r5.otp2.rangeraptor.transit.TripScheduleSearch;
+
+
+/**
+ * Perform the transit part of the routing on behave of the {@link RangeRaptorWorker}.
```
ansoncfit: Could we add a bit more detail on this class? I'm not sure what a transit strategy is (and it might be worth differentiating from passenger boarding strategies [e.g. https://espace.library.uq.edu.au/view/UQ:20ab185])

abyrd:
@ansoncfit as far as I know the R5 point to point routing was / is mainly or only used for Modeify, and I expect that uses a specific frozen past version of R5. I have never used R5 point-to-point for much myself, but that work really enriched the multi-modal street model and street routing in R5 (which we also use in Analysis). We could re-wire R5 point-to-point routing to this new OTP2 range-raptor implementation, just to serve as a demo, but eventually OTP2 should supersede that entirely. One main difference is that (at least for now) OTP2 will continue to use the OTP street model and street routing. OTP2 will also be significantly slower to start up for a while, because in this prototype stage it has both an OTP graph and R5 TransportNetwork. Since we are not maintaining existing R5 point-to-point, we should eventually rewire it to this new range raptor implementation or remove it when OTP2 is providing similar services and performance characteristics.

src/main/java/com/conveyal/r5/otp2/rangeraptor/PerformTransitStrategy.java
abyrd: I just updated the comment, and renamed the interface for clarity. We're planning a renaming session where we'll go through and change a lot of identifiers and class names for clarity. Hopefully in the same PR with Javadoc updates.

ansoncfit:
Thanks for the clarifications 👍

Issue #503, opened 2019-04-12
Auto-scaled workers repeatedly perform car linking


abyrd:
I'm running a regional analysis on staging of the Basel region, using a car-wait modification. It automatically scaled up to 100 workers. The original worker on which I performed single point requests is working fine, handling over 200 tasks per minute. All the other auto-scaled workers are very slow, mostly at about 5 tasks per minute. Inspection of their logs reveals that they are continuously re-performing the car linkage. The car linkage is apparently being sought on a different street layer instance for every task, and these linkages are being evicted.

Filtering one worker's log for "CAR" we see:
```
10:20:40
1609204 [pool-2-thread-3] INFO com.conveyal.r5.streets.LinkedPointSet - The base linkage is for street mode CAR
10:20:40
1609204 [pool-2-thread-4] INFO com.conveyal.r5.streets.LinkedPointSet - The base linkage is for street mode CAR
10:20:40
1609204 [pool-2-thread-2] INFO com.conveyal.r5.streets.LinkedPointSet - The base linkage is for street mode CAR
10:20:40
1609204 [pool-2-thread-1] WARN com.conveyal.r5.analyst.PointSet - Linkage cache evicted Tuple2[com.conveyal.r5.streets.StreetLayer@621c98d6, CAR], cause: SIZE
10:20:40
1609204 [pool-2-thread-1] INFO com.conveyal.r5.streets.LinkedPointSet - The base linkage is for street mode CAR
10:20:40
1609239 [pool-2-thread-2] WARN com.conveyal.r5.analyst.PointSet - Linkage cache evicted Tuple2[com.conveyal.r5.streets.StreetLayer@51b152fa, CAR], cause: SIZE
10:20:40
1609287 [pool-2-thread-3] WARN com.conveyal.r5.analyst.PointSet - Linkage cache evicted Tuple2[com.conveyal.r5.streets.StreetLayer@401cd0d6, CAR], cause: SIZE
10:20:40
1609320 [pool-2-thread-4] WARN com.conveyal.r5.analyst.PointSet - Linkage cache evicted Tuple2[com.conveyal.r5.streets.StreetLayer@53aa1427, CAR], cause: SIZE
10:20:40
1609322 [pool-2-thread-1] WARN com.conveyal.r5.analyst.PointSet - Linkage cache evicted Tuple2[com.conveyal.r5.streets.StreetLayer@27aff22a, CAR], cause: SIZE
10:20:41
[...]

PR #508, opened 2019-04-29
Various fixes to PointSet Linking

abyrd:
## Various fixes to PointSet Linking
This is for review only - do not merge this PR yet, as the solution to #503 may be related and may be included in this PR.

**Fix broken assumptions in LinkedPointSet constructor**

Recently we re-enabled some assertions in the LinkedPointSet constructor, checking that the supplied base base linkage is compatible with the linkage being constructed. With those assertions enabled, R5 throws an exception as soon as I do a single point search.

The initial failure is due to a check on the base street layer. That check just seems to have been expressed incorrectly. But another problem came to light: The part of the LinkedPointSet constructor that copies the linkages clearly assumes that the base point set and supplied point set are the same size (and says so in its comments). It appeared that this was not the case in practice, for example when setting custom geographic bounds: a pointset with a smaller extent is used. Yet the results displayed on the map (isochrones) look correct.

Because of how the base linkage is fetched, a pointset mismatch should be impossible here. The only way a mismatch could occur is if the linkage cache within a particular pointset contained linkages for a different pointset, which should never happen. I added assertions to ensure such a mismatch would be caught.

From there I traced the problem to another constructor of LinkedPointSet: `LinkedPointSet (LinkedPointSet sourceLinkage, WebMercatorGridPointSet subGrid)`. It was incorrectly setting the pointSet field of newly created sub-grid linkages. The linkages were in fact for the correct sub-grid, but the pointSet field was being set to the super-grid. 

**Fix Car Linking**

Several other problems were fixed:

- The field specifying the linkage units was not being copied from a base linkage. 
- The car router was not able to exit transit stop vertices, linking method was (temporarily?) changed
- Car linking code was changed to return an interleaved array of (pointId, cost) as done by other modes

### DIFF
--- a/src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
+++ b/src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
@@ -134,11 +134,19 @@ public class LinkedPointSet implements Serializable {
         Geometry linkageCostRebuildZone = null;
 
         // TODO general purpose method to check compatability
-        if (baseLinkage != null && (
-                baseLinkage.pointSet != pointSet ||
-                baseLinkage.streetLayer != streetLayer.baseStreetLayer ||
-                baseLinkage.streetMode != streetMode)) {
-            throw new UnsupportedOperationException("Requested characteristics do not match baseLinkage");
+        // The supplied baseLinkage must be for exactly the same pointSet as the linkage we're creating.
+        // This constructor expects them to have the same number of entries for the exact same geographic points.
+        // The supplied base linkage should be for the same mode, and for the base street network of this street network.
+        if (baseLinkage != null) {
+            if (baseLinkage.pointSet != pointSet) {
+                throw new AssertionError("baseLinkage must be for the same pointSet as the linkage being created.");
+            }
+            if (baseLinkage.streetMode != streetMode) {
+                throw new AssertionError("baseLinkage must be for the same mode as the linkage being created.");
+            }
+            if (baseLinkage.streetLayer != streetLayer.baseStreetLayer) {
+                throw new AssertionError("baseLinkage must be for the baseStreetLayer of the streetLayer of the linkage being created.");
+            }
         }
 
         if (baseLinkage == null) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
```diff
@@ -134,11 +133,19 @@ public LinkedPointSet (PointSet pointSet, StreetLayer streetLayer, StreetMode st
...
+        if (baseLinkage != null) {
+            if (baseLinkage.pointSet != pointSet) {
+                throw new AssertionError("baseLinkage must be for the same pointSet as the linkage being created.");
+            }
+            if (baseLinkage.streetMode != streetMode) {
+                throw new AssertionError("baseLinkage must be for the same mode as the linkage being created.");
+            }
+            if (baseLinkage.streetLayer != streetLayer.baseStreetLayer) {
+                throw new AssertionError("baseLinkage must be for the baseStreetLayer of the streetLayer of the linkage being created.");
+            }
```
ansoncfit: Much better for debugging 👍

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
```diff
@@ -491,31 +508,40 @@ public void makeStopToPointLinkageCostTables(Geometry treeRebuildZone) {
...
                 StreetRouter sr = new StreetRouter(transitLayer.parentNetwork.streetLayer);
                 sr.streetMode = streetMode;
-                sr.setOrigin(transitLayer.streetVertexForStop.get(stopIndex));
+                int vertexId = transitLayer.streetVertexForStop.get(stopIndex);
+                if (vertexId < 0) {
+                    LOG.warn("Stop unlinked, cannot build distance table: {}", stopIndex);
+                    return null;
+                }
+                // TODO setting the origin point of the router to the stop vertex does not work.
+                // This is probably because link edges do not allow car traversal. We could traverse them.
```
ansoncfit: Would link edges need to consider permissions of the street edges they connect to?  See comment at https://github.com/conveyal/r5/blob/v4.5.3/src/main/java/com/conveyal/r5/streets/Split.java#L101

Issue #509, opened 2019-05-03
Total distance after split is higher than before split


REASY:
Hello, team,

We use R5 in [Beam project](https://github.com/LBNL-UCB-STI/beam). It seems that split algorithm has some issue. We see that the total distance after split exceed original edge length. It seems that it is due to the double adjustment. First adjustment happens [Split.java#L185](https://github.com/conveyal/r5/blob/dev/src/main/java/com/conveyal/r5/streets/Split.java#L185) and then it happens again in [Split.java#L209-L210](https://github.com/conveyal/r5/blob/dev/src/main/java/com/conveyal/r5/streets/Split.java#L209-L210)

Here is the output from the run to build the network on https://github.com/conveyal/r5/commit/52e564f7f3aa6a34169958a0ce6bb40fbe331eea, [v4.5.3](https://github.com/conveyal/r5/releases/tag/v4.5.3) with [extra patch to log](https://pastebin.com/afXrzCtn) some information during the split. Network and other data for R5: https://drive.google.com/open?id=1OqxwDU8JGsxGoiL6LjrfdSEKeO8hfC4j. 
Run it in the following way
![image](https://user-images.githubusercontent.com/5107562/57149013-046a4800-6df5-11e9-960e-1ccb15ade572.png)
```
00:00  INFO:  OsmId[50620] Original edge length: 1700.891000, total distance before: 1700.891000, total distance after: 2588.831000, ratio: 1.522044
00:15  INFO:  OsmId[50620] Original edge length: 2041.239000, total distance before: 2041.239000, total distance after: 2927.271000, ratio: 1.434066
00:15  INFO:  OsmId[50620] Original edge length: 2473.944000, total distance before: 2473.944000, total distance after: 2760.300000, ratio: 1.115749
00:15  INFO:  OsmId[50620] Original edge length: 1775.445000, total distance before: 1775.445000, total distance after: 1786.047000, ratio: 1.005971
00:15  INFO:  OsmId[50620] Original edge length: 1070.569000, total distance before: 1070.569000, total distance after: 1104.891000, ratio: 1.032060
00:15  INFO:  OsmId[50620] Original edge length: 984.855000, total distance before: 984.855000, total distance after: 1054.135000, ratio: 1.070345
00:15  INFO:  OsmId[50620] Original edge length: 715.478000, total distance before: 715.478000, total distance after: 887.050000, ratio: 1.239801
[...]

PR #510, opened 2019-05-03
Splitting fixes/refactors

ansoncfit:
We've gone back and forth on whether the `distance0` and `distance1` values for splits (and in LinkedPointSet) should include the perpendicular (off-street) component as well as the parallel component (along the edge).  In #358, the perpendicular component was added to these values, but other parts of the code were not updated to reflect this change (leading to #509).

This PR keeps the distance along the edge separate from the distance to the edge.


abyrd:
I read through all changed files, the fix looks good to me. I applied some style changes while reviewing and included them in an additional commit.

PR #523, opened 2019-06-05
Mark destinations exceeding maxTripDurationMinutes UNREACHED

### LINKED ISSUE #522: TravelTimeReducer does not check maxTripDurationMinutes limit for single-point requests
The TravelTimeReducer checks whether total travel times (including pointset linkage costs)  exceed `maxTripDurationMinutes` (as set from the request coming from the UI by `populateTask() ` in `analysis-backend AnalysisRequest`) and correctly calculates accessibility results.

For single point requests, however, total travel time can be set in the UI after results are received, and this check is skipped.  It shouldn't be strictly necessary, as `populateTask()` doesn't reset the default (maximum) value of `maxTripDurationMinutes`. But without the check, overflow issues can lead to anomalies in the UI for some destinations more than 255 minutes away from origins.

See discussion at https://github.com/conveyal/analysis-ui/issues/665

ansoncfit:
Fixes #522

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/TravelTimeReducer.java
+++ b/src/main/java/com/conveyal/r5/analyst/TravelTimeReducer.java
@@ -22,7 +22,10 @@ public class TravelTimeReducer {
 
     private static final Logger LOG = LoggerFactory.getLogger(TravelTimeReducer.class);
 
-    /** The task used to create travel times being reduced herein. */
+    /** Maximum total travel time, above which a destination should be considered unreachable. Note the logic in
+     * analysis-backend AnalysisRequest, which sets this to the requested value for regional analyses, but keeps
+     * it at the default value from R5 ProfileRequest for single-point requests (which allow adjusting the cutoff
+     * after results have been calculated) */
     private int maxTripDurationMinutes;
 
     /** Travel time results for a whole grid of destinations. May be null if we're only recording accessibility. */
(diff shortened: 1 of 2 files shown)


src/main/java/com/conveyal/r5/analyst/TravelTimeReducer.java
```diff
@@ -164,6 +156,25 @@ private static int findPercentileIndex(int nElements, double percentile) {
         return percentileTravelTimesMinutes;
     }
 
+    /**
+     * Divide by 60 and return if the value is less than the maxTripDurationMinutes; otherwise return UNREACHED.
+     */
+    private int convertToMinutes (int timeSeconds) {
+        if (timeSeconds == FastRaptorWorker.UNREACHED) return FastRaptorWorker.UNREACHED;
+
+        // Int divide will floor; this is correct because value 0 has travel times of up to one minute, etc.
```
abyrd: It's great to have this logic pulled out into its own method. Could we change the Javadoc though to reflect the semantics of the operation without emphasizing the numeric constant 60? I think we should also hoist the inline comments about truncation into the method-level Javadoc. I was thinking about the integer division implications and only realized later they were explained right below.

For example: "Convert the given timeSeconds to minutes. If timeSeconds equals or exceeds the maxTripDurationMinutes, instead return a value indicating that the location is unreachable. The minutes to seconds conversion uses integer division, which truncates toward zero. This approach is correct for use in accessibility analysis, where we are always testing whether a travel time is less than a certain threshold value. For example, all travel times between 59 and 60 minutes will truncate to 59, and will correctly return true for the expression (t < 60 minutes). We are converting seconds to minutes before we export a binary format mainly to narrow the times so they fit into single bytes (though this also reduces entropy and makes compression more effective). Arguably this is coupling the backend too closely to the frontend (which makes use of UInt8 typed arrays); the front end could in principle receive a more general purpose format using wider or variable width integers."

Issue #512, opened 2019-05-04
Test conservation of edge length and geometry when splitting


abyrd:
In #509 we had problems with edges changing length when they were split. We should have some automated tests that cause long, curved edges to be recursively split and make sure the resulting distances add up to the original edge length, and that the split points are properly positioned. Good test data would involve long snaking roads with no intersections and many stops along the way.

Issue #511, opened 2019-05-04
Preserve geometry of split edges


abyrd:
Currently when an edge is split (e.g. to link in a transit stop) at least half of its detailed geometry is destroyed and it's replaced with a straight line. In cases where an edge is split only once, this is relatively harmless for routing because the length of the edges is updated during the split. However when more than one transit stop lies along the same edge, edges are split recursively. The split location may be very approximate or completely incorrect since a straight line is being used instead of the actual street shape. This problem is much worse in places where streets are very curved, there are long distances between intersections, and there are several stops on a segment between intersections.

PR #524, opened 2019-06-05
Preserve geometry after splits

ansoncfit:
- [x] #511
- [x] #512


src/main/java/com/conveyal/r5/streets/EdgeStore.java
```diff
@@ -728,6 +729,32 @@ public boolean canTurnFrom(StreetRouter.State s0, StreetRouter.State s1,
...
 
+        /**
+         * Cut the packed list of intermediate coordinates in two after specified segment
+         * @return List with first element int[] of coordinates before end of specified segment, and second element
+         * int[] of coordinates after end of specified segment.
+         */
+        public ArrayList<int[]> splitGeometryAfter(int segment) {
+            // Original packed coordinates of edge
+            int[] original = geometries.get(pairIndex);
+            ArrayList<int[]> geoms = new ArrayList<>();
```
abyrd: Returning a List works from a practical point of view, but it feels a bit like it can have an arbitrary number of elements and can be modified / extended. We have pair and tuple types defined (`P2` and `T2`), so maybe this could return a `P2<int[]>`.

src/main/java/com/conveyal/r5/streets/EdgeStore.java
```diff
@@ -728,6 +729,32 @@ public boolean canTurnFrom(StreetRouter.State s0, StreetRouter.State s1,
             return true;
         }
 
+        /**
+         * Cut the packed list of intermediate coordinates in two after specified segment
```
abyrd: I think the Javadoc should be revised slightly. Because the arrays store only the intermediate (non-end) points of the geometry, this operation can be tricky to think about and a little more detail could save some confusion in the future. Especially the expression "coordinates after end of specified segment" gave me pause, since I believe the second return value includes the end of the specified segment. I thought through the logic and this method seems correct to me. Here's my attempt at revised Javadoc: 

Cut the packed array of intermediate coordinates in two at the specified segment index. The intermediate coordinates array does not include the first and last point of the geometry (which are supplied by vertices) so the segment at index N ends at intermediate point N, which occupies array indexes N*2 and N*2+1 in the packed array. All the points before the specified segment (including the segment's beginning point) should be the first return array, and all the points after the specified segment (including the segment's ending point) should be in the second return array. The first or second return array can be of length zero if we are splitting at the first or last segment respectively.

abyrd: It's also worth noting in the Javadoc that this is a nondestructive split, the original geometry is left as-is and copies are returned by the method.

src/main/java/com/conveyal/r5/streets/StreetLayer.java
```diff
@@ -1170,16 +1170,16 @@ public int getOrCreateVertexNear(double lat, double lon, StreetMode streetMode)
...
             // The edge we are going to split is mutable.
             // We're either building a baseline graph, or modifying an edge created within the same scenario.
             // Modify the existing bidirectional edge pair to serve as the first segment leading up to the split point.
             // Its spatial index entry is still valid, since the edge's envelope will only shrink.
             edge.setLengthMm(split.distance0_mm);
             edge.setToVertex(newVertexIndex);
-            // Turn the edge into a straight line.
-            // FIXME split edges and new edges should have geometries!
-            edge.setGeometry(Collections.EMPTY_LIST);
+            geoms = edge.splitGeometryAfter(split.seg);
```
abyrd: It looks like we split the geometry and assign the split geometry to the new edges only when the original edge is mutable. The else clause below would result in straight-line geometries for splits in scenarios. I think once we've reached this section of code (`// The split is somewhere along a street away from an existing intersection vertex`) we always need to split any existing intermediate geometry, so rather than making a `new ArrayList<>()` above, we can just always perform the split with `ArrayList<int[]> geoms = edge.splitGeometryAfter(split.seg)`, then use the returned split geometries differently in these two different branches of the conditional.

src/main/java/com/conveyal/r5/streets/StreetLayer.java
```diff
@@ -1207,6 +1207,9 @@ public int getOrCreateVertexNear(double lat, double lon, StreetMode streetMode)
         EdgeStore.Edge newEdge1 = edgeStore.addStreetPair(newVertexIndex, oldToVertex, split.distance1_mm, edge.getOSMID());
         // Copy the flags and speeds for both directions, making newEdge1 like the existing edge.
         newEdge1.copyPairFlagsAndSpeeds(edge);
+
+        if (geoms.size() > 1) newEdge1.setGeometry(geoms.get(1));
```
abyrd: I think `geoms` should always be defined, and always have length 2 (hence my suggestion to use a P2 type above). If in some branch the geometry split would not happen, I would expect an uninitialized value of `geoms == null`.

src/main/java/com/conveyal/r5/streets/EdgeStore.java
```diff
@@ -728,6 +729,32 @@ public boolean canTurnFrom(StreetRouter.State s0, StreetRouter.State s1,
...
+            geoms.add(Arrays.copyOfRange(original, 0, segment * 2)); // Multiply by 2 because packed array
+            geoms.add(Arrays.copyOfRange(original, segment * 2, original.length + 1));
+            return geoms;
+        }
+
+        /**
+         * Set intermediate coordinates directly with packed array
+         * @param coordinates Packed lists of lat, lon, lat, lon... as fixed-point integers
+         */
+        public void setGeometry (int[] coordinates) {
```
abyrd: There will be cases where one or both of the split geometry arrays is empty. We can save a bit of memory by detecting this and replacing it with the shared `EMPTY_INT_ARRAY` in this method. This optimization can also be mentioned in the Javadoc if added.

src/test/java/com/conveyal/r5/streets/StreetLayerTest.java
```diff
@@ -124,6 +125,55 @@ public void testSplits() throws Exception {
...
+         for (Coordinate stop : stopCoordinates) {
+            streetLayer.createAndLinkVertex(stop.y, stop.x);
+         }
+
+         int accumulatedLength = 0;
+         int originalGeometryIndex = 1;
+
+         EdgeStore.Edge edge = streetLayer.edgeStore.getCursor();
+
+         while(edge.advance()) {
```
abyrd: The first time advance() is called, the edge cursor is pointing at edge 0, so the same edge that was split, and from which the OSM ID and length were read. So I believe this test would not detect several failure modes, e.g. if the geometry split process completely failed or was a no-op, or if the split was non-destructive, did not temporarily delete the original edge, and did not set geometries on any of the resulting split edges.

To detect a wider range of failure modes, we should probably detect that each call to streetLayer.createAndLinkVertex returns a non-negative vertex ID, that the edge store has the expected number of edges after the operation, that each of those edges begins and ends at the expected vertex index, and that the original intermediate coordinates are split across all three resulting edges (rather than being concentrated in any one edge, i.e. that each new edge contributes some points to the comparison with the original geometry).

I know you need to make use of these changes ASAP and don't want to delay the merge too much, so I'm going to merge this one and take a shot at refining the test myself in another pull request.

abyrd: I'm working on the test now. I see it makes an assumption about the order of the generated edges, although order is not guaranteed by the contract of the splitting method. I'm making it a bit more tolerant of implementation changes and the addition of more split points to the test.

abyrd: I have revised the test over in PR #525, which now contains the changes in this PR via a merge to dev.

Issue #514, opened 2019-05-04
Silent (from user perspective) scenario application failure


abyrd:
In a recent Analysis run there are stripe artifacts:
![image](https://user-images.githubusercontent.com/112871/57180401-d6e7d200-6eba-11e9-96f5-59f4092ae5dc.png)

This might be the same thing we saw in #483: errors happen when finding/applying scenarios. In the log in #483 we see: `TransportNetworkCache - No scenario provided or loaded. Replacing with empty scenario.` 

This means that when an error happens, instead of just failing to submit results for those origin points, and the backend redelivering and retrying them, it instead applies no scenario semi-silently and continues to calculate and return results. We need to change that behavior.

abyrd:
Although I've patched R5 to avoid returning bogus regional results with no scenario applied, it would still be interesting to identify what was causing the scenario application failures in the first place. It might be something to do with fetching scenarios from S3 or saving them locally on the workers. We will need to dig through the staging worker logs to see.

abyrd:
It turns out that the striping was not due to a limited number of workers failing; it was due to only one of 100 workers actually applying the scenario. All other workers were failing to apply the scenario and falling back on the baseline. This was not happening silently in the sense that the problem was logged on the worker, but from an end user perspective and from the backend's perspective it was invisible. 

The patch ensures that workers fail hard and refuse to submit work results instead of just failing to apply the scenario. It causes analyses to stall forever instead of appearing to complete normally, without applying any scenario. 

Generally we want things to fail fast and loudly whenever something is amiss. Also throughout R5 I think we should avoid falling back on / initializing with defaults - I recently also encountered a problem with a field that was being initialized to a default value and never overwritten.

trevorgerhardt:
We believe this issue is solved, the only known cause of the issue has been fixed. Re-open this issue if the striping has been seen again.

Issue #103, opened 2016-03-14
Proliferation of mode enums


abyrd:
We currently have profile/Mode, api/util/LegMode, api/util/TransitModes, api/util/NonTransitMode. Ideally there should only be 2: TransitMode and NonTransitMode. Also, all the class names should be singular.

Is there any reason for there to be more than 2?

buma:
profile/Mode should actually be named StreetMode and is mode used for routing specific part so it can only be simple (WALK, CYCLE, CAR)

LegMode is mode of ful streetSegment and can be simple or complex (Bicycle rent(consists of walking and cycling), P+R, which actually in implementation consists of multiple streetRouters with different streetMode)

TransitModes are for transit parts of the route and are the same as in GTFS.

NonTransitMode are for specifying which mode is used in specific street edges under streetSegment and are the same as profile/Mode since they can consist only of simple modes. This should be joined with profileMode.

I think that there is value in having separated LegMode and NonTransitMode/profileMode since LegMode can have complex mode which can't be used in routing. And LegMode supports complex modes which are forbidden/can't be used in profileMode

ansoncfit:
https://github.com/conveyal/r5/pull/497 adds AIR. Perhaps the distinction we should aim for is scheduled vs. non-scheduled modes?

abyrd:
Just a note: `com.conveyal.r5.api.util.LegMode#toStreetMode` is performing conversions between these types.

PR #525, opened 2019-06-12
Basic support for pick-up delay for autonomous or hailed vehicles

### LINKED ISSUE #426: Allow non-walk egress modes

abyrd:
We want to evaluate scenarios where small vehicles pick up passengers and take them to or from transit. Some users see these as "autonomous vehicle" scenarios; others are thinking of dynamically dispatched vans or cars, which may be driven by people.

A relatively simple, parametric way to model this is to add a spatially varying wait time to the beginning of car legs. This is implemented as a new Analysis scenario modification type. As we did with the road congestion modifications, a GeoJSON file of polygons is supplied and pulled by the workers from S3.

Together with the road congestion modifications, these are the first modifications to the road network (all previous modifications affected only transit, except in places where new transit stops split street edges). 

This implementation is useable but is not really complete. Some details of how modes compete on access and egress legs remain to be resolved. Different example implementations of these trade-offs are provided in the branch `ad-hoc-av-workers`. 

This PR should also resolve #426. Not only does it allow non-walk egress modes, it allows multiple different egress modes and will use the one with the shortest travel time from each stop to the destination. Different stops may use different modes because a car might travel faster but require a pick-up wait, so it's faster to walk in practice.

This PR also introduces a change to access and egress routing accounting for the fact that a stop might be linked to a walk-only way. After completing a non-walk access search, we must continue the search on foot to travel down any pedestrian walkways. A similar operation is needed when building non-walk egress tables. We should audit whether this problem has affected multimodal analysis results in the past.

Changes introduced here:
- Multiple access, egress, and direct modes can be supplied. Searches will be performed with all modes. On each path, whichever mode results in the shortest travel time will be used. 
- Allow specifying whether a worker should accept single point requests. This allows Analysis to start N workers on the same machine, without all of them trying to listen on the same port. That is useful for testing changes to regional analysis locally.
- Add walk segment when performing access or egress search by non-walk mode.
- Store and handle new perpendicular distances from Splits and base linkages in LinkedPointSets.
- Fail hard on missing scenarios, addressing #514
- Tolerate unrecognized fields when deserializing modifications, which allows custom modifications to work.
- Improved toString methods and logging.


src/main/java/com/conveyal/r5/analyst/scenario/Scenario.java
```diff
@@ -156,6 +156,16 @@ public TransportNetwork applyToTransportNetwork (TransportNetwork originalNetwor
         return copiedNetwork;
     }
 
+    /**
+    * @param lat Latitude of origin in floating point (not fixed int) degrees.
+    * @param lon Longitude of origin in floating point (not fixed int) degrees.
+    * @return time (in seconds) to wait at origin before beginning the trip
+    **/
+    public int computeOriginDelay (double lat, double lon) {
+        //TODO iterate through TNCPickupDelay modifications and return delay corresponding to lat/lon
```
ansoncfit: Is this method even used, or is it replaced by `getWaitTime()`?

src/main/java/com/conveyal/r5/profile/PerTargetPropagater.java
```diff
@@ -209,19 +220,52 @@ private void invertTravelTimes() {
...
+         * These two variables used to be computed once when the propagator was constructed, but now they must be
+         * computed separately for each egress StreetMode.
+         * At one point we believed float math was slowing down this loop, however it's debatable whether that was due
+         * to casts, the mathematical operations themselves, or the fact that the operations were being completed with
+         * doubles rather than floats.
+         */
+        int speedMillimetersPerSecond = (int) (request.getSpeedForMode(linkedTargets.streetMode) * MM_PER_METER);
+        int egressLegTimeLimitSeconds = request.getMaxTimeSeconds(linkedTargets.streetMode);
+
+        // FIXME: MASSIVE HACK THAT ONLY WORKS ON ONE STUDY IN SWITZERLAND
```
ansoncfit: Does this comment apply here, or to something that was relegated to an ad-hoc branch?

src/main/java/com/conveyal/r5/streets/StreetRouter.java
```diff
@@ -1199,5 +1200,19 @@ public boolean shouldBreakSearch() {
...
 
+    /**
+     * Continue a search by walking (presumably after a car or bicycle search is complete).
+     * This allows accessing transit stops that are linked to edges that are only walkable, but not drivable or bikeable.
+     * This maintains the total travel time limit and other parameters.
+     * NOTE: this conflicts with the rule that a router should not be reused.
+     * This is a good example of why we may want to change that rule. Alternatively this could construct a new StreetRouter.
+     * Just allowing more than one mode doesn't give the desired effect - we really want a sequence of separate modes.
+     */
+    public void keepRoutingOnFoot() {
```
ansoncfit: 🚶

ansoncfit:
As we just discussed, this PR should also resolve https://github.com/conveyal/r5/issues/426

src/main/java/com/conveyal/r5/analyst/TravelTimeComputer.java
```diff
@@ -13,16 +13,19 @@
...
+import com.conveyal.r5.transit.TransitLayer;
 import com.conveyal.r5.transit.TransportNetwork;
 import gnu.trove.iterator.TIntIntIterator;
 import gnu.trove.map.TIntIntMap;
 import gnu.trove.map.hash.TIntIntHashMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.Arrays;
+import java.util.*;
```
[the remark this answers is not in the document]
abyrd: No, this must be the stupid default setting on IntelliJ. I'll fix it.

src/main/java/com/conveyal/r5/profile/PerTargetPropagater.java
abyrd: This was initially a massive hack since it had a hard-coded wait time in it. Now, after a change, rather than being a hack I'd say it's implementing one of several possible approaches. Will update the comment.

src/main/java/com/conveyal/r5/api/util/LegMode.java
```diff
@@ -20,4 +22,29 @@ public static StreetMode getDominantStreetMode(Set<LegMode> modes) {
         else if (modes.contains(LegMode.BICYCLE)) return StreetMode.BICYCLE;
         else return StreetMode.WALK;
     }
+
+    /** Convert between these two enum types */
```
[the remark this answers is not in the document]
abyrd: I added a brief note, please go ahead and add any other comments you have in mind to #103.

src/main/java/com/conveyal/r5/analyst/TravelTimeComputer.java
```diff
@@ -46,68 +49,83 @@ public TravelTimeComputer(AnalysisTask request, TransportNetwork network, GridCa
...
         }
 
-        // The mode of travel that will be used to reach transit stations from the origin point.
-        StreetMode accessMode = LegMode.getDominantStreetMode(request.accessModes);
-        // The mode of travel that will be used to reach destinations from transit stations.
-        StreetMode egressMode = LegMode.getDominantStreetMode(request.egressModes);
-        // The mode of travel that would be used to reach the destination directly without using transit.
-        StreetMode directMode = LegMode.getDominantStreetMode(request.directModes);
+        // Convert from floating point meters per second (in request) to integer millimeters per second (internal).
+        int walkSpeedMillimetersPerSecond = (int) (request.walkSpeed * 1000);
```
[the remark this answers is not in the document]
abyrd: Thanks, I've imported that constant.

src/main/java/com/conveyal/r5/streets/PointSetTimes.java
```diff
@@ -22,11 +22,57 @@ protected PointSetTimes(PointSet pointSet, int[] travelTimes) {
...
         return travelTimes[p];
     }
+
+
+    /**
+     * Increment all reachable points by the given number of seconds.
+     */
+    public void incrementAllReachable(int seconds) {
+        for (int i = 0; i < travelTimes.length; i++) {
+            if (travelTimes[i] != Integer.MAX_VALUE) {
```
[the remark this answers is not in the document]
abyrd: Good idea, I've used that constant.

src/main/java/com/conveyal/r5/analyst/scenario/Scenario.java
abyrd: Deleted.

PR #528, opened 2019-06-17
Split geom and scenario application fixes

ansoncfit:
Need to cherry pick a few commits...


abyrd:
Would it make more sense to merge #524 (Preserve geometry after splits) into dev, then merge dev into #525 (av-wait-groundwork)? I'm a little wary of applying the same commits to separate branches. And I should do a final review of #524 anyway.

PR #530, opened 2019-06-26
Shapefile to OSM conversion with bike LTS, fixes for non-transit searches and CSV BOM

abyrd:
This PR adds a stand-alone tool for converting multiple shapefiles to routable OSM data. For the time being this tool is very specific to an analysis in Bogotá but it serves as an example of how to perform such a conversion, and could eventually become a standard import step. The usual problems with shapefile loading are encountered: shapes that fall short of crossing one another, and shapes that cross without sharing nodes. The tool attempts to handle these problems. It also copies LTS properties over to a tag like `lts=4` on the OSM data. The LTS labeler in R5 is updated to read such tags and store them in the internal routable network. They are marked as "explicit LTS" from the data source (as opposed to inferred LTS based on road type). For these explicit LTS edges, we skip a step that usually propagates high LTS from main streets to cross streets.

During this work I realized that non-transit routing was not working in Analysis. The origin point of the on-street routing was not being set in the code path for non-transit routing. This problem must have been introduced when working on handling of multiple access and direct modes, as part of the AV wait time modification.

I have finished up the transition begun with the AV wait time modification. Park+ride was reinstated as an access mode, and non-transit searches ("direct" on-street paths) are now handled by the same multi-modal code path that handles transit searches ("access" on-street paths). ~~This has the side effect that the bike and walk limits are always applied, even when transit is not in use. This is a change but I don't see it as a problem. It seems more coherent to me that the same limits are always applied.~~

Finally I have also included a fix for #532 to make Analysis tolerant of UTF-8 files starting with a BOM.


abyrd:
@ansoncfit I have made the change so that walk / bike time limits are only applied when these modes are used to access transit.

src/main/java/com/conveyal/r5/streets/EdgeStore.java
```diff
@@ -234,6 +234,9 @@ public EdgeStore (VertexStore vertexStore, StreetLayer layer, int initialSize) {
         // Comments below pasted from that document.
         // FIXME bicycle LTS should not really be flags, the categories are mutually exclusive and can be stored in 2 bits.
 
+        // If this flag is set, then its LTS has been loaded from OSM tags and we will not apply further processing to infer it.
+        BIKE_LTS_EXPLICIT (27),
```
ansoncfit: Worth a comment on why we skip to 27?

abyrd: This is just because the Bike LTS 1-4 were implemented as the highest 4 bit flags in the int. I just went one lower to group all LTS flags together. I suppose this would benefit from a comment but this is already released now.

abyrd: I added the comments and pushed directly to dev since this is just a pure documentation update.

PR #531, opened 2019-06-28
Pin node version for maven semantic release in travisci

trevorgerhardt:


abyrd:
Interestingly `bike-lts-shapefile` just built properly on Travis CI when I pushed to the branch. I will still merge this change for good measure though.

PR #534, opened 2019-07-01
Create Release

abyrd:
This brings a batch of changes into `master` from `dev` and should trigger the next Maven release. Unlike some releases this release is not just composed of refactors, and it will lead to different results for the same accessibility analysis when used as a worker. Therefore the commit message should be altered when merging to indicate that it affects results.


ansoncfit:
Priorities for testing:
- Access, egress, and transfer legs each use the requested maxWalkTime and maxBikeTime parameters. It's probably easiest to test this with a few simple add-trip modifications at locations/dates without baseline service.
- Requesting access/egress speed <= 0 throws an error
- Non-walk egress modes behave as expected (e.g. bike egress with different LTS limits)
- Repeated switching between access/egress modes does not trigger relinking (except when linkage tables are expected to be evicted from cache)
- When switching to smaller custom bounds, base pointsets are reused correctly
- Stations that are linked to walk-only edges can still be reached when car is used as an access mode
- Single-point and regional analyses with transit disabled work as expected for various travel time cutoffs

Associated changes for front-end/user manual:
- Set add-trip modification modes (https://github.com/conveyal/analysis-ui/issues/806). At least one user is relying on added-trips being trams (and therefore not excluded when bus is disabled as a transit mode in analyses), so we probably shouldn't deploy the updated backend until this feature is available in the UI.
- Clarify access/egress limits (https://github.com/conveyal/analysis-ui/issues/396).  The related R5 changes are grouped in https://github.com/conveyal/r5/pull/463
- Re-enable bike egress (https://github.com/conveyal/analysis-ui/issues/850)

These changes may affect results in the following ways:
- Newly created add-trip modifications will default to bus, instead of tram/streetcar.
- In single-point analyses with transit disabled, past versions of R5 and the front-end incorrectly considered some destinations more than 255 minutes away from an origin to be reachable even at low travel times. The fix in this version of R5 correctly marks such destinations as unreachable, which may decrease some single-point accessibility results relative to past (erroneously) high results.
- Street mode time limits may decrease accessibility relative to past versions of R5, especially for journeys requiring access, transfer, or egress segments exceeding 20 minutes.
- This version of R5 fixes a bug that increased the effective length of street segments when stops were linked to them. This fix may increase accessibility, especially for origins around street segments that are long, curving, or have many stops added along them. One way to to think of the bug: past versions of R5 ran people into bus stop poles (!), delaying them (and possibly triggering repeated concussions); this version of R5 has street users navigate around bus stop poles, reducing the erroneously long traversal times for these street segments.

abyrd:
Additional testing to perform:

- [ ] Custom JSON modification passed to worker correctly (log request on worker using special request property)
- [ ] Custom JSON modification for pick-up delay has expected effect
- [ ] Custom JSON modification for congestion has expected effect
- [ ] Test street splitting on very curved road with custom GTFS and OSM

PR #541, opened 2019-08-14
Only build cost tables on egress modes

### LINKED ISSUE #540: Build distance tables only for egress
We currently assume that when we perform a linkage for a particular StreetLayer and mode, we should also build distance tables for that mode at the same time. However, we only really need distance tables if the mode is being used for transit egress, not when it's an access or direct (non-transit) mode.

Building distance tables is often the slowest part of setting up a network. And new options make it more likely that we'll use a fast street mode for access or direct travel, without using it for egress from transit. We really shouldn't make the user wait for the worker(s) to build millions of huge tables they don't need.

Unfortunately the whole system does not distinguish between access and egress modes, and the preloader is keyed just on all modes, assuming that we need the same kind of linkage-with-distance-tables for every mode.

The real solution is probably to store distance tables outside the linkages, as their own set of objects, and even build them manually based on requests from the front end, saving them for later rapid loading.

In the short term, one possible solution is to only have the network preloader trigger distance table building for access and direct modes, skipping the egress modes. We'd then need a mechanism for lazy distance table building on already-existing pointset linkages, which would only be triggered lazily for car or bike egress. This would make the preloader essentially useless for these cases though, and these are cases that benefit from the preloader.

Alternatively, the preloader needs to have keys that consider egress modes separately from access and direct modes, plus the above on-demand lazy distance table building (as workers that can never be sure if the linkage they already have in memory might not have distance tables).

This is getting increasingly complex though, so we should continue working toward the "real solution" described above.

abyrd:
This PR fixes #540 (Build distance tables only for egress).

Recently we made it possible to use car or bicycle, rather than just walking, for egress from public transit. Egress from transit is "propagated" using precomputed cost tables, so we now make those cost tables for all modes used in a search, in case a mode is used for egress from transit.

However, building these cost tables is very time consuming, and not necessary for a mode that serves only for access to transit, or direct access to destinations, but not for egress. We needed to separate the concept of "linkage" from "cost tables", building linkages for all street modes in use (access, direct, and egress), but cost tables only for egress modes.

This PR makes those changes, defining a new class for EgressCostTables. The refactoring process has revealed some redundant code (for geographic cropping) and a somewhat confusing divergence in the code path when handling two different kinds of "base linkages". However the redundant code and multiple code paths have been maintained in the interest of performing one refactor at a time, avoiding excessive disruption to these critical chunks of code.

This PR also adds a mechanism that continuously reports progress of distance table construction to the NetworkPreloader, making it visible to the UI in status messages. Contrary to what I stated in a recent call, although there's a percentComplete field internally, this field is not currently exposed to the UI, so I have just included the percentage in the message text. Therefore it should work with no changes to the UI.

This PR also makes some changes that should significantly reduce memory consumption. Previously, we would end up with both stop-to-point and point-to-stop distance tables on many LinkedPointSets. The stop-to-point tables are source data and never actually used during analysis. In our current system, distance tables are either used as sources for other distance tables, or are used in routing, but never both. So I have made the table transposition operation irreversible - after the transposed tables are built (to facilitate routing) we release the original non-transposed data for garbage collection. 

There are a bunch of TODOs in this code. I may clean some of them out before merge, but I do intend to eventually do another round or two of refactoring in this area.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: java
 
-# R5 uses Java 8 features and Travis doesn't (yet) support OpenJDK 8
+# Oracle JDK is EOL, use OpenJDK 8
 jdk:
-- oraclejdk8
+- openjdk8
 
 # Maven semantic-release uses v10 now
 node_js:
--- a/src/main/java/com/conveyal/r5/analyst/NetworkPreloader.java
+++ b/src/main/java/com/conveyal/r5/analyst/NetworkPreloader.java
@@ -2,6 +2,7 @@ package com.conveyal.r5.analyst;
 
 import com.conveyal.r5.analyst.cluster.AnalysisTask;
 import com.conveyal.r5.analyst.cluster.RegionalTask;
+import com.conveyal.r5.analyst.progress.ProgressListener;
 import com.conveyal.r5.api.util.LegMode;
 import com.conveyal.r5.profile.StreetMode;
 import com.conveyal.r5.streets.LinkedPointSet;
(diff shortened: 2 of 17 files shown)


abyrd:
The CI build has passed, and I consider this ready for testing. So we can try it out on staging using 
custom worker v4.6.0-17-g011bfa6.

General test ideas:
Perform a walk-egress search and ensure that no additional time is needed to build the walk tables.
Use a mode (bicycle or car) as access only, then using it as egress, ensuring that is is quickly usable in the first case, but that cost tables are built in the second case (with progress reporting in the UI).

abyrd:
Thanks for adding those issue cross-references @ansoncfit. I will try to integrate those more directly into commit messages in the future. I'm currently working on the two problems you reported.

abyrd:
@ansoncfit with the last two commits, scenario vs. baseline egress table construction is visible in the progress messages, and CAR_PARK is now usable as an access mode.

However, searching with CAR_PARK,WALK as both access and direct modes, I see that the isochrones only represent post-transit travel and don't seem to be merged with the pre-transit (direct) travel time surface. This was not the behavior I expected, but perhaps I stranded the direct-with-post-transit surface merging functionality in one of those ad-hoc worker commits.

Please let me know if you have any thoughts on this, otherwise I'll keep investigating tomorrow.

abyrd:
About CAR_PARK handling: looking through TravelTimeComputer, I see that the CAR_PARK mode crudely overwrites any other access modes, and has a comment that specifically says `Disallow non-transit access.` So it is to be expected that in a combination like `CAR_PARK,BICYCLE` we see no direct bicycle access or car access, and only egress from transit stations that were reached by a combination of `CAR_PARK` and transit.

I did this when creating the multi-access-mode code, in order to preserve the existing CAR_PARK behavior since we hadn't decided on any new behavior. I may merge this as-is and then propose an improvement to handle CAR_PARK plus WALK or BICYCLE access modes.

PR #551, opened 2019-09-12
Bugfix release: do not limit number of analysis worker HTTP threads

abyrd:
We were setting the max threads for the analysis worker single-point HTTP listener to 10. This would fail on machines with a lot of cores (even if those cores were not available to the container the worker is running on) because the underlying HTTP server would insist it "needs" lots of request handler threads. Removing this limit lets analysis-backend tests succeed in containerized environments.

This was just not a workable approach for limiting the number of simultaneous single point requests, they should be fed asynchronously through a thread pool like the regional tasks.


abyrd:
@trevorgerhardt I had added you for review for proper form, but considering we already discussed this on the work session today, I'm just going to hit merge and trigger the release once the build passes.

PR #555, opened 2019-10-10
Santiago enhancements

ansoncfit:
For upcoming work in Santiago, we expect to need:
- more route info details passed to the `transitive.json` used by Taui
- ways to test the half-headway waiting time assumption


ansoncfit:
@abyrd I've updated from dev, so you may want to update your branch to facilitate this review

ansoncfit:
The two changes related to half-headway mode are:
- Activating it when `monteCarloDraws = 0` (which can be done in the custom profile request of the UI)
- Setting board time to half-headway after a passenger arrives at a stop, rather than half-headway after the first trip of a frequency entry arrives at a stop (https://github.com/conveyal/r5/pull/555/commits/523a7601ef2a2d9302d0f48dd7463f78b6a3a421#diff-1b10c1587ece1a5ee7a4ce929d4bbb82L742)

PR #556, opened 2019-10-13
Build under Java 11, update library and plugin versions.

abyrd:
We recently found a successful path to update OTP to build under Java 11: no module-info file, only adding an automatic-module-name to the JAR manifest via some Maven config, and allowing "illegal" reflective access with JVM switches only within the test runner.

I have applied these same changes to R5 in this pull request. As part of the migration to java 11, several dependencies and Maven plugins had to move to new versions. Of course this cascades other dependency version changes and creates possible conflicts.

Currently this PR does not compile because R5 dependencies (specifically gtfs-lib) are using the old JTS package name, but JTS has moved to the org.locationetch package hierarchy. So this PR is on hold until R5's dependencies to also migrate to Java 11 and bump all their dependency versions. Ideally we should cross-check that the whole dependency tree across projects (OTP, R5, gtfs-lib, osm-lib) is aligned on version numbers, including things like Protobuf, JTS, and Geotools, but also any others we spot.


abyrd:
The need to run R5 on Java 8 is starting to cause some friction. Today after loading a very large GTFS I got `java.lang.IllegalArgumentException: org.opengis.referencing.datum.DatumFactory is not an ImageIO SPI class` because I happened to be using a Java 11 runtime. Current R5 `dev` branch cannot run on Java 9+, primarily due to the Geotools version. We should try to get everything working on Java 11+ by the end of 2019.

abyrd:
I have the backend using a locked snapshot build of this right now and it seems to be working.

abyrd:
There's a problem with updating JTS in gtfs-lib and R5: we have large amounts of serialized TransportNetworks and GTFS feeds (in MapDBs) that expect all JTS objects to be in package `com.vividsolutions` instead of `org.locationtech`. This breaks all deserialization, we'd have to rebuild all the TransportNetworks and GTFS MapDBs from scratch, which could be done but is not the most straightforward thing.

abyrd:
I have confirmed that updating gtfs-lib to use new suffixes `.v2.db` and `.v2.db.p` may be straightforward. The only places I see these suffixes in the entire project and dependencies is gtfs-lib file BaseGTFSCache. A quick search and replace there, replacing `.db` with `.v2.db` causes all databases to be rebuilt and they should be uploaded to S3 under these new names, leaving the old ones in tact. We'd want to do a migration just to reduce lag the first time people use a feed, which should involve getting all recently used feed IDs from Mongo and hitting the GraphQL API to get something like all the route names from each feed, one by one in a loop.

abyrd:
We're back to using the same old GTFS MapDB files, thanks to an update to gtfs-lib. This seems to be working, ready for review.

PR #563, opened 2019-11-14
Revise definition of non-interpolated percentile

abyrd:
These changes were made during today's work session. We decided to exactly follow the standard definition of percentile to avoid any future confusion. Also added some comments based on our discussions about inclusive/exclusive comparisons when computing accessibility.


src/main/java/com/conveyal/r5/analyst/Grid.java
```diff
@@ -75,6 +75,8 @@
  * Class that represents a grid in the spherical Mercator "projection" at a given zoom level.
  * This is actually a sub-grid of the full-world web mercator grid, with a specified width and height and offset from
  * the edges of the world.
+ *
+ * Note that writing a grid out and reading it back in rounds the data values, which start out as fractional doubles.
```
ansoncfit: ...truncates the data values?

abyrd: It actually does round them, before truncating the rounded values (see L288 below).

PR #565, opened 2019-11-19
Use counting sort instead of general purpose sort in reducer

abyrd:
The counting sort approach is pretty interesting: if we pre-divide the travel times, scaling them down to minutes, there are only 120 values. Instead of rearranging thousands of array elements, you can just count how many times each value occurs (like making a histogram). Then you can read off the percentiles just by stepping through the minutes and tracking the cumulative count as they increase. This uses very small arrays that should fit in processor cache.

This is in fact advantageous. In profiling, it sped up the sorting and percentile selection by 75%. The total amount of time spent on these is smaller than on propagation, but can still be significant at upwards of 800 msec per origin.

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/TravelTimeReducer.java
+++ b/src/main/java/com/conveyal/r5/analyst/TravelTimeReducer.java
@@ -89,7 +89,12 @@ public class TravelTimeReducer {
         this.nPercentiles = task.percentiles.length;
 
         // We pre-compute the indexes at which we'll find each percentile in a sorted list of the given length.
-        // TODO check that indexes are increasing here.
+        // Check precondition: The percentiles in the request must be in ascending order.
+        for (int i = 1; i < task.percentiles.length; i++) {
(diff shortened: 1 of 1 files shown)


abyrd:
Marked work in progress because this requires the percentiles to be in ascending order, which is not checked yet. And the sort process should ideally have some tests.

abyrd:
Ascending-order precondition check is added, but there are still no tests for the percentile extraction. These will be a little difficult to add because the percentile extraction is not a static pure function, it relies on the precomputed percentile indexes and max travel time fields in the reducer.

abyrd:
When applying distance decay functions, we want to apply the function at one-second resolution to  smooth out as Monte Carlo randomization noise as possible. This means sorting the travel times as seconds rather than minutes before extracting the percentiles. The counting sort only works because there is a very small number of one-minute bins (120). When we switch to seconds there are way too many bins for this to be effective. Closing.

Issue #568, opened 2020-01-06
Sort exact-times timetables by departure time


ansoncfit:
A user has reported a bug related to add-trip modifications with multiple exact-times timetables.

 ## Expected behavior ##
For a simple add-trip modification with two exact-times timetables running every 30 minutes:
- Timetable 1, active only on Monday, from 7:01 to 8:30.
- Timetable 2, active all days of week, from 12:01 to 13:30.

Departing Monday between 7:00 and 8:00 ✔️:
![image](https://user-images.githubusercontent.com/2173529/71794670-8eb41d00-3010-11ea-95d0-5ac4670884a5.png)

Departing Monday between 12:00 and 13:00 ✔️:
![image](https://user-images.githubusercontent.com/2173529/71794692-b5725380-3010-11ea-8ec8-1ae685739465.png)

Departing Tuesday between 7:00 and 8:00 (service should be inactive) ✔️ :
![image](https://user-images.githubusercontent.com/2173529/71794739-e5b9f200-3010-11ea-8871-a77cb4528f1f.png)

Departing Tuesday between 12:00 and 13:00 ✔️:
![image](https://user-images.githubusercontent.com/2173529/71794780-0bdf9200-3011-11ea-8018-1cb2c8a1a601.png)

## Unexpected behavior ##
Similar to above, except timetables are replaced by:
- Timetable 3, active only on Monday from 12:01 to 13:30
- Timetable 4, active all days of week, from 7:01 to 8:30

Departing Monday between 7:00 and 8:00 (**service should be provided by Timetable 4** ❌ ):
![image](https://user-images.githubusercontent.com/2173529/71794987-020a5e80-3012-11ea-9056-5bfa26b1f3ce.png)

Departing Monday between 12:00 and 13:00 ✔️ :
![image](https://user-images.githubusercontent.com/2173529/71795022-1cdcd300-3012-11ea-9707-ccbb1aa48519.png)

Departing Tuesday between 7:00 and 8:00 ✔️ :
![image](https://user-images.githubusercontent.com/2173529/71795045-3b42ce80-3012-11ea-88a5-c7865ad942dc.png)

Departing Tuesday between 12:00 and 13:00 (service inactive ✔️ ):
![image](https://user-images.githubusercontent.com/2173529/71795142-bf955180-3012-11ea-9c5c-66702015d521.png)

## Troubleshooting ##
[...]

PR #569, opened 2020-01-06
Fix for exact-times add-trip modifications

ansoncfit:
We haven't yet determined a fix for #568, but it can go here when we do.


Nate-Wessel:
It looks like this is working. I tested a simple case with v4.7.1-127-ge1b96f7 and out of order exact times were working correctly as far as I could tell. 
I was able to reproduce the bug with the default v4.7.0.

PR #573, opened 2020-02-04
Github Workflows / Actions

abyrd:
This builds R5 via Github Workflows instead of Travis CI.


trevorgerhardt:
Have you considered using https://github.com/features/packages instead of our own S3 bucket?

abyrd:
> Have you considered using https://github.com/features/packages instead of our own S3 bucket?

That does look promising, will check it out.

PR #576, opened 2020-02-08
Multi cutoff

abyrd:
This adds the ability to compute multiple cutoffs at once. This has been tested on staging and when computing single cutoffs, results exactly match previous versions. When computing multiple cutoffs at once accessibility is slightly lower than in previous version for cutoffs below the maximum cutoff. That is currently considered an experimental feature though, and fixes will be applied as work continues.

I would just like to get this branch merged into R5 dev so we can include it in a release, which will then be a dependency of analysis-backend.

Note that this also adds some sanity checking code, reduces log clutter, and keeps the workers alive a little longer to avoid some problems with processing several jobs in rapid succession. It also includes the Github Actions changes that were already approved (that PR was on top of this one).


abyrd:
The build is failing because the gtfs-lib release is not yet visible on Maven Central. It should succeed in the near future.

PR #586, opened 2020-03-26
Refinements: travel time for origin and destination on same edge, adjustable web Mercator zoom level

### LINKED ISSUE #287: Origin and destination linked to same edge does not work
Consider a case where there is a long edge with no intersections. In analysis, placing the origin on this edge will create a sample that is linked to the ends of that edge. Reaching any destinations along that edge will require the router to walk to the end of the edge and back, causing an unrealistic drop in accessibility in the middle of the edge, as evidenced by the isochrone below:

image

ansoncfit:
Fixes #287

Consider an origin O and destination D linked to edge W-Z at split points X and Y:
```   
    O   D
    |   |    
W---X---Y-----Z
```
Previously, the travel time from O to D calculated by R5 would be too long, because it included required travel to one end of the edge (O-X-W-X-Y-D).  The change in this PR passes the origin split to LinkedPointSet::eval, allowing the calculation of the correct time (O-X-Y-D).

For analyses with cumulative-opportunity cutoffs shorter than the time to traverse an edge, and for analyses that use decay functions, this change will increase accessibility relative to past versions of R5.

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/LinkageCache.java
+++ b/src/main/java/com/conveyal/r5/analyst/LinkageCache.java
@@ -95,7 +95,7 @@ public class LinkageCache {
                             key.streetLayer,
                             key.streetMode
                     );
-                    if (basePointSetLinkage != null) {
+                    if (basePointSetLinkage != null && basePointSet.zoom == keyPointSet.zoom) {
                         LOG.info("Cutting linkage for {} out of existing linkage for {}.", keyPointSet, basePointSet);
                         return new LinkedPointSet(basePointSetLinkage, keyPointSet);
                     }
--- a/src/main/java/com/conveyal/r5/analyst/NetworkPreloader.java
+++ b/src/main/java/com/conveyal/r5/analyst/NetworkPreloader.java
@@ -167,19 +167,7 @@ public class NetworkPreloader extends AsyncLoader<NetworkPreloader.Key, Transpor
             // Egress modes must be tracked independently since we need to build EgressDistanceTables for those.
             this.allModes = LegMode.toStreetModeSet(task.directModes, task.accessModes, task.egressModes);
             this.egressModes = LegMode.toStreetModeSet(task.egressModes);
-
-            if (task.isHighPriority() || task.makeTauiSite) {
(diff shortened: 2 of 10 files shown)


ansoncfit:
Pedestrian routing using the previous R5 release made it to the ends of this segment of Mullholland Dr in 10 minutes, but not back to intermediate points
image

The first commit of this PR allows a pedestrian to access intermediate points without first going to the end vertices (fixing #287). But at zoom = 9, all of the destination cells along this segment are linked to other nearby streets, so the result is the same as above.

At zoom = 11, results are closer to expectations
image

There's still a "break" in the isochrones southeast of the origin along Mullholland, where cells are linked to other streets.  That's probably fine for accessibility calculations -- in this case, it would be correct to link destinations at these cells to those other streets.

For good measure, zoom = 12:
image


Two additional comments:
- Because WebMercatorExtents includes a zoom field, and the extents are part of the NetworkPreloader Key, enabling non-default zoom levels ended up being easier than expected. A network can be serialized with a base linkage at the default zoom, and linkages for other zoom levels will be created on the fly as requested.
- While "Johnny Depp" may be a relevant point of interest in the Hollywood Hills (see screenshots above), I adjusted our basemap poi filter to be a bit more selective (schools, hospitals, museums, sports complexes... excluding general "tourist attractions")

abyrd:
Discussing now on the phone: the tricky part is going to be the opportunity dataset changes, which could involve some straightforward but wordy coordinate conversions that will need heavy testing. Currently we do accessibility computations separately in the UI and backend, but are planning to do them entirely on the backend to allow for distance decay functions. Variable zoom level is another reason to do all these computations in one place, on the backend only.
However @ansoncfit points out that once we have resource management in place and opportunity source files are retained, we can just ask the user to re-generate the opportunity grids at the desired zoom level, and only allow combining data at the same zoom.

ansoncfit:
Note that for isochrone display, no corresponding front-end changes are needed because it already reads the zoom from the travel time surface's header (https://github.com/conveyal/analysis-ui/blob/0b98fa6d24cc344a2e1c4948a3fed7c1fcdb9249/lib/actions/analysis/parse-times-data.js#L30).

PR #597, opened 2020-04-14
Freeform pointset and build fixes

ansoncfit:
Miscellaneous fixes
* Re-enabling travel time calculations for freeform pointset origin-destination pairs
* Updating dependency locations
* Uploading R5 jars to our S3 bucket (for use as experimental workers); this happened automatically with previous CI tooling but was not yet working with GH actions

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -34,13 +34,13 @@ jobs:
           mvn clean verify --settings maven-settings.xml -B -V -Dgpg.skip
         fi
       env:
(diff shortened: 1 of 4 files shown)


abyrd:
Based on forum posts from other people encountering S3 upload errors, I regenerated the AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY many times, attempting to get a secret key that did not contain slashes or plus signs. All of the secret keys I generated contained these characters, so I abandoned this (hackish) approach, and moved on to creating a new user and set of permissions. I don't know exactly what is different about this new set of permissions or credentials, but it is now able to upload the R5 jars.

PR #634, opened 2020-10-15
Return information on paths between origin and destination

ansoncfit:
This branch writes the access mode and egress mode for path summaries output by workers for Taui sites.

~In conjunction with the path summaries branch in the ui,~ it also allows inspecting paths from the origin to the toLat toLon coordinates in the profile request.


ansoncfit:
Points we discussed at a high level:
- Adding paths to `OneOriginResult`
- Terminology on paths: basic path/path template (board stops, alight stops, pattern ids, implies walk time), itinerary (adds specific trips and associated in-vehicle time), iteration (adds specific departure minute and associated waiting time). "Avoid subclassing by composition."

ansoncfit:
This PR now includes three main features:
1. Include `accessMode` and `egressMode` in paths used for Taui sites
2. Includes `pathSummaries` in the `jsonBlock` sent from the worker via the backend to the UI. This can be tested on staging by running the appropriate `path-info` backend and worker `v6.0.0-54-gc6d5fe4`, and adding to the analysis request something like:
```
  "toLat": 42.366,
  "toLon": -71.187,
  "recordPaths": true
```
The response will then allow path results to be inspected in the browser console:
image

As an aside, note that the result above includes multiple iterations per departure minute, even for a network with no frequency-based lines (see #661)

3. Writes a CSV of path details for regional analyses with `originPointSeId` (freeform), `destinationPointSetId` (freeform), and `recordPaths` set. Workers return `RegionalWorkResult`s (now including `pathResult`) to the broker, where the multi-origin assembler writes a CSV that looks like:

origin | destination | path* | nIterations | avgWaitTime | inVehicleTime | avgTotalTime
-- | -- | -- | -- | -- | -- | --
170 | 65 | 5032 to 29001 on 95->70030 to 70020 on Orange->70077 to 70089 on   Red | 8 | 1199 | 1980 | 4560
170 | 65 | 2376 to 5015 on 94->70063 to 70089 on Red | 16 | 1159 | 2340 | 4560

* using GTFS route/stop ids; some commented out code would add route_short_name values.

src/main/java/com/conveyal/r5/profile/Path.java
```diff
@@ -24,13 +27,43 @@
...
+
+    // Additional characteristics (not keys for hashing)
+    public int inVehicleTime;
+    public int waitTime;
     public final int length;
 
+    /**
+     * Copy a path and add the specified egress mode
+     */
+    public Path(Path input, StreetMode egressMode) {
```
abyrd, line 52: This might be easier to maintain over the long term if it is a static method that performs a clone() and then sets egressMode.

src/main/java/com/conveyal/r5/profile/Path.java
```diff
@@ -149,10 +185,14 @@ public boolean equals(Object o) {
...
                 Arrays.equals(trips, path.trips) &&
                 Arrays.equals(boardStopPositions, path.boardStopPositions) &&
-                Arrays.equals(alightStopPositions, path.alightStopPositions);
+                Arrays.equals(alightStopPositions, path.alightStopPositions) &&
+                accessMode == path.accessMode &&
+                egressMode == path.egressMode;
     }
 
     @Override
+    // We tried replacing this custom implementation with Objects.hash(...), but it did not produce the expected
```
abyrd, line 194: We should investigate why it doesn't behave as expected. I'd ideally like to rely on stock functions to compute things like hash codes. Similar to copy constructors, it can become a maintenance liability to manually carry out a lot of actions on all fields in a class.

src/main/java/com/conveyal/r5/analyst/cluster/PathWriter.java
```diff
@@ -33,6 +36,9 @@
     /** This is a symbolic value used in the output file format an in our internal primitive int maps. */
     public static final int NO_PATH = -1;
 
+    /** Version of our path grid format **/
```
abyrd, line 39: Maybe we should introduce a changelog of our format versions inline here in the Javadoc.

src/main/java/com/conveyal/r5/profile/Path.java
```diff
@@ -24,13 +27,43 @@
...
+    /**
+     * Used only in propagation for writing paths for Taui outputs
+     * TODO update egress mode outside of path (e.g. in wrapper or array parallel to perIterationTravelTimes in
+     *  propagater)
+     */
+    public StreetMode egressMode;
+
+    // Additional characteristics (not keys for hashing)
+    public int inVehicleTime;
+    public int waitTime;
```
abyrd, line 46: The new class IterationDetails has these same fields. If I understand right every IterationDetails associated with a Path will have the same inVehicleTime and access/egress time as that path, just different initial wait times (and therefore different total times).

What if instead of a set of IterationDetails, each Path instead had a map from iterations (or departure times) to initial waits? To me this seems to more concisely factor out only the information that changes from one iteration's path to another. It could also eliminate the nonstandard inclusion of fields that are not hashed in Path, which may cause problems for code comprehension in the future.

It would also be possible to include this iteration-to-wait-time map directly as a field in Path, leaving it null when not needed. That's the composition approach I was thinking of, but maybe the Multimap is better if for a reason I'm not thinking of.

abyrd (commented):
Overall approach seems good. I added some comments on some refactors that could improve future maintenance and code comprehension.

src/main/java/com/conveyal/r5/profile/Path.java
ansoncfit: Our release version currently includes this custom hash function; let's address in a subsequent PR.

abyrd (approved):
Looking good. I made a couple of comments about parameter naming and branch structure, just for future legibility. I know I'm commenting on details here so I'm happy to do some of that minor clean-up myself.

ansoncfit:
I will merge now and plan to address outstanding comments on branch `path-info-cleanup`.

abyrd:
Note for future reference, this PR includes the refactor discussed in #662 that eliminates redundant output for non-frequency networks, including the removal of  superfluous loop and an assertion that dimension is 1, as discussed at https://github.com/conveyal/r5/pull/662#issuecomment-744419645

PR #635, opened 2020-10-20
Explain IntelliJ Gradle setup, add R5 name before expansion

abyrd:


ansoncfit (approved):
Thanks for the updates. After trying out the default "Gradle wrapper" approach over the last week, I concur with:

>We have encountered problems with this approach where IntelliJ seems to have insufficient control over the build/run/debug cycle. IntelliJ has its own internal implementation of the Gradle build process, and in our experience this works quite smoothly and is better integrated with the debug cycle.

Issue #638, opened 2020-10-22
Transfer walk distance not minimized


abyrd:
R5 was originally designed to report only travel times, so it doesn't take any special steps to break ties between paths with identical travel times.

When walking or transferring to a route, we might be able to board it at several stops. Because of the way we scan down patterns we will usually board at the earliest viable stop to ride a particular trip, even if that involves more walking than boarding at a later stop. In practice almost no one would do this because they'd maximize the ratio of wait time to walk time, to minimize the risk of missing a trip, or sometimes just to minimize the effort of walking.

Even though this doesn't affect reported travel times, now that we are reporting and displaying paths more often, it may confuse users or just look ugly. We want the paths to look like ones passengers would use in practice.

The problem is partially masked by the fact that we filter precomputed transfers to only those between the closest stops on any two patterns. But when used in routing, transfers are not between patterns, just between stops. So early boarding can still happen, especially on the access leg, which does not use precomputed transfers.

We expect this can be fixed by inserting a check around the section that checks whether we can board earlier trips at each stop:
https://github.com/conveyal/r5/blob/c7e1fa25439fae3b80923579e94f2a94570cf3f7/src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java#L558

We will probably want to check whether boarding at each stop can improve on cumulative walk time (or just walk time from the last transfer) and if so, reset the boarded-at stop even when the trip does not change.

PR #640, opened 2020-10-24
Refine paths

ansoncfit:
https://github.com/conveyal/r5/commit/4a3b115bfff49be13f49303ed0d3e571fd2f3c23 favors paths that reduce access and transfer time, addressing #638. Other commits on this PR address a few miscellaneous cleanup TODOs.

We should discuss the assertions.

Based on the way we iterate through trips, it appears we assume trips on a pattern do not overtake each other. We may want to add a note about that and other limitations in how we use GTFS (e.g. transfers, handling of exact_times) to the user manual.

### DIFF
--- a/src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
+++ b/src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
@@ -546,33 +546,73 @@ public class FastRaptorWorker {
                             }
                         }
                     } else {
-                        // We are already on a trip, but check if this stop was reached early enough to back up
-                        // to an earlier trip on the same pattern.
-                        // For example, consider two potential boarding stops on a pattern (A and B), with the search
-                        // origin closer to B:
-                        //  A----B
-                        //       |
-                        //       O
-                        // In this setup, the earliestBoardTime at A is less than (before) the earliestBoardTime at B.
-                        // It might be possible to board a certain trip at B, but only a later trip at A. Because A
-                        // is evaluated first (providing updates to downstream stops), we need to check for earlier
-                        // trips that could be boarded at B.
-                        int bestTripIdx = onTrip;
-                        while (--bestTripIdx >= 0) {
-                            TripSchedule trip = pattern.tripSchedules.get(bestTripIdx);
-                            if (trip.headwaySeconds != null || !servicesActive.get(trip.serviceCode)) {
+
+                        // We have already boarded a trip on this pattern at an upstream stop; check for cases in which
+                        // it would be preferable to board the pattern at this stop instead.
+
+                        TripSchedule onTripSchedule = pattern.tripSchedules.get(onTrip);
+
+                        // First, check whether boarding the same trip at this stop instead of the upstream one would
+                        // be preferred. This will not affect total travel time, but it will affect the breakdown of
+                        // walk vs. wait. So if needed as an optimization, these blocks could be disabled unless
+                        // request.computePaths = true;
+                        boolean boardHereInstead = false;
+                        if (inputState.previous == null) {
+                            // In the first transit round, arriving at this stop before arriving at the upstream stop
+                            // implies shorter access time. And arriving earlier or at the same time implies more
+                            // buffer time to board the trip, so boarding at this stop is preferable to boarding at
+                            // the upstream stop when the non-strict inequality is satisfied.
+                            if (inputState.bestTimes[stop] <= inputState.bestTimes[boardStop]) {
+                                boardHereInstead = true;
+                            }
+                        } else {
+                            // In subsequent transit rounds, we want to compare the length of the transfer legs.
+                            // Boarding a trip at this stop is preferable if it requires a shorter transfer than
+                            // boarding the same trip at the upstream stop.
+                            int upstreamTransferFromStop = inputState.transferStop[boardStop];
+                            if (upstreamTransferFromStop != -1) {
+                                int upstreamTransferTime =
+                                        inputState.bestTimes[boardStop] - inputState.previous.bestTimes[upstreamTransferFromStop];
+                                int transferFromStop = inputState.transferStop[stop];
+                                int transferTime =
+                                        inputState.bestTimes[stop] - inputState.previous.bestTimes[transferFromStop];
+                                if (transferTime < upstreamTransferTime) {
+                                    boardHereInstead = true;
+                                }
+                            }
+                        }
+
+                        if (boardHereInstead) {
+                            boardTime = onTripSchedule.departures[stopPositionInPattern];
+                            waitTime = onTripSchedule.departures[stopPositionInPattern] - inputState.bestTimes[stop];
+                            boardStop = stop;
+                        }
+
+                        // Second, it might even be possible to board an earlier trip at the downstream stop. This
+                        // must be checked to enssure correctness of total travel time results.
+                        // TODO handle trips on the same pattern that overtake each other?
+                        int earlierTripIdx = onTrip;
+                        while (--earlierTripIdx >= 0) {
+                            TripSchedule earlierTripSchedule = pattern.tripSchedules.get(earlierTripIdx);
+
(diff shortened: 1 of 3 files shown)


src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
```diff
@@ -546,33 +546,73 @@ private void doScheduledSearchForRound (RaptorState outputState) {
...
+                        // We have already boarded a trip on this pattern at an upstream stop; check for cases in which
+                        // it would be preferable to board the pattern at this stop instead.
+
+                        TripSchedule onTripSchedule = pattern.tripSchedules.get(onTrip);
+
+                        // First, check whether boarding the same trip at this stop instead of the upstream one would
+                        // be preferred. This will not affect total travel time, but it will affect the breakdown of
+                        // walk vs. wait. So if needed as an optimization, these blocks could be disabled unless
+                        // request.computePaths = true;
+                        boolean boardHereInstead = false;
```
abyrd, line 559: We should factor this whole decision whether to board here out into a method, and make that method call and the subsequent `if (boardHereInstead)` clause all conditional on request.computePaths.

src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
```diff
@@ -546,33 +546,73 @@ private void doScheduledSearchForRound (RaptorState outputState) {
...
+                                boardHereInstead = true;
+                            }
+                        } else {
+                            // In subsequent transit rounds, we want to compare the length of the transfer legs.
+                            // Boarding a trip at this stop is preferable if it requires a shorter transfer than
+                            // boarding the same trip at the upstream stop.
+                            int upstreamTransferFromStop = inputState.transferStop[boardStop];
+                            if (upstreamTransferFromStop != -1) {
+                                int upstreamTransferTime =
+                                        inputState.bestTimes[boardStop] - inputState.previous.bestTimes[upstreamTransferFromStop];
```
abyrd, line 575: If I'm not mistaken, we don't need to look at the previous round at all. We can look at the bestNonTransferTime for the upstreamTransferFromStop in the input state. The transferStop tells us which stop we transferred from within that same round, if any. The difference between those two arrival times should be proportionate to the the transfer distance.

If there is no need to look at the previous round, then I'm hopeful that these two main clauses (`if (inputState.previous == null) ... else`)  can somehow be collapsed into one. It seems to me like there should be a single expression that works to break ties in all rounds.

src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
```diff
@@ -546,33 +546,73 @@ private void doScheduledSearchForRound (RaptorState outputState) {
...
+                                int transferFromStop = inputState.transferStop[stop];
+                                int transferTime =
+                                        inputState.bestTimes[stop] - inputState.previous.bestTimes[transferFromStop];
+                                if (transferTime < upstreamTransferTime) {
```
abyrd, line 579: We're breaking the tie based on the one previous transfer time, but it seems roughly equivalent to me to maximize wait time over the whole path (as long as boarding time is the same). We are in fact already tracking cumulative waiting time in the array nonTransferWaitTime. Is there some way we could use that instead?

src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
```diff
@@ -546,33 +546,73 @@ private void doScheduledSearchForRound (RaptorState outputState) {
...
+                            boardTime = onTripSchedule.departures[stopPositionInPattern];
+                            waitTime = onTripSchedule.departures[stopPositionInPattern] - inputState.bestTimes[stop];
+                            boardStop = stop;
+                        }
+
+                        // Second, it might even be possible to board an earlier trip at the downstream stop. This
+                        // must be checked to enssure correctness of total travel time results.
+                        // TODO handle trips on the same pattern that overtake each other?
+                        int earlierTripIdx = onTrip;
+                        while (--earlierTripIdx >= 0) {
```
abyrd, line 595: It strikes me looking at this that once we're on a trip, we look at at least one earlier trip on every single downstream stop. This checking of earlier trips is in fact only necessary if the previous-round arrival time at that stop is earlier than (less than, not equal to) the departure time on the trip we're currently riding. I guess checking that additional condition is an optimization and may not actually save much computation over stepping back one trip at every stop and checking whether that previous trip is boardable.

Additionally, if we board an earlier trip, then all the decision logic about whether to boardHereInstead was in vain. So ideally that boardHereInstead logic would only run once it's clear we won't board an earlier trip. But again this is an optimization and may have negligible impact. I think it might just better reflect the structure of the decision tree.

abyrd (commented):
This seems like the right idea - when remaining on board a trip, break ties using some characteristics of the transfers. It does seem to me that there is probably a more compact / clear way to express this breaking of ties, maybe a way that has less special cases. I added some comments that might help identify a different way. I'm not totally sure of the observations in some of those comments, but wanted to share impressions immediately. I will revisit later and see if I have any clarifications to make.

src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
abyrd: The tie breaking rule you’re using in the latest updates in 782936e would do the job. I was just wondering if there was a single expression with less conditionals that would achieve something similar. One alternative might be the one mentioned in another comment below: maximizing wait time instead of minimizing walk time (which should be equivalent when boarding the same trip at different stops). I'll provide some more detail on that comment.

src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
abyrd: To clarify: I'm not sure maximizing wait time is necessarily better than minimizing walk time. It seems like maximizing wait time might correspond better with an underlying reason for choosing the shorter path (all other things being equal it makes you less likely to miss your bus). But most importantly, we are already tracking nonTransferWaitTime cumulatively, and I was thinking it might result in exactly the same expression for the first round as for subsequent rounds. If it doesn't then there's no real advantage. Anyway since you factored that tie-breaker predicate out into a method, the presence of a conditional is less problematic because it no longer affects the readability of the logic one level up.

src/main/java/com/conveyal/r5/profile/RaptorState.java
```diff
@@ -337,4 +340,28 @@ public boolean stopWasUpdated (int stop, boolean withinMinute) {
         }
     }
 
+    /**
+     * Checks whether boarding a at one stop allows for shorter access/transfer legs than boarding at another stop.
```
abyrd, line 344: The Javadoc should specify (maybe with `@param`) that here and there are stop indexes, and that a return value of `true` means here is better than there.

As this is an instance method in the context of a single round's state, we should also make it clear that we're talking about boarding in the _next_ round, not boarding in this round. But the meaning of the method might be clearer if we just don't mention boarding at all, only what the method computes. It simply tells whether the walk or access leg to reach one stop is shorter than the other within this round.

src/main/java/com/conveyal/r5/profile/RaptorState.java
```diff
@@ -337,4 +340,28 @@ public boolean stopWasUpdated (int stop, boolean withinMinute) {
...
+            return this.bestTimes[here] < this.bestTimes[there];
+        } else {
+            // In subsequent transit rounds, we want to compare the length of the transfer legs.
+            // We cannot directly compare arrival times or waiting times at the stops, because the transfer legs may
+            // have started at different times.
+            return transferTime(here) < transferTime(there);
+        }
+    }
+
+    private int transferTime(int stop) {
```
abyrd, line 362: Continuing the Javadoc expansion from shorterAccessOrTransferLeg, this method should state that it returns the length of time spent walking (or using some other street mode) within this round, to achieve the optimal route to the given stop in this round.

abyrd (approved):
Looks good, just suggested a few changes to Javadoc.

Issue #641, opened 2020-10-28
transfers.txt should override BOARD_SLACK_SECONDS and OSM walk distance


abyrd:
At least for the standard stop-to-stop entries in transfers.txt, and especially for the transfer type 2 ("minimum amount of time between arrival and departure to ensure a connection") we should be overriding the default OSM walk distance calculations. Although the GTFS spec text says "minimum amount of time", I believe the proper interpretation is to completely replace any transfers found by walking through the OSM network: transfers.txt is providing a typical safe amount of time needed to walk between the two stops.

This gets a little tricky because we currently apply a hard lower limit of 60 seconds between alighting and boarding, which would have to be configurable and interact with the transfer entries.

Related to #330. Some of these behaviors may need to be configurable at the routing or network building stage, as not all users will want stations and transfers to be interpreted in the same way. Many of these characteristics must be made optional, but we'll need a mechanism to configure that (#644).

abyrd:
Strangely BOARD_SLACK_SECONDS appears to only be used in classes for displaying paths, not for routing. This should be described in more detail in the docs.

abyrd:
Discussed again today - we decided that if transfers.txt is present, especially with transfers of type 2, then overriding any OSM transfers is more correct and should be the default behavior - in fact this probably doesn't even need a switch to disable it unless someone comes up with a concrete need to disable it.

ansoncfit:
@abyrd and I just discussed enabling three choices via `TransportNetworkConfig`:
- `OSM_ONLY` (ignore transfers.txt)
- `GTFS_EXHAUSTIVE` (skip OSM-based transfers)
- `GTFS_REPLACES_OSM` (use time from transfers.txt if specified for stop pair, otherwise use OSM-based time)

As noted above, the last option is probably preferred as a default. But there may be edge cases (e.g., in a complex with subway platforms separated by long times specified in transfers.txt, but no times specified for transfers between those platforms and nearby bus stops). In these cases, path results may be strange (but time results would be no worse than the current `OSM_ONLY` option)

PR #642, opened 2020-10-30
Use times from transfers.txt #641

abyrd:


abyrd:
One thing that remains here to make this production-ready is to ensure we handle the interactions with scenarios being applied.

ansoncfit:
Let's make sure to test with the MBTA's feed, which includes transfers and pathways.

RicoFio:
Hi @abyrd ! Is there a current roadmap on merging this into production by any chance? I'm asking since I noted a recent pick-up in interest [here](https://github.com/conveyal/r5/issues/971) and [here](https://github.com/r5py/r5py/issues/502).

I'm working with @vvaye (OP of the R5PY issue) and the circumvention we took was to remove the `transfers.txt` all together for now from the MBTA GTFS feeds. However, I don't know the impact on the R5 routing when doing so. If you could share a few words on the implications, that would be super useful.

Thanks for all your work!

abyrd:
Hi @RicoFio, sorry for the delayed response. We do not currently have a detailed plan for transfers.txt handling, but it is a good candidate for the upcoming roadmap.

For a long time most of our users were not relying on transfers.txt so it was simply skipped. But obviously there are an increasing number of cases where GTFS feeds rely on this file for correct interpretation.

The basic implementation is already here, but the problem is ensuring we've correctly identified the most common use cases and providing the right set of configuration options to give the intended results. Notably, we need to clarify whether transfers.txt replaces, complements, or restricts the auto-generated on-street transfers, and whether it does so per origin stop, per origin-destination stop pair, or fully replaces automatic transfer generation.

We also need to decide whether to handle the nuances of transfer specificity (e.g. trip to trip transfer rules) which may have some unwieldy interactions with the routing algorithm. And as noted above, scenarios can create and remove stops, leading to re-calculation of which transfers are available at all stops in the vicinity. We need to fully specify how this would interact with the various interpretations of transfers.txt mentioned above.

If you have any input on which approach(es) would be most useful please let us know.

PR #646, opened 2020-11-04
Remove backend auto shutdown

abyrd:
This PR removes the auto-shutdown activity tracker. Instead, it fires events after each API request. Components that produce dashboards or handle automated shutdown (for some particular hosting environment) can then subscribe to these messages. The response time of each API request is tracked at millisecond resolution, enabling performance monitoring.

This PR also adds projectId and variantIndex fields to SinglePointEvents. Components that subscribe to these events can then more easily summarize activity without taking the fragile step of extracting the projectId as a substring of the scenarioId.

These changes remove the **backend** auto-shutdown, not the **worker** auto-shutdown. The latter is separately implemented using `com.conveyal.r5.analyst.cluster.AnalysisWorker#adjustShutdownClock`. We may eventually modularize the worker in the same way as the backend, at which time we can extend this refactoring to the worker.


src/main/java/com/conveyal/analysis/components/HttpApi.java
```diff
@@ -79,14 +96,27 @@ public HttpApi (
...
-                UserPermissions user = authentication.authenticate(req);
-                LOG.info("{} {} by user {} in groups {}", method, pathInfo, user.email, user.groups);
+                // Determine which user is sending the request, and which permissions that user has.
+                // This method throws an exception if the user cannot be authenticated.
+                // Store the resulting permissions object in the request so it can be examined by any handler.
+                UserPermissions userPermissions = authentication.authenticate(req);
+                req.attribute(USER_PERMISSIONS_ATTRIBUTE, userPermissions);
+                // TODO stop using these two separate attributes, and use the permissions object directly
+                req.attribute(USER_EMAIL_ATTRIBUTE, userPermissions.email);
+                req.attribute(USER_GROUP_ATTRIBUTE, userPermissions.groups.iterator().next());
```
trevorgerhardt, line 108: I noticed the `events` collection in the database has a `groups` array, where every other collection has a single `accessGroup` string.

Unless we migrate everything else over we should revert `groups` being an array. Everywhere else we consider there to be a single `accessGroup`. We discussed a possible move to an array awhile back, but haven't had a need since we've been able to distinguish between administrators and everyone else.

abyrd: Yes, unless we have a specific plan to use multiple groups, we can switch it back to being a single string. It's a higher priority to make the database schema coherent across various tables.

abyrd: Marking this resolved since we've got a separate PR #648 to fix it.

abyrd:
I took a look and don't think this disrupts worker auto-shutdown at all, only changing the backend. This is ready for review now.

src/main/java/com/conveyal/analysis/components/HttpApi.java
```diff
@@ -79,14 +96,27 @@ public HttpApi (
...
             }
-            BackendMain.recordActivityToPreventShutdown();
+        });
+
+        sparkService.after((req, res) -> {
+            // Firing an event after the request allows us to report the response time,
+            // but may fail to record requests experiencing authentication problems.
```
trevorgerhardt, line 114: Does `after` get skipped when an exception is thrown? 
Should we add a separate event to the `exception` handler that records the request time and the exception?

src/main/java/com/conveyal/analysis/components/HttpApi.java
```diff
@@ -167,4 +196,9 @@ public HttpApi (
         return sparkService;
     }
 
+    // Maybe this should be done or called with a JVM shutdown hook
```
trevorgerhardt, line 199: Are there other services that we should add to the shutdown hook? AWS connections or the MongoDB driver?

trevorgerhardt (approved):
A few comments, but nothing that should hold this up.

src/main/java/com/conveyal/analysis/components/HttpApi.java
abyrd: Yes, I think we'll want to move toward cleanly shutting down MapDB, HTTP servers, AWS connections, MongoDB, everything in a JVM shutdown hook. I'll make a ticket for that.

src/main/java/com/conveyal/analysis/components/HttpApi.java
abyrd: I actually don't know, will have to dig in and see. I definitely intended to fire Exception events so we can show them on Slack and/or log them in the database. Will also ticket this one so we can follow up on it.

PR #647, opened 2020-11-05
Propagation cleanup

ansoncfit:
We implemented support for pickup-delay modifications in https://github.com/conveyal/r5/pull/537, using a couple phases of work. My commits through `2c8a110` focused on on-demand `BICYCLE_RENT`. Subsequent commits were focused on enabling `CAR` with options to specify service between certain "stop" polygons and "zone" polygons.

This PR harmonizes the approaches and addresses a number of outstanding tasks:
* https://github.com/conveyal/r5/commit/e8d5e7d0e4e4999353b0bf8121ddcd474faffe07 includes clarifying comments and a no-op refactor for clarity
* https://github.com/conveyal/r5/commit/aadb689faf9aac0623650ac95b2b5943fe6ac5fa applies `defaultWait` values specified in PickupDelay modifications
* https://github.com/conveyal/r5/commit/aadb689faf9aac0623650ac95b2b5943fe6ac5fa moves tracking of egress stop delays from LinkedPointSet to EgressCostTable, as suggested in a TODO. It also deletes comments about `BICYCLE_RENT` vs. `BICYCLE`. PickupDelay modifications now include a streetMode field, and we should just assume users intend to apply the delays in the modification if it is active in their scenario and its streetMode matches the requested streetMode.
* https://github.com/conveyal/r5/commit/4cd82b15691114989622def03023f7e0f0a23760 is a larger refactor addressing a couple of optimizations previously suggested in TODOs, moving egress stop/area "filtering" and "transforming" (i.e. addition of wait time delays) to shared methods that work for both `BICYCLE` and `CAR`.

### DIFF
--- a/src/main/java/com/conveyal/r5/streets/EgressCostTable.java
+++ b/src/main/java/com/conveyal/r5/streets/EgressCostTable.java
@@ -293,8 +293,10 @@ public class EgressCostTable implements Serializable {
                 // values with any stop-specific values.
                 if (egressStopDelaysSeconds != null) {
                     PickupWaitTimes.EgressService egressService = pickupWaitTimes.getEgressService(stopIndex);
-                    egressArea = egressService.serviceArea;
-                    egressStopDelaysSeconds[stopIndex] = egressService.waitTimeSeconds;
+                    if (egressService != null) {
+                        egressArea = egressService.serviceArea;
+                        egressStopDelaysSeconds[stopIndex] = egressService.waitTimeSeconds;
+                    }
                 }
 
                 if (streetMode == StreetMode.BICYCLE) {
--- a/src/main/java/com/conveyal/r5/streets/EgressCostTable.java
+++ b/src/main/java/com/conveyal/r5/streets/EgressCostTable.java
@@ -302,7 +302,7 @@ public class EgressCostTable implements Serializable {
                     sr.distanceLimitMeters = linkingDistanceLimitMeters;
                     sr.quantityToMinimize = linkageCostUnit;
                     sr.route();
-                    return linkedPointSet.extendDistanceTableToPoints(sr.getReachedVertices(), envelopeAroundStop);
+                    return linkedPointSet.extendCostsToPoints(sr, envelopeAroundStop, egressArea);
                 } else if (streetMode == StreetMode.CAR) {
                     // The speeds for Walk and Bicycle can be specified in an analysis request, so it makes sense above to
                     // store distances and apply the requested speed. In contrast, car speeds vary by link and cannot be
@@ -311,28 +311,7 @@ public class EgressCostTable implements Serializable {
                     sr.timeLimitSeconds = CAR_TIME_LINKING_LIMIT_SECONDS;
                     sr.quantityToMinimize = linkageCostUnit;
                     sr.route();
-                    // TODO optimization: We probably shouldn't evaluate at every point in this LinkedPointSet in case
-                    //      it's much bigger than the driving radius.
-                    PointSetTimes driveTimesToAllPoints = linkedPointSet.eval(
-                            sr::getTravelTimeToVertex,
-                            null,
-                            LinkedPointSet.OFF_STREET_SPEED_MILLIMETERS_PER_SECOND,
-                            null
-                    );
-                    // TODO optimization: should we make spatial index visit() method public to avoid copying results?
-                    TIntList packedDriveTimes = new TIntArrayList();
-                    for (int p = 0; p < driveTimesToAllPoints.size(); p++) {
-                        int driveTimeToPoint = driveTimesToAllPoints.getTravelTimeToPoint(p);
-                        if (driveTimeToPoint != Integer.MAX_VALUE) {
-                            packedDriveTimes.add(p);
-                            packedDriveTimes.add(driveTimeToPoint);
-                        }
-                    }
-                    if (packedDriveTimes.isEmpty()) {
-                        return null;
-                    } else {
-                        return packedDriveTimes.toArray();
-                    }
+                    return linkedPointSet.extendCostsToPoints(sr, envelopeAroundStop, egressArea);
                 } else {
                     throw new UnsupportedOperationException("Tried to link a pointset with an unsupported street mode");
                 }
@@ -340,45 +319,6 @@ public class EgressCostTable implements Serializable {
         }).collect(Collectors.toList());
         computeCounter.done();
         copyCounter.done();
-
-        // Now that we have a full set of tables, filter and transform them to reflect on-demand egress service.
-        // TODO optimization: we could return null immediately above for all stops not served by the on-demand egress:
-        // if (onDemandEgressService != null && ! onDemandEgressService.allows(stopIndex)) return null;
-        // We could also integrate the geographic filtering into that same loop and extendDistanceTableToPoints.
-        // In fact this is a major optimization since building car tables is very slow, and we might only need a few.
-        StreetLayer streetLayer = transitLayer.parentNetwork.streetLayer;
-        PickupWaitTimes pickupWaitTimes = streetLayer.pickupWaitTimes;
-        if (pickupWaitTimes != null && pickupWaitTimes.streetMode == streetMode) {
-            if (streetMode != StreetMode.CAR) {
-                // FIXME only cars have egress cost tables in seconds. Others will need a constant time offset field.
-                throw new RuntimeException("Only car egress tables can have a baked in time delay.");
-            }
-            for (int s = 0; s < stopToPointLinkageCostTables.size(); s++) {
-                PickupWaitTimes.EgressService egressService = pickupWaitTimes.getEgressService(s);
-                if (egressService == null || egressService.waitTimeSeconds < 0) {
-                    stopToPointLinkageCostTables.set(s, null);
-                    continue;
-                }
-                int[] costs = stopToPointLinkageCostTables.get(s);
-                TIntList filteredCosts = new TIntArrayList();
-                for (int i = 0; i < costs.length; i += 2) {
-                    int point = costs[i];
-                    int cost = costs[i + 1];
-                    // TODO normalize variable names (to costs?), these are not just times they may be distances.
-                    // TODO linkedPointSet.pointSet.getPointsInGeometry(), and pointInsideGeometry? default defs.
-                    double lat = linkedPointSet.pointSet.getLat(point);
-                    double lon = linkedPointSet.pointSet.getLon(point);
-                    if (GeometryUtils.containsPoint(egressService.serviceArea, lon, lat)) {
-                        filteredCosts.add(point);
-                        filteredCosts.add(cost + egressService.waitTimeSeconds);
-                    }
-                }
-                // null represents an empty array, which we presume may be more efficiently serializable than lots of
-                // references to a single empty array instance.
-                int[] filteredCostsArray = filteredCosts.isEmpty() ? null : filteredCosts.toArray();
-                stopToPointLinkageCostTables.set(s, filteredCostsArray);
-            }
-        }
     }
 
     /**
--- a/src/main/java/com/conveyal/r5/streets/EgressCostTable.java
+++ b/src/main/java/com/conveyal/r5/streets/EgressCostTable.java
@@ -313,7 +313,10 @@ public class EgressCostTable implements Serializable {
                 }
                 sr.quantityToMinimize = linkageCostUnit;
                 sr.route();
-                return linkedPointSet.extendCostsToPoints(sr, envelopeAroundStop, egressArea);
+                return linkedPointSet.extendCostsToPoints(sr.getReachedVertices()::get,
+                        sr.quantityToMinimize,
+                        envelopeAroundStop,
+                        egressArea);
             }
         }).collect(Collectors.toList());
         computeCounter.done();
--- a/src/main/java/com/conveyal/r5/streets/EgressCostTable.java
+++ b/src/main/java/com/conveyal/r5/streets/EgressCostTable.java
@@ -324,11 +324,12 @@ public class EgressCostTable implements Serializable {
      * Private constructor used by factory methods or other constructors to allow fields to be immutable.
      */
     private EgressCostTable (LinkedPointSet linkedPointSet,
-                            StreetRouter.State.RoutingVariable linkageCostUnit,
+                            EgressCostTable superCostTable,
                             List<int[]> stopToPointLinkageCostTables) {
         this.linkedPointSet = linkedPointSet;
-        this.linkageCostUnit = linkageCostUnit;
+        this.linkageCostUnit = superCostTable.linkageCostUnit;
         this.stopToPointLinkageCostTables = stopToPointLinkageCostTables;
+        this.egressStopDelaysSeconds = superCostTable.egressStopDelaysSeconds;
     }
 
     /**
@@ -384,7 +385,7 @@ public class EgressCostTable implements Serializable {
                 })
                 .collect(Collectors.toList());
 
-        return new EgressCostTable(subLinkage, superCostTable.linkageCostUnit, stopToPointLinkageCostTables);
+        return new EgressCostTable(subLinkage, superCostTable, stopToPointLinkageCostTables);
     }
 
     /**
--- a/src/main/java/com/conveyal/r5/streets/EgressCostTable.java
+++ b/src/main/java/com/conveyal/r5/streets/EgressCostTable.java
@@ -135,18 +135,19 @@ public class EgressCostTable implements Serializable {
                 : baseLinkage.getEgressCostTable(progressListener);
 
         this.linkedPointSet = linkedPointSet;
+        final StreetMode streetMode = linkedPointSet.streetMode;
 
         TransitLayer transitLayer = linkedPointSet.streetLayer.parentNetwork.transitLayer;
         int nStops = transitLayer.getStopCount();
         StreetLayer streetLayer = transitLayer.parentNetwork.streetLayer;
         PickupWaitTimes pickupWaitTimes = streetLayer.pickupWaitTimes;
 
-        if (pickupWaitTimes != null && linkedPointSet.streetMode == pickupWaitTimes.streetMode) {
+        if (pickupWaitTimes != null && streetMode == pickupWaitTimes.streetMode) {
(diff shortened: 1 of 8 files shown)


src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
```diff
@@ -497,36 +520,88 @@ public PointSetTimes eval (
...
+    }
+
+    /**
+     * Given accumulated time to reach vertices, add the remaining time needed to reach a point that is linked to
+     * an edge. This remaining time consists of a time traversing part of the edge (at onStreetSpeed), plus a
+     * time needed to travel the perpendicular distance (at offStreetSpeed) from the split point to the final point.
+     *
+     * @return minimum time needed to reach point, or Integer.MAX_VALUE if point is not reachable.
+     */
+    private int timeToPoint(int time0, int time1, int onStreetSpeed, int offStreetSpeed, int pointIndex) {
```
abyrd, line 594: The Javadoc should ideally say what time0 and time1 are, and what the units of speed are.

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
```diff
@@ -497,36 +520,88 @@ public PointSetTimes eval (
...
+                return true; // Point is outside supplied area, continue to next iteration.
             }
-            if (distanceTableToVertices.containsKey(edge.getToVertex())) {
-                t2 = distanceTableToVertices.get(edge.getToVertex()) + distances1_mm[p] + distancesToEdge_mm[p];
+
+            edge.seek(edges[p]);
+
+            int cost = Integer.MAX_VALUE;
+
+            if (routingVariable == RoutingVariable.DISTANCE_MILLIMETERS) {
```
abyrd, line 532: I see that the distance to point map is being generalized to a cost to point map, introducing different code paths for distance and duration. I'm still trying to contrast with how we handled this dichotomy before. This is a section I'd like to discuss.

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
```diff
@@ -479,14 +470,46 @@ public PointSetTimes eval (
...
+     * This is a pure function i.e. it has no side effects on the state of the LinkedPointSet instance.
+     *
+     * @param sr results of an on-street search from the transit stop
+     * @param distanceTableZone the envelope in FIXED POINT DEGREES within which we want to find all points.
+     * @param egressArea area served by on-demand service from this stop. If null, there are no restrictions on which
+     *                  points can be reached in the egress leg from this stop.
+     * @return A packed array of (pointIndex, cost), or null if there are no reachable points. Cost units match
+     * supplied sr.routingVariable
+     */
+    public int[] extendCostsToPoints(StreetRouter sr, Envelope distanceTableZone, Geometry egressArea) {
```
abyrd, line 500: The Envelope is still called "distanceTableZone" even though everything has been changed to "cost".

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
```diff
@@ -497,36 +520,88 @@ public PointSetTimes eval (
...
+            edge.seek(edges[p]);
+
+            int cost = Integer.MAX_VALUE;
+
+            if (routingVariable == RoutingVariable.DISTANCE_MILLIMETERS) {
+                cost = distanceToPoint(costTableToVertices, edge, p);
+            } else if (routingVariable == RoutingVariable.DURATION_SECONDS) {
+                int time0 = costTableToVertices.get(edge.getFromVertex());
+                int time1 = costTableToVertices.get(edge.getFromVertex());
+                if (time0 == 0 && time1 == 0) {
```
abyrd, line 537: These zeros as default values seem risky. Is this default value being used anywhere else?

abyrd (commented):
I have read this through but am finding the full set of changes somewhat hard to follow. I think it would be beneficial to work through it together during our work session.

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
```diff
@@ -497,36 +520,88 @@ public PointSetTimes eval (
...
+
+            edge.seek(edges[p]);
+
+            int cost = Integer.MAX_VALUE;
+
+            if (routingVariable == RoutingVariable.DISTANCE_MILLIMETERS) {
+                cost = distanceToPoint(costTableToVertices, edge, p);
+            } else if (routingVariable == RoutingVariable.DURATION_SECONDS) {
+                int time0 = costTableToVertices.get(edge.getFromVertex());
+                int time1 = costTableToVertices.get(edge.getFromVertex());
```
abyrd, line 536: This second one should be `edge.getToVertex()`

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
ansoncfit: I agree the 0 special values are risky. See comment below on setting `Integer.MAX_INTEGER` as the `noEntryValue` in the street router. But as you note, we need to check that changing the `noEntryValue` doesn't have unintended side effects elsewhere.

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
```diff
@@ -497,36 +520,88 @@ public PointSetTimes eval (
...
+                int onStreetSpeed = (int) edge.getCarSpeedMetersPerSecond() * 1000;
+                cost = timeToPoint(time0, time1, onStreetSpeed, OFF_STREET_SPEED_MILLIMETERS_PER_SECOND, p);
             }
-            int t = Math.min(t1, t2);
-            if (t != Integer.MAX_VALUE) {
-                if (t < distanceToPoint.get(p)) {
-                    distanceToPoint.put(p, t);
+
+            if (cost != Integer.MAX_VALUE) {
+                if (cost < costToPoint.get(p)) { // When is this false?
```
abyrd, line 546: Response posted in Slack: The method initializes distanceToPoint to an empty int-int map, then iterates over all the points in the envelope and saves one distance to each point in the map, keyed on the point number p. It does look like this conditional could only ever be hit once for a given value of p, and the default value before we write to the map is the max integer, so yes I think it would always be true. It looks like maybe the method was adapted from a pre-existing method that iterated over many candidates and kept the best one. The loop that associates the points with edges would work that way.

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
abyrd: Additional explanation from @ansoncfit: this zero-check is new and was not in pre-existing code. It's an optimization in response to the observation in a comment that `We probably shouldn't evaluate at every point in this LinkedPointSet in case it's much bigger than the driving radius`. Here we still do iterate over every point, but short circuit fast on unreachable points.

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
```diff
@@ -479,14 +470,46 @@ public PointSetTimes eval (
...
+     *
+     * @param sr results of an on-street search from the transit stop
+     * @param distanceTableZone the envelope in FIXED POINT DEGREES within which we want to find all points.
+     * @param egressArea area served by on-demand service from this stop. If null, there are no restrictions on which
+     *                  points can be reached in the egress leg from this stop.
+     * @return A packed array of (pointIndex, cost), or null if there are no reachable points. Cost units match
+     * supplied sr.routingVariable
+     */
+    public int[] extendCostsToPoints(StreetRouter sr, Envelope distanceTableZone, Geometry egressArea) {
+        return extendCostsToPoints(sr.getReachedVertices(),
```
abyrd, line 501: `TIntIntMap costTableToVertices` depends heavily on having a default value of zero for missing entries. This is not documented in the Javadoc. This does happen to be a characteristic of the map returned by `sr.getReachedVertices()`, but `sr.getTravelTimeToVertex()` inconsistently returns MAX_INT for unreachable vertices. The latter function is used in `LinkedPointSet.eval()`, which therefore has to include an extra check for MAX_INT before passing any values down into timeToPoint.

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
abyrd: Additional info from @ansoncfit: this is due to factoring out some repetitive code dealing with creating packed cost/distance arrays that appeared in several places. The two cases were apparently handled in separate copies of similar code before the refactoring in this PR.

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
ansoncfit: Could `getTravelTimesToVertex` be used instead of `sr.getReachedVertices()`? One caution -- the former calls `getStateAtVertex()`, which has javadoc stating "NB this should not be used for propagating to samples, as you need to apply turn costs/restrictions during propagation."

abyrd: I see, getReachedVertices is iterating over all the edges and letting edges leading into the same vertex compete with each other (since our states are actually at the end of edges). But getStateAtVertex (via getTravelTimeToVertex) is doing the same thing.

I don't see how LinkedPointSet#eval() could be handling turn restrictions when propagating. It is only looking at one travel time per vertex, the lowest one. We need to decide whether imposing turn restrictions and costs on the final turn is important enough to justify making the propagation more complex and slow, because even the current approach does not seem to apply them.

src/main/java/com/conveyal/r5/streets/EgressCostTable.java
```diff
@@ -99,6 +99,14 @@
...
     private transient List<TIntIntMap> pointToStopLinkageCostTables;
 
+    /**
+     * For each transit stop, extra seconds to wait due to a pickup delay modification (e.g. for autonomous vehicle,
+     * scooter pickup, etc.). We track this separately so the egress time limit can exclude waiting time.
+     *
+     * Should remain null if there is not a pickup delay modification applied for this streetmode.
+     */
+    public int[] egressStopDelaysSeconds;
+
```
ansoncfit, line 109: Moved from LinkedPointSet

src/main/java/com/conveyal/r5/streets/EgressCostTable.java
```diff
@@ -127,8 +135,19 @@ public EgressCostTable (LinkedPointSet linkedPointSet, ProgressListener progress
                 : baseLinkage.getEgressCostTable(progressListener);
 
         this.linkedPointSet = linkedPointSet;
+        final StreetMode streetMode = linkedPointSet.streetMode;
+
+        TransitLayer transitLayer = linkedPointSet.streetLayer.parentNetwork.transitLayer;
+        int nStops = transitLayer.getStopCount();
```
ansoncfit, line 141: Re-ordered (were further down, but now needed up here for egressStopDelays)

src/main/java/com/conveyal/r5/streets/EgressCostTable.java
```diff
@@ -269,96 +307,38 @@ public EgressCostTable (LinkedPointSet linkedPointSet, ProgressListener progress
...
-                    );
-                    // TODO optimization: should we make spatial index visit() method public to avoid copying results?
-                    TIntList packedDriveTimes = new TIntArrayList();
-                    for (int p = 0; p < driveTimesToAllPoints.size(); p++) {
-                        int driveTimeToPoint = driveTimesToAllPoints.getTravelTimeToPoint(p);
-                        if (driveTimeToPoint != Integer.MAX_VALUE) {
-                            packedDriveTimes.add(p);
-                            packedDriveTimes.add(driveTimeToPoint);
-                        }
-                    }
```
ansoncfit, line 299: Consolidated in linkedPointSet.extendCostsToPoints (around https://github.com/conveyal/r5/pull/647/files#diff-dd60435552e26291d0b9b9a9ad8bfa02c92db82d5217698bc03f98665b96eebdR537)

src/main/java/com/conveyal/r5/streets/EgressCostTable.java
```diff
@@ -269,96 +307,38 @@ public EgressCostTable (LinkedPointSet linkedPointSet, ProgressListener progress
...
-                for (int i = 0; i < costs.length; i += 2) {
-                    int point = costs[i];
-                    int cost = costs[i + 1];
-                    // TODO normalize variable names (to costs?), these are not just times they may be distances.
-                    // TODO linkedPointSet.pointSet.getPointsInGeometry(), and pointInsideGeometry? default defs.
-                    double lat = linkedPointSet.pointSet.getLat(point);
-                    double lon = linkedPointSet.pointSet.getLon(point);
-                    if (GeometryUtils.containsPoint(egressService.serviceArea, lon, lat)) {
-                        filteredCosts.add(point);
-                        filteredCosts.add(cost + egressService.waitTimeSeconds);
```
ansoncfit, line 342: Consolidated in linkedPointSet.extendCostsToPoints (around https://github.com/conveyal/r5/pull/647/files#diff-dd60435552e26291d0b9b9a9ad8bfa02c92db82d5217698bc03f98665b96eebdR537)

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
```diff
@@ -477,56 +459,139 @@ public PointSetTimes eval (
...
+            } else if (routingVariable == RoutingVariable.DURATION_SECONDS) {
+                // The routing variable is seconds only if we're doing a car search, so look up the car speed on the
+                // linked edge.
+                int onStreetSpeed = (int) (edge.getCarSpeedMetersPerSecond() * 1000);
+                cost = timeToPoint(costToVertex, edge, p, onStreetSpeed);
             }
-            int t = Math.min(t1, t2);
-            if (t != Integer.MAX_VALUE) {
-                if (t < distanceToPoint.get(p)) {
-                    distanceToPoint.put(p, t);
```
ansoncfit, line 513: Surrounding code factored out into consolidated `distanceToPoint` method

src/main/java/com/conveyal/r5/streets/LinkedPointSet.java
```diff
@@ -387,87 +382,74 @@ public int size () {
...
-            if (time1 != Integer.MAX_VALUE) {
-                time1 += distances1_mm[i] / onStreetSpeed + offstreetTime;
+                travelTimes[i] = // origin.distanceToEdge_mm / offStreetSpeed + TODO origin to origin split point
+                                onStreetDistance_mm / onStreetSpeed + // along street
+                                distancesToEdge_mm[i] / offStreetSpeed; // from destination split point to destination
+            } else {
+                travelTimes[i] = timeToPoint(timeToVertex, edge, i, onStreetSpeed);
             }
-
-            travelTimes[i] = time0 < time1 ? time0 : time1;
```
ansoncfit, line 470: Surrounding code factored out into consolidated `timeToPoint` method

ansoncfit:
I've added some in-line comments that should facilitate the next round of review.

We've already discussed a few of the optimizations. Recent testing on staging confirms some expected speedups.

To summarize a few of the other changes, there are multiple cases where times/distances need to be extended from vertices to points:
- finding times directly to destinations without transit (using pointset.eval() with the results of a street router search)
- finding times during egress propagation for:
  - car, using freshly calculated times to vertices
  - bike, using freshly calculated distances to vertices
  - walk, using pre-saved distance tables

Prior to this PR, certain logic was inconsistent or repeated (e.g. packing arrays, checking for overflow, choosing the minimum travel time from two vertex options at the destination edge). Refactors in this PR attempt to clarify and consolidate the logic with a few shared functions:

```
direct ---------- eval() ------------ timeToPoint()
                                     /
        car -- extendCostsToPoints()
       /
egress -- bike -- extendCostsToPoints() ---- distanceToPoint()
       \                                     /
        walk -- extendDistanceTableToPoints() 

```

src/main/java/com/conveyal/r5/streets/StreetRouter.java
```diff
@@ -229,7 +231,7 @@ public TIntIntMap getReachedStops() {
      * objective variable for the optimal path to that vertex.
      */
     public TIntIntMap getReachedVertices () {
-        TIntIntMap result = new TIntIntHashMap();
+        TIntIntMap result = new TIntIntHashMap(DEFAULT_CAPACITY, DEFAULT_LOAD_FACTOR, -1, Integer.MAX_VALUE);
```
abyrd, line 234: Javadoc for this method should mention that the returned map has a noEntryValue of Integer.MAX_VALUE and a noEntryKey of -1 independent of the objective variable.

As I was auditing the code, this constructor threw me off for a while because I misinterpreted the -1 as the noEntryValue, but it's actually the noEntryKey. I didn't expect Trove to have a configurable noEntryKey, and don't really understand why. Although -1 does seem more appropriate than zero, even after digging into the Trove source code I don't quite understand whether there are any side effects to setting noEntryKey to -1. 

Taking a step back, maybe a more robust solution would be for this method to return a more clearly defined type than TIntIntMap, such as CostToVertexFunction (which has Javadoc unambiguously stating that MAX_VALUE always means unreachable). Then TransitLayer#stopToVertexDistanceTables could also be more strongly typed as a List<CostToVertexFunction>, or better yet an even more strongly typed List<DistanceToVertex>, where DistanceToVertex extends CostToVertexFunction but doesn't add anything other than a method returning a RoutingVariable of DISTANCE_MILLIMETERS. Contrary to the TODO comment on CostToVertexFunction I don't think we'd want to wrap CostToVertexFunction to achieve this, but instead make it an abstract class with two subclasses (one returning DISTANCE_MILLIMETERS and the other DURATION_SECONDS from a getRoutingVariable method). CostToVertexFunction could even be made less functional, given a name like CostToVertex (or CostToVertexTable) and extend or contain a TIntIntMap, so StreetRouter or anything else could just directly instantiate a CostToVertex and add values to it, making it impossible to misconfigure any maps with wrong noEntryValues.

That may be too much to add to this PR though - we could try that later in another round of refactoring.

abyrd (approved):
I read through this pretty thoroughly and it looks good to me. Given the depth of the changes though I can't be completely sure, and there is a potential for new routing bugs, so I suggest that we ensure this is covered by a new batch of regression tests before we release it. I did add a request for a small amount of Javadoc you may want to add before merging, and a suggestion for a refactor allowing stronger typing that we could do at a later stage.

PR #648, opened 2020-11-06
Turn groups into a string

trevorgerhardt:


abyrd:
We intentionally converted the group from a string to a set of strings to allow user membership in multiple groups. Are we no longer planning to do that? This is part of the wider discussion on permissions management.

trevorgerhardt:
I understand that this is very much related to ongoing discussions we are having on other subjects. That being said...

I believe we prematurely converted the `accessGroup` to a `Set`. It didn't affect anything when we initially did it because it was only being used in the Java class related to authentication and never getting stored. The newest model, `Event`, stores `accessGroup`s in an array deviating from the common model shared by all other database models. 

We shouldn't be in limbo. We should either 

1) Merge this PR to convert `Event` to use `accessGroup`. Migrate stored `Event`s to share the same structure as other models. 
2) Convert all other Models to use an `accessGroup` array and migrate all other stored objects to use the same structure.

Because we have no solid plan for using multiple `accessGroup`s at the moment. I believe we should do #1 and plan out the changes ahead of using the new structure.

trevorgerhardt:
Btw, I didn't add a description initially because this was PR done in haste and mainly meant to be a discussion to be resolved, not something that was urgent. Thanks for getting the ball rolling! @abyrd

abyrd:
Ah, I just now saw the reference over at https://github.com/conveyal/r5/pull/646#discussion_r518558604. The motivation for the change now makes perfect sense - the fact that they are saved in the database in events, whose groups field differs from every other document. But without a description I didn't catch this reason. I just saw the newly created PR from Github notifications.

abyrd (approved):
Looks correct to me and build passes, any problems should be immediately visible in development and testing.

PR #651, opened 2020-11-10
Vector tiles

abyrd:
This is a basic proof of concept creating Mapbox vector tiles. Several more steps are needed for this to be production ready, some of which are described in the Javadoc in the PR. Notably we should improve the spatial indexing and integrate it with any indexing already in the gtfs-lib / gtfs-api code.

The PR also includes an endpoint that can be adapted to proxy requests from the UI to a worker on a given network, which could be used to fetch debug tiles from a worker.

Together these two endpoints should allow visualizing all the OSM and GTFS data being used in a particular network, or just visualizing the routes in an uploaded GTFS feed.

There is simple Mapbox GLJS 2 client code included in `src/main/resources/vector-client`.


ansoncfit (commented):
Test LODES fetch before merging.

build.gradle
```diff
@@ -230,12 +232,12 @@ dependencies {
...
-    implementation 'com.google.protobuf:protobuf-java:2.6.1'
+    // Google Protocol Buffers are used in decoding OSM data and seamless census data (in Geobuf format).
+    // implementation 'com.google.protobuf:protobuf-java:3.13.0'
 
-    // OSM PBF format support. Transitive dependency on Protobuf 3.4.0.
-    // TODO investigate any conflict with above
-    implementation 'org.openstreetmap.osmosis:osmosis-osm-binary:0.45'
+    // OSM PBF format support. Transitive dependency on Protobuf 3.12.2,
+    // align with mapbox-vecor-tile and geobuf generated classes
+    implementation 'org.openstreetmap.osmosis:osmosis-osm-binary:0.48.3'
```
ansoncfit, line 240: To summarize: seamless census was depending on 2.6.1 via this import and osmosis was pulling in 3.4.0; now the single 3.12.2 version (compatible with mapbox-vector-tile and geobuf) is used via transitive dependency from osmosis.

ansoncfit (approved):
Tested LODES import on staging, worked well.

ansoncfit (approved):

abyrd:
This exposes a new endpoint that could have performance implications for the backend if used heavily. However that endpoint is authenticated, and would only realistically be hit if we add functionality to our UI. We plan to do that gradually, starting with the bundle editor tab, so we can evaluate the performance effects before expanding usage.

PR #658, opened 2020-11-18
Migrate from Junit 3 and 4 to Junit 5

abyrd:
This PR migrates all our tests from the Junit 3 and Junit 4 testing frameworks to Junit 5, which has been out for about three years and is widely used. Junit 5 is adapted to Java 8+ making use of lambda functions and streams, and has better support for parameterizing, repeating, and nesting tests.

This PR serves as groundwork for adding more thorough and wide-ranging tests of the travel time computer.

It should be a pure refactor/migration, with all tests essentially equivalent and still passing.

It does remove a couple of tests of the HTTP API that start up a full server. That territory is better covered by our Cypress full stack tests, and was related to several problems we were experiencing including cruft files left over from snapshotting and the need to maintain test-specific config files and classes for wiring up analysis components. This allows us to remove several dependencies.


trevorgerhardt (approved):
LGTM

PR #662, opened 2020-12-07
Limit to one iteration per minute unless network has frequency-based patterns

### LINKED ISSUE #661: Don't perform multiple iterations per minute unless the transit network has frequency-based routes
I remember a comment somewhere explaining why we choose to repeat results to obtain an expected/requested number of iterations. But I don't think we need to perform multiple iterations per minute.

ansoncfit:
Fixes #661

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/TravelTimeReducer.java
+++ b/src/main/java/com/conveyal/r5/analyst/TravelTimeReducer.java
@@ -98,8 +98,13 @@ public class TravelTimeReducer {
                 // HALF_HEADWAY boarding, returning a single travel time per departure minute per destination.
                 this.timesPerDestination = task.getTimeWindowLengthMinutes();
             } else {
-                // MONTE_CARLO boarding, using several different randomized schedules at each departure time.
-                this.timesPerDestination = task.getTimeWindowLengthMinutes() * task.getMonteCarloDrawsPerMinute();
+                if (network.transitLayer.hasFrequencies) {
+                    // MONTE_CARLO boarding, using several different randomized schedules at each departure time.
+                    this.timesPerDestination = task.getTimeWindowLengthMinutes() * task.getMonteCarloDrawsPerMinute();
+                } else {
+                    // No frequency-based routes, so only one iteration per departure time.
+                    this.timesPerDestination = task.getTimeWindowLengthMinutes();
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


abyrd:
Thanks, this is definitely something worth cleaning up. 

At first I was surprised by how simple this change is, and wondered why it works. It seems strange that we went to the trouble of replicating results and processing those duplicate results if it was not necessary somewhere downstream. And I was wondering if this would work in special cases like HALF_HEADWAY. As I mentioned in comments on #661 we have several different variables and methods with slightly different meanings holding numbers of iterations which we should simplify. They make it harder to understand what's going on in code like the statements changed here.

I see that the router adapts to differing numbers of iterations via the else clause at L430 of FastRaptorWorker. Before the changes in this PR it was apparently already producing only one iteration per departure time (no MC randomization), but replicating the results as many times per minute as was requested by the caller. That else block is only ever run when the transit layer has no pure frequency routes. With this PR, it seems to me that now nIterationsPerMinute can only be 1 in this block, so the result is essentially unidimensional (the outer array always has length 1) and the loop is only ever run once. If this is the case, we should insert an assertion that `iterationsPerMinute == 1` and then remove the loop and hard-wire the outer array dimension to 1. Besides providing runtime checks that assumptions are correct, it also makes it clear to readers of the code that this block no longer replicates iterations. We could later add more assertions about array lengths e.g. when boarding mode is HALF_HEADWAY.

In response to the comments in this else block at L430 of FastRaptorWorker, I traced execution and confirmed that in existing code on pure scheduled networks, this code block was being run and replicating the results. With the UI default of 200 MC draws selected, each departure minute yields two identical arrays of travel times. This means that travelTimesToStopsPerIteration in FastRaptorWorker#route is twice as large as it needs to be (240 total iterations over a 2 hour departure window). That inflated array is returned up to TravelTimeComputer and from there passed down into PerTargetPropagater#propagateTransit, where we propagate twice as many values to every destination, then sorted twice as many values to extract percentiles. Of course if more MC draws were specified in the request, this could be 10 times more values instead of twice as many. This could cause significant amounts of extra computation in the propagation phase. I guess the reason that this has gone largely unnoticed is that we tend to develop and test on the complex cases (with MC draws) and then treat pure scheduled networks as an easier base case. But in reality many of our largest analyses are on pure scheduled networks.

All of the values from all these iterations are collated together and fed into a single percentile extraction operation. So contrary to what I commented on #661 this does not increase network traffic, it just leads to copying, scanning, and sorting unnecessarily long lists of identical values.

abyrd:
Closing, as this refactor was included in the merged PR #634, around https://github.com/conveyal/r5/pull/634/files#diff-5f18165b58fbd4226f78715339d1f19234dc278ebc96efe9da8b04b793ba6381R424

PR #666, opened 2020-12-11
synchronize reads from map to avoid concurrent modification

abyrd:
These were previously used only in tests but will be used in auto shutdown checks.


ansoncfit (approved):
I don't see the shutdown checks using `logJobStatus()` yet, but no problem making it synchronized in anticipation of such use.

PR #667, opened 2020-12-21
Additional path info changes

ansoncfit:
This implements additional ideas/cleanup discussed in #634 and

- Adds a package (transit.path) with classes:
  - StopSequence: The legs of an itinerary with transit (access, egress, boardStops, alightStops, and in-vehicle times)
  - PatternSequence: stopSequence and the patterns ridden between stops
  - Path (rename Itinerary?): patternSequence and associated departure time and waiting times for a single Raptor iteration
  - RouteSequence: aggregation grouping patterns by route. A further extension grouping by mode might be useful in some cases to reduce output file sizes.
- Replaces most Analysis uses of the profile.Path class (which had various extra fields) with these new transit.path classes.
- Returns JSON with human-readable path details for a single OD pair to the UI when toLat, toLon, and `recordPaths` set in the analysisRequest
- Cleans up the broker assembly and CSV writing of path results from workers processing regional analyses with freeform origins and destinations
- Allows autoscaling for regional analyses with freeform origins (limited to 5 workers for now)
- Add endpoint for downloading CSV files from FileStorage.

Still to-do (in a subsequent PR):
- [ ] Review `text/plain` mime content types on returned URLs
- [ ] Clean up FileStorage keys (especially consistent/correct use of .gz extension in already-gzipped files on S3).
- [ ] Clean up specification of origin pointset in analysis request (using an explicit pointset type field for regional analysis origins, rather than just taking the presence of an `originPointSetId` value as an indication of freeform origins?).
- [ ] Check behavior when `oneToOne = true`. This was implemented for travel time CSV writing with large numbers of linked OD pairs, but it has not been tested successfully with path CSV writing.
- [ ] Re-enable Taui site creation (with associated path scoring updates?), which still relies on the previous profile.Path class.
- [ ] Deprecate the profile.Path class (used for Point-to-Point routing) and replace remaining uses of it (in Taui) with the new transit.path.Path; remaining uses in point-to-point router will be removed (probably along with the p-to-p router itself, which will interfere with R5R and maybe Matt's fare debugging)


src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
```diff
@@ -354,11 +355,12 @@ private Object getCsvResults (Request req, Response res) {
...
 
-        String key = regionalAnalysisId + resultType + ".csv.gz";
+        // TODO result path is stored in model, use that?
+        String key = regionalAnalysisId + '_' + resultType + ".csv.gz";
         FileStorageKey fileStorageKey = new FileStorageKey(config.resultsBucket(), key);
-        JSONObject json = new JSONObject();
-        json.put("url", fileStorage.getURL(fileStorageKey));
-        return json.toJSONString();
+
+        res.type("text");
```
ansoncfit, line 362: Doesn't the type need to be a MIME type?

Used here: https://github.com/conveyal/analysis-ui/commit/f714b49b74f5b08792ea3c8f8896cd88e8c0a478#diff-b19858859df3fd3e2d21ecd70a5d70a41995b47fa337915004a99b7eba4d39faR122-R127

abyrd (approved):
Reviewed all new changes since we last approved the changeset. All remaining work to be done (on a subsequent PR) is documented in the PR description comment as a checklist.

PR #670, opened 2020-12-30
Minor fixes for bugs found when modifying Freeform pointset upload

trevorgerhardt:
Count field was not being checked during freeform upload. It appears like it was meant to be, but the id field was instead checked twice.

During form upload, if the name field did not exist it would throw an `AnalysisServerException` that was then caught and re-thrown but with a new message that was less specific and harder to decipher. This refactors the form to allow those exceptions to get thrown to the appropriate handler.


ansoncfit (approved):
Thanks for catching my mistake (probably a too hasty copy/paste)

ansoncfit:
Actually, as just discussed with @trevorgerhardt, we don't need to require a count field. Later in the fromCsv method, there is sane default behavior if no count is specified:

```
// If count column was specified and present, use it. Otherwise, one opportunity per point.
ret.counts[rec] = countCol < 0 ? 1D : Double.parseDouble(reader.get(countCol));
```

PR #674, opened 2021-01-19
Fix getCsvResults end point

trevorgerhardt:
Notes: 
- `analysis.complete` does not seemed to be set correctly, so until we fix that we shouldn't use it.
- Downloading locally does not currently work due to the `csv.gz` file type. That needs to be fixed in the File package.
- If we are storing the URLs we should be able to use them in the download endpoint right? Either way, what we store should be compatible with the `FileStorageKey` so that we don't need to manually create the URL in the controller and can just do something like `new FileStorageKey(config.resultsBucket(), analysis.resultStorage.get(resultType))`.


ansoncfit (approved):
Merging now, we can discuss further changes as part of #667

Issue #650, opened 2020-11-10
Write path information


ansoncfit:
Path details (boarding/alighting stops, patterns, breakdown of travel time into walk/wait/in-vehicle) would be useful for batch origin-destination travel time calculations, as well as debugging single-point results.

For batch analyses, we could add these to the existing CSV writing. This would involve sending larger messages to the broker, so our initial implementation could include guardrails rather than fully developed, adaptive scaling.

For single-point requests, a quick UI update could set the request `toLat`/`toLon` when the destination marker is pinned, and the response could include a summary of path details (e.g. paths roughly corresponding to the standard travel time cutoffs).

abyrd:
More path information can now be written thanks to PR #634 and #667. There are some tasks remaining in a subsequent PR:
- [ ] Review `text/plain` mime content types on returned URLs
- [ ] Clean up FileStorage keys (especially consistent/correct use of .gz extension in already-gzipped files on S3).
- [ ] Clean up specification of origin pointset in analysis request (using an explicit pointset type field for regional analysis origins, rather than just taking the presence of an `originPointSetId` value as an indication of freeform origins?).
- [ ] Check behavior when `oneToOne = true`. This was implemented for travel time CSV writing with large numbers of linked OD pairs, but it has not been tested successfully with path CSV writing.
- [ ] Re-enable Taui site creation (with associated path scoring updates?), which still relies on the previous profile.Path class.
- [ ] Deprecate the profile.Path class (used for Point-to-Point routing) and replace remaining uses of it (in Taui) with the new transit.path.Path; remaining uses in point-to-point router will be removed (probably along with the p-to-p router itself, which will interfere with R5R and maybe Matt's fare debugging)

ansoncfit:
Additional cleanup: 

https://github.com/conveyal/r5/blob/06bc4d3e61ed002671ee219a828779c90c308ff8/src/main/java/com/conveyal/analysis/components/broker/Job.java#L139-L140

ansoncfit:
https://github.com/conveyal/r5/blob/96a965c9258a68513481747370e5f094b25c4173/src/main/java/com/conveyal/analysis/results/MultiOriginAssembler.java#L159-L160

PR #675, opened 2021-01-26
Further path cleanup

ansoncfit:
- Path results now include rows for direct itineraries (e.g. walking without transit)
- Re-enables oneToOne CSV results (part of #650)


abyrd (approved):
Looks good. Related follow-up changes will be on PR #678.

Issue #677, opened 2021-01-29
Allow distance decay functions truncated at 120 minutes


abyrd:
When using non-step decay functions with higher cutoffs (e.g. logistic decay with cutoff over 60 minutes), an analysis may fail with the error "Distance decay function must reach zero at or before 120 minutes." This is one of several general assertions that are checked on any decay function, but this one can easily fail on reasonable inputs. We should tolerate functions that are still above zero at 120 minutes.

Nonetheless, as this may produce surprising results we will want to show a warning to the user (e.g. in single point results). But once you're aware of the situation, the result is a straightforward interaction with our usual 120 minute travel time cap.

Issue #660, opened 2020-12-04
Improve Taui path scorer


abyrd:
The Taui path scorer currently gets the paths closest to the fastest travel time. We want to instead get the most common paths that close to the chosen percentile of travel time.

- [ ]  add validation that we have only one percentile
- [ ]  have PathScorer can remember percentile index or target time
- [ ]  Find most common paths, and among those select the ones closest to target time
- [ ]  Coefficients for scoring, thresholds for paths close to target time

ansoncfit:
Note that we select the itineraries closest to the fastest travel time (https://github.com/conveyal/r5/blob/v6.9/src/main/java/com/conveyal/r5/profile/PerTargetPropagater.java#L285), even though the path scorer could be called with travel times at other percentiles (https://github.com/conveyal/r5/blob/v6.9/src/main/java/com/conveyal/r5/analyst/PathScorer.java#L87-L93)

PR #678, opened 2021-02-09
Cleanup of paths, freeform, and OD matrix functionality

abyrd:
Recently much work has been done on reporting information other than accessibility, such as full paths or travel times. This is often done using freeform points (rather than grids) as origins or destinations. This PR will include various additional fixes and cleanup that were agreed upon in reviews of previous PRs on these subjects.

This includes:
- Issue #660 (improving the Taui path scorer) and reuse of that scorer in selecting a small set of paths to report in single point or regional analyses, including those with freeform pointsets.
- The list of follow up points on #650 in this comment https://github.com/conveyal/r5/issues/650#issuecomment-766857585


src/main/java/com/conveyal/analysis/results/MultiOriginAssembler.java
```diff
@@ -159,8 +160,8 @@ public MultiOriginAssembler (RegionalAnalysis regionalAnalysis, Job job, String
                 // It's kind of fragile to read from an external network service here. But this is
                 // only triggered when destinations are freeform, which is an experimental feature.
                 destinationPointSet = PointSetCache.readFreeFormFromFileStore(job.templateTask.grid);
-                if (job.templateTask.recordTimes && !job.templateTask.oneToOne) {
-                    if (nOriginsTotal * destinationPointSet.featureCount() > 1_000_000) {
+                if ((job.templateTask.recordTimes || job.templateTask.includePathResults) && !job.templateTask.oneToOne) {
+                    if (nOriginsTotal * destinationPointSet.featureCount() > 16_000_000) {
```
abyrd, line 164: O*D will limit the size of the output CSV, but it won't really limit the number of destinations, In the extreme, 16 origins and 1 million destinations is allowed, which would still produce very large results from the worker (probably all 16 million at once). Maybe we should restrict D in addition to or instead of O*D.

abyrd: Also, the message below does not report the true limit. I'll add a destination limit and some new messages.

abyrd:
The Simpson Desert tests seem to be failing occasionally, but they pass when re-run. This is probably because they are checking Monte Carlo results against theoretical distributions, and the goodness of fit can change from one run to the next.

abyrd:
Recent changes to Taui Path/PatternSequence have been tested locally on a tiny Taui regional run. Files were successfully generated without exceptions and uploaded to S3. the XY_paths.dat were downloaded, gunzipped and examined in a hex editor. They appear to contain valid path information.

src/main/java/com/conveyal/r5/analyst/cluster/AnalysisWorker.java
```diff
@@ -523,7 +523,10 @@ protected void handleOneRegionalTask(RegionalTask task) {
             int maxCutoffMinutes = Arrays.stream(task.cutoffsMinutes).max().getAsInt();
             int maxTripDurationSeconds = task.decayFunction.reachesZeroAt(maxCutoffMinutes * SECONDS_PER_MINUTE);
             int maxTripDurationMinutes = (int)(Math.ceil(maxTripDurationSeconds / 60D));
-            checkState(maxTripDurationMinutes <= 120, "Distance decay function must reach zero at or before 120 minutes.");
+            if (maxTripDurationMinutes > 120) {
+                LOG.warn("Distance decay function reached zero above 120 minutes. Capping travel time at 120 minutes.");
+                maxTripDurationMinutes = 120;
+            }
```
ansoncfit, line 529: Fixes https://github.com/conveyal/r5/issues/677

src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
```diff
@@ -509,16 +521,18 @@ private RegionalAnalysis createRegionalAnalysis (Request req, Response res) thro
...
-        if (analysisRequest.recordPaths) regionalAnalysis.addCsvStoragePath(Result.PATHS, config.resultsBucket());
-        if (analysisRequest.recordAccessibility) regionalAnalysis.addCsvStoragePath(Result.ACCESS, config.resultsBucket());
-        Persistence.regionalAnalyses.modifiyWithoutUpdatingLock(regionalAnalysis);
 
         // Register the regional job with the broker, which will distribute individual tasks to workers and track progress.
         broker.enqueueTasksForRegionalJob(regionalAnalysis);
 
+        // Flush to the database any information added to the RegionalAnalysis object when it was enqueued.
+        // This includes the paths of any CSV files that will be produced by this analysis.
+        // TODO verify whether there is a reason to use regionalAnalyses.modifyWithoutUpdatingLock() or put().
```
trevorgerhardt, line 533: `put` updates the `nonce`, `createdAt`, and `updatedAt` and should be used by in simple updates from HTTP handlers. `modifyWithoutUpdatingLock` should be used when updating a model multiple times by the backend. There could be a better name for it, suggestions welcome.

trevorgerhardt: There's a case that could be made to always use `put` also 🤷

abyrd: The main thing I was wondering is why you'd avoid updating the nonce and update time when something was changed by the backend. I just wasn't sure of the reasoning - we should probably explain it on the method's Javadoc.

trevorgerhardt: It may be a premature optimization, mainly skipping updates to fields that aren't needed when the multiple writes are done by the backend without user input.

Looking at all the instances it's used, I'm not sure it's necessary anywhere except in `BundleController.setServiceBundleDates`.

trevorgerhardt: Even there, it could probably be removed

trevorgerhardt (approved):

PR #682, opened 2021-02-12
fix(TravelTimeReducer): null check, fixes #679

### LINKED ISSUE #679: NPE in TravelTimeReducer: travelTimeResult is null
When running a regional analysis after #665 was merged I get an NPE. This line appears to be the culprit: https://github.com/conveyal/r5/pull/665/files#diff-98772725b2b1ffbbef0b035f1bed2f7d91fd116f2b1afe72aa296323530a95d4R205

A few questions:
1. Should we be checking the `calculateTravelTimes` flag or more direclty if `travelTimeResult` is `null` before calling that method?
2. Should we default `recordTimes = true` on a `RegionalTask`? 
3. Is there a UI change that was supposed to occur simultaneously that I was unaware of?

abyrd:


abyrd:
Manually ran Cypress tests locally. Cypress tests are passing on a merge of this PR #682 merged with #678 in its current form.

trevorgerhardt (approved):

PR #683, opened 2021-02-13
No local CORS

mattwigway:
This disables CORS when running without authentication locally, and requires conveyal/analysis-ui#1457 to proxy to the backend. It makes no changes in production.

### DIFF
--- a/src/main/java/com/conveyal/analysis/components/HttpApi.java
+++ b/src/main/java/com/conveyal/analysis/components/HttpApi.java
@@ -88,7 +88,11 @@ public class HttpApi implements Component {
             //       Or now with non-static Spark we can run two HTTP servers on different ports.
 
             // Set CORS headers, to allow requests to this API server from any page.
-            res.header("Access-Control-Allow-Origin", "*");
+            // but do not do this when running offline with no auth, as this may allow a malicious website to use a local
+            // browser to access the analysis server.
+            if (!config.offline()) {
+                res.header("Access-Control-Allow-Origin", "*");
+            }
 
             // The default MIME type is JSON. This will be overridden by the few controllers that do not return JSON.
             res.type("application/json");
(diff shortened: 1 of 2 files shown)


abyrd:
Hi @mattwigway, I know CORS issues have come up several times and have been bothering you for a while but have not been resolved. Hopefully we can get this resolved now.

For discussion and future reference I think any PRs should be connected to a ticket describing what root problem is being addressed. I know a few tickets have been created before, but they were additions to issues about sanitizing inputs, or were limited to discussion of a specific service like PointToPointRouterServer. So I would like to clearly document in a single issue what CORS issue we're working on (even if that means copy/pasting some text from other issues to summarize the problem).

Am I correct in understanding that you want to remove the CORS headers for security reasons in local usage (when authentication is disabled)? This would be protecting against the case where a user is tricked into executing a malicious script while running a local instance of the Analysis backend with authentication disabled, and the malicious script is then able to access the analysis API endpoints?

If CORS headers are simply removed for simplicity, and the solution in this PR is applied, the front end will be sending data through its own proxy to the backend, which then proxies certain requests to a worker. I am already not excited about proxying HTTP through the backend to a worker in local operation, and passing through two proxies feels like continuing in this direction, adding more memory buffers, latency, and potential points of failure or misconfiguration. Although in production usage there is a reverse proxy like nginx in front of the whole system, which runs communications through a similar pipeline of repeated proxies, it still feels to me like something to avoid if possible. CORS itself always feels like an immense, awkward hack to me, and I'm hesitant to bend our system more and more to accommodate it.

If we consider this a security risk and want to eliminate it, I think I'd rather just make the allow-origin headers configurable and set in the example config to localhost and the default port used by a local frontend. The origin (protocol, port, and host) of the frontend should be fairly easy to determine in local usage and would rarely need to be modified.

I think I'd prefer not to have to configure this on two different proxies in two different ways (nginx and the frontend's JS proxy) if it can be done at the source in our API endpoints.

In production usage we actually want to de-restrict the origins that can contact the API so people can perform analyses with custom scripts, but in that case it has additional authentication mechanisms.

So: If my understanding is right on all this, I can create an issue about the reasons for making CORS headers more restrictive, and then my preferred solution would be to make it configurable with `localhost:3000` as the default, to be overridden in other environments.

abyrd:
Thinking about this from a few different angles, I think the real ideal is to fully enforce the single origin policy and not use CORS at all. CORS inherently increases latency by doubling up all the requests, so it seems best to confine it to exceptional cases. 

Eliminating CORS would not mean running the UI server on the same machine as the API server - in fact that would not be sufficient since they would be running on different ports. It would probably mean running a single reverse proxy in front of both of them, and mapping the two services onto subpaths of the same (protocol, port, domain) combination.

This doesn't really fit into our current setup, and making CORS headers configurable is easy, so I favor doing the latter, then looking for opportunities to do the former as things evolve in the future.

mattwigway:
Good points, and I don't see why just setting the access control allow
origin header to localhost:3000 or whatever wouldn't work. Most other
frontend frameworks use a reverse proxy for this, probably to avoid having
to alter backends, but it should work either way. I'll make a new PR, and a
new pair describing the full situation.

abyrd:
Superseded by #718

PR #684, opened 2021-02-15
Allow creating regional analyses with different destination grid extents

### LINKED ISSUE #676: Allow destination grids with different extents
R5 now has a PointSet implementation called GridTransformWrapper which can wrap a Grid opportunity dataset to show the same opportunity densities within any arbitrary web mercator grid at its same zoom level. Code exists in the workers to find the smallest grid containing the whole set of destination grids (checking that they're all the same zoom level) and then adapt them all to that one common super-grid using GridTransformWrapper instances in loadAndValidateDestinationPointSets(), which is called by the methods that handle both single point and regional tasks on the worker. 

This means the worker can already handle a mix of grids with different extents. But RegionalAnalysisController on the backend does a "Preflight check that all destination pointsets are exactly the same size" and refuses any analysis with grids of mixed sizes, so they never reach the worker.

On the worker, grid transform wrappers are only instantiated for destination grids that don't match the unified super-grid, so the preflight check effectively prevents creation of any GridTransformWrappers by blocking all jobs that would create them. This is a temporary precaution while we make sure the index transform math is correct - if it's not, it could easily fail silently returning plausible looking accessibility figures that were actually using densities from the wrong place, or just zeros since the transform wrapper supports reading cells outside the transformed opportunity grid.

We should add tests to GridTransformWrapper, then remove the preflight check on the backend. Besides automated tests, we should also manually test some cases where we run regional analyses on two wildly different destination pointsets, then run a single regional analysis with both destination pointsets, and ensure the results are identical with both approaches.

abyrd:
Fixes #676. After adding some tests on the GridTransformWrapper, this relaxes restrictions on the backend so analyses with multiple grids of varying extents can be created.

### DIFF
--- a/src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
+++ b/src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
@@ -429,7 +429,7 @@ public class RegionalAnalysisController implements HttpController {
                 } else {
                     checkArgument(
                         dataset.getWebMercatorExtents().zoom == opportunityDatasets.get(0).getWebMercatorExtents().zoom,
-                        "If multiple grids are specified as destinations, they must have identical extents."
+                        "If multiple grids are specified as destinations, they must have identical resolutions (web mercator zoom levels)."
                     );
                 }
             }
--- a/src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
+++ b/src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
@@ -417,8 +417,9 @@ public class RegionalAnalysisController implements HttpController {
             if (nPointSets == 1) {
                 task.grid = task.destinationPointSetKeys[0];
             }
-            // Preflight check that all destination pointsets are exactly the same size. Equivalent to worker-side
-            // checks in AnalysisWorkerTask.loadAndValidateDestinationPointSets and WebMercatorExtents.forPointsets.
+            // Check that we have either a single freeform pointset, or only gridded pointsets at indentical zooms.
+            // The worker will perform equivalent checks via the GridTransformWrapper constructor,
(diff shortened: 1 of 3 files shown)


src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
```diff
@@ -427,7 +428,7 @@ private RegionalAnalysis createRegionalAnalysis (Request req, Response res) thro
                     );
                 } else {
                     checkArgument(
-                        dataset.getWebMercatorExtents().equals(opportunityDatasets.get(0).getWebMercatorExtents()),
+                        dataset.getWebMercatorExtents().zoom == opportunityDatasets.get(0).getWebMercatorExtents().zoom,
                         "If multiple grids are specified as destinations, they must have identical extents."
```
ansoncfit, line 432:
```suggestion
                        "If multiple grids are specified as destinations, they must have identical resolutions (web mercator zoom levels)."
```

src/test/java/com/conveyal/r5/analyst/GridTransformWrapperTest.java
```diff
@@ -0,0 +1,86 @@
...
+    void testMismatchedZoomLevels () {
+        Grid grid = new Grid(10, 10, 10, 10, 10);
+        WebMercatorExtents webMercatorExtents = new WebMercatorExtents(10, 10, 10, 10, 11);
+        assertThrows(IllegalArgumentException.class, () -> new GridTransformWrapper(webMercatorExtents, grid));
+    }
+
+    /*
+     * TODO lat/lon based testing
+     * Given a set of points at latitudes and longitudes, write the same points into overlapping grids of different
+     * dimensions. Then transform all grids into a single super-grid, and make sure the opporunity counts at different
```
ansoncfit, line 81:
```suggestion
     * dimensions. Then transform all grids into a single super-grid, and make sure the opportunity counts at different
```

ansoncfit (approved):
Write-up of regional tests pending.

PR #686, opened 2021-02-16
Include header in output file length, fixes #671

### LINKED ISSUE #671: Index out of bounds in small grids
It's possible to get an `Index out of bounds error` on line78 of the following:

https://github.com/conveyal/r5/blob/d8dbcda9677ab0fe5271f6d767a69266331eaff4/src/main/java/com/conveyal/analysis/SelectingGridReducer.java#L70-L82

Create a regional analysis with bounds: 
```
bounds: { // This is taken from Cypress tests. Location is in Covington, Kentucky.
  north: 38.9577,
  east: -84.3969,
  south: 38.9570,
  west: -84.3966
}
```
This will generate region with 2 origins, with a height = 1, and a width = 2.

It's unlikely to occur in practice. But being able to generate this type of error in general may uncover the need for more bounds checking in this code.

abyrd:
When the output file containing only a header was resized based on the number of expected pixels and channels, the header bytes (9x int32) were not included in the requested length. This means the initial size of the file was too small by 9 integers. If the regional analysis completed, this had no effect on the regional analysis results themselves, because RandomAccessFiles allow seeking beyond the end of the file and will be automatically resized upon writing to such a position. When results were received for the last few origins, the file would grow to its correct length.

However, if the file was extremely small (less than 9 int32s) the initial resize operation would truncate the header. For example, 2 origins with 4 values each would require 8 int32s, and resizing the file without accounting for the header would chop off the last int32 in the header. Subsequent writes of accessibility results for all origins would then re-enlarge the file to its rightful size, but that last part of the header was lost. This is exactly what happened in #671.

At first I was baffled how we were reading results from files that were systematically too small without triggering exceptions, and wanted to add more safeguards against this. But in fact the files were the right size due to the seek+write behavior described in `RandomAccessFile.seek()` Javadoc. Because this bug had no effect at all on files of normal size (over 72 bytes long) I do not consider it a high priority to add a unit test for this exact situation.

I also removed GridResultAssembler which was a deprecated version of the GridResultWriter no longer in use.


trevorgerhardt (approved):
Glad that only affected small/incomplete files 👍

PR #688, opened 2021-02-17
Fix regional analysis with Grids of different extents

abyrd:
This is a fix to be applied on top of the already-merged #684. In addition to that backend change, it turns out we also needed a worker-side change to tolerate GridTransformWrappers in a method that used to only look for Grids. I also revised some comments explaining what's going on when we wrap multiple Grids to give them identical extents.

I ran a manual test against this PR where I took random (2D gaussian) points around Portland and split them into two halves, east and west, overlapping somewhat in the middle. These files were added to the analysis-e2e Fixtures directory alongside others used in manual tests. I ran three regional analyses: one with both east and west pointsets selected, one with only east, and one with only west. The analysis with both pointsets selected will require each pointset to undergo a different transformation to match the minimum bounding extents of the two taken together. The single pointset runs will not undergo any transformation, as the minimum bounding extents are equal to the extents of the single destination pointset.

Results from each of the single-pointset regional runs appear to exactly match those from the two-pointset run, corroborating the passing unit test on GridTransformWrapper. So this should complete the solution to issue #676.

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/WebMercatorExtents.java
+++ b/src/main/java/com/conveyal/r5/analyst/WebMercatorExtents.java
@@ -40,15 +40,23 @@ public class WebMercatorExtents {
         this.zoom = zoom;
     }
 
+    /**
+     * Return the analysis extents embedded in the task object itself.
+     * Sometimes these imply gridded origin points (non-Taui regional tasks without freeform origins specified) and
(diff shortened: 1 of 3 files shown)


ansoncfit (approved):
Thanks for the local tests. We'll merge now and test on staging once staging is restarted with this change in the backend.

Issue #566, opened 2019-11-28
Fractional opportunity quantization error


abyrd:
Our opportunity grids are saved using integers. The intent is to make them relatively compact, fast and easy to operate on within Javascript typed arrays.

However, internally in the backend and workers we allow fractional opportunities. These can especially arise when rasterizing polygons into an opportunity grid. We want to preserve the total number of opportunities within the polygon, and we want overlapping polygons to sum properly. Fractional opportunities are also convenient for certain types of advanced capacity-constrained analysis where opportunities are distributed fractionally to all individuals that can reach them.

Problems arise when converting these fractional counts to integers. You can either truncate or round them, and we do the latter. If you've got two overlapping polygons, the first one having 0.49 opportunities per cell and the second 0.51 per cell, the first one will just disappear while the second one remains. All the cells within a polygon will often round one way or the other in lockstep. The exception with overlapping polygons is even more weird: if the first one has 0.49 opportunities per cell and the second 1.01 per cell, the first will disappear, the second will round to 1, and the overlap will round to 2. 

This problem is most visible where cells have zero or one opportunity, but the effect exists for larger counts as well. It may seem minor since each cell is only off by one, but since they're all off in the same direction it can bias the total number of opportunities by up to half the the total number of cells.

This kind of quantization problem is thoroughly understood and solved in digital audio or image processing. The usual solution is to apply dithering. A simple approach of adding random dithering noise could work for us, with the downside that the allocation of polygon opportunities to grid cells then becomes non-deterministic, so accessibility results are not stable. It would also be strange to accept that the total number of opportunities would only probabilistically match the inputs. We would probably need a more sophisticated dithering approach that ensured the sum of opportunities from each polygon or in each neighborhood of a certain size post-dithering remained the same. This could get quite complicated. 

This is more or less the idea of error-diffusion dithering. As you iterate over the cells (either for a polygon or the whole grid) the remainder from each rounding operation is carried over to the next cell (or to several adjacent cells across and down). However this has a spatial/directional bias, and doesn't exactly preserve the total number of opportunities because error can "fall off" the right or bottom edge of the grid. The bias comes from the fact that these approaches operate in one pass, and avoid diffusing error backward to pixels that were already examined. There's probably an iterative convergence approach that diffuses in all directions. There must be research on this but it's buried in image processing literature that is not concerned with preserving the sum of the pixels.

[...]

PR #689, opened 2021-02-17
Grid error diffusion with tests

abyrd:
This should solve #566. Tests are intended to demonstrate preservation of opportunity count across the whole grid, and at several local spatial scales (adjacent pixels and rows).


src/main/java/com/conveyal/r5/analyst/Grid.java
```diff
@@ -310,11 +312,22 @@ public void write (OutputStream outputStream) throws IOException {
         out.writeInt(width);
         out.writeInt(height);
         // The rest of the file is 32-bit integers in row-major order (x changes faster than y), delta-coded.
-        for (int y = 0, prev = 0; y < height; y++) {
+        // Delta coding and error diffusion are reset on each row to avoid wrapping.
+        int prev = 0;
+        for (int y = 0; y < height; y++) {
+            // Reset error on each row to avoid diffusing to distant locations.
+            // An alternative is to use serpentine iteration or iterative diffusion.
```
ansoncfit, line 319: Serpentine iteration seemed appealing to me initially. But it's probably not worth the added complexity, and it's not clear that southward bias is preferable to eastward bias.

ansoncfit (approved):
Looks good.

PR #690, opened 2021-02-17
Preserve stop order in hops

### LINKED ISSUE #687: Hop stop order not guaranteed after feedscoping
In some modifications specifying hops (at least adjust-speed), the order of stops in each hop is occasionally reversed. The modification will then fail to apply to an intended hop without any warnings (unless all hops happen to fail so the whole modification has no effect).

For example, if the scenario model has `hops = [[stop1, stop2]]`, the scenario received by the worker sometimes has stops reversed: `[feedId:stop2, feedId:stop1]`.

See https://github.com/conveyal/r5/blob/dev/src/main/java/com/conveyal/analysis/models/AdjustSpeed.java#L40-L41

The feedscoping method collects to a set, which is not guaranteed to maintain order, at https://github.com/conveyal/r5/blob/dev/src/main/java/com/conveyal/analysis/models/Modification.java#L59

abyrd:
This should fix  #687.

Currently marked as draft PR, a few tasks remain: 

- [x] add warnings during scenario application based on number of patterns affected by each hop
- [x] test on some newly created modifications to make sure the conversions still happen smoothly

@ansoncfit indicated that the failing case described in #687 is nondeterministic (as expected) and hard to reproduce on production. So we will keep an eye on the new warning output from adjust speed modifications during our next round of manual testing.

### DIFF
--- a/src/main/java/com/conveyal/analysis/models/AdjustDwellTime.java
+++ b/src/main/java/com/conveyal/analysis/models/AdjustDwellTime.java
@@ -28,12 +28,12 @@ public class AdjustDwellTime extends Modification {
         adt.comment = name;
 
         if (trips == null) {
-            adt.routes = feedScopeIds(feed, routes);
+            adt.routes = feedScopedIdSet(feed, routes);
         } else {
-            adt.patterns = feedScopeIds(feed, trips);
+            adt.patterns = feedScopedIdSet(feed, trips);
         }
 
-        adt.stops = feedScopeIds(feed, stops);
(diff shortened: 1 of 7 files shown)


abyrd:
@ansoncfit if you get a chance maybe test this out on the modifications that were failing

abyrd:
This seems to be working well to me, but I'm unable to perform as many tests I wanted to do because the adjust-speed modification editor is behaving strangely in my local UI build. We will follow up with complete testing before the release. Merging now based on phone conversation earlier today, with the goal of creating a unified branch to serve as a release candidate.

PR #692, opened 2021-02-18
Add workflow to run cypress tests from the UI

trevorgerhardt:
Create a workflow file that runs Cypress Integration Tests from the `dev` branch of `analysis-ui`.


abyrd:
Java backend workflow takes 4:14 (most of which is unit tests) and Cypress integration workflow takes 16:14 (with unit tests disabled on its own backend build). Strangely some previous runs of the Cypress integration workflow were faster (15 minutes) when its backend build unit tests were still enabled.

.github/workflows/cypress-integration.yml
```diff
@@ -0,0 +1,47 @@
...
+
+      # Build .jar and copy to ./ui directory
+      - uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - run: gradle shadowJar -x test
+        working-directory: r5
+      - run: cp $(ls ./r5/build/libs/*-all.jar | head -n1) ./ui/latest.jar
+
+      # Install / cache dependencies with Cypress to handle caching Cypress binary.
```
abyrd, line 34: Which line below does this comment refer to? I'm trying to determine which Actions have caching behavior, to see if it's affecting Cypress test duration. https://github.com/conveyal/analysis-ui/blob/dev/.github/workflows/lint-jest-build.yml has a few `actions/cache@v2` steps with paths like `${{ github.workspace }}/.next`. Would something like that be beneficial here?

trevorgerhardt: I could make that comment a bit more verbose. The Cypress Github Action handles the dependency installation and caching, which isn't that hard to set up for a normal `node_modules` cache, but with the Cypress binary is a bit trickier.

abyrd: It could be helpful to update it, just to better understand what's being cached (not just Cypress itself and its dependencies, but also dependencies for the rest of the project?) I wasn't quite sure why there were separate cache steps in the analysis-ui workflow if the Cypress action takes care of caching.

abyrd (approved):
Could be helpful to clarify comments on caching behavior, but this is clearly already working so approving.

abyrd:
Oh right, it's going to auto-merge after my approval. Don't worry about updating that comment, but maybe post any clarification you have about what's being cached as a comment on this PR.

PR #700, opened 2021-02-26
Use "[new]" as stopId in path details for stops added in scenarios

ansoncfit:
Works on staging:
image


abyrd (approved):
Just discussed - we don't want to solve this by assigning a generated ID to each new stop, because we don't want to have to update this stop ID index independently for every scenario. So we just assume any stop that's not in the index is a scenario-added (new) one. This worker has been tested by @ansoncfit on staging, and it behaves as expected.

PR #701, opened 2021-02-26
Change initialization order for Taui site tasks

abyrd:
This fixes a null pointer reported by @ansoncfit. We were reading a field on the task before it was copied from the incoming analysisRequest.


ansoncfit (approved):
Looks good. This will now use the correct value of `task.makeTauiSite`, after it has been set from the `analysisRequest`, rather than the default `false`.

The failure was related to the block above populating destinationPointSetKeys conditional on `analysisRequest.makeTauiSite`:
```
// Set the destination PointSets, which are required for all non-Taui regional requests.
        if (! analysisRequest.makeTauiSite) {
            //
            for (int i = 0; i < nPointSets; i++) {
                //
                task.destinationPointSetKeys[i] = opportunityDataset.storageLocation();
            }
```

PR #702, opened 2021-02-27
Create a "Bucket" abstraction

trevorgerhardt:
This is worth a quick review before further work is done. I intentionally did not update all the locations that we use `FileStorage`, only a few places to show an example of what the difference would be.

A full implementation would include all of the `Bucket`s created in `Components`.

There appear to be instances where we still want to use `FileStorage` directly. But most cases we should be using a specific `Bucket`  and prevent each usage of the `FileStorage` to also be passed, and save an additional configuration `String bucket`.


trevorgerhardt:
Closing in favor of #664

Issue #706, opened 2021-03-02
User "status" endpoint


trevorgerhardt:
PR https://github.com/conveyal/analysis-ui/pull/1492 establishes a poll to the API at a regular interval to ensure connectivity to the API. At the moment, it hits the simple `/version` endpoint. 

But with a regular poll occurring, this could be used to return user/access group "status" and progress updates and also log user activity in the system. We should instead create a specific HTTP endpoint and related controller to use instead of the `/version`. (Maybe `/api/status`?) The first implementation should be very simple, with the main goal of returning a `200` response to the UI to indicate the user is logged in correctly and the API is available. Possibly log the user activity for assistance with deployments, but eventually we could return progress updates across the application.

Related to #461 #444

PR #707, opened 2021-03-03
Add heartbeat endpoint for per-user status (#706)

abyrd:
This adds an endpoint at `/api/activity` endpoint where the UI can signal that a user is active, and receive updates on any background tasks that user has running. This currently reports only OSM/GTFS upload progress. Task tracking will be cleaned up and expanded to report any OSM/GTFS errors encountered.


trevorgerhardt:
Nice. This looks great. My first thought is wondering if we should be getting tasks just for the user, for the entire access group, or both?

trevorgerhardt (approved):

Issue #478, opened 2018-12-05
Disallow selection by patterns


ansoncfit:
We've agreed to explicitly specify all trips for modifications.  There may be R5 code to allow selection by patterns.  If we don't plan to use/test it, we should make it throw an `UnsupportedOperationException`.

abyrd:
We discussed this at today's meeting: it seems like we didn't make the details clear enough when we made the decision referenced in this ticket, as it contradicts another decision in #144 that is a basic element of how the backend and workers work.

There are several places where trip IDs are used to specify patterns and implementation may be different in different places. What is uniform across the stack (frontend to backend to worker) is that we avoid storing or transmitting pattern IDs to identify patterns, as these are not present in the GTFS itself and are derived when we load the GTFS, so could be created in nondeterministic ways. Instead we use trip IDs to signify the pattern that contains them.

The "selection by patterns" disallowed in this ticket probably means selecting patterns _directly by their pattern IDs_. We clearly decided not to do that. However, it is still common and necessary to reference patterns somehow, for example in modifications. It's just that we reference the patterns using one or more "exemplar" trips from that pattern, not an ID for the pattern itself. A modification may have effects on an entire pattern at once, but we use trip IDs to select patterns instead of pattern IDs. Each trip belongs to exactly one pattern, so a trip unambiguously identifies a pattern. There may be hundreds or thousands of other trips (and trip IDs) that identify the same pattern, and specifying more than one of them just repeatedly/redundantly specifies the same pattern as far as the workers are concerned. 

Specifying the full set of trips on a pattern does not somehow more securely specify that pattern, as the workers do not look for a pattern with _all_ those trips present, they select patterns with _any_ of the listed trips present. This is by intentional design, as it allows sending one ID instead of hundreds, an could enable some faster and simpler ways find the pattern associated with a trip. There's also no real need to validate/ensure that the selected pattern contains a full exact set of trips, as the pattern is known to be drawn from the exact same feed and the trip-to-pattern relationships are fixed.

In many cases we actually _do_ send every trip on a pattern, or store them all in the modification in the database, but the workers do not need this information. I believe this ticket was documenting a decision that we would not send pattern IDs and/or that it was acceptable to continue sending large amounts of trip IDs from the same pattern where this was the status quo, and the workers should continue accepting this. The concerned modification types check against sets of trip IDs instead of lists to make this status quo approach more efficient.

[...]

Issue #256, opened 2017-03-17
Refactor / replace active pattern filtering


abyrd:
In FastRaptorWorker we filter the patterns to separate them into frequency and schedule patterns, creating new lists that contain only the active patterns.

This requires six fields including some lookup tables to convert list indexes. I think these could be collapsed down to just two bitsets for active frequency and active scheduled patterns (not doing any mapping, just pre-filtering and using the original indexes). Maybe I'm missing something though. I'd also be curious to see the speed difference with no filtering at all, just to see if the complexity is worth the speedup.

This is an optimization, and one that requires maintaining index lookup tables and copying lists - do we have evidence that it provides any benefit over not filtering? Perhaps there isn't much advantage over simply making a bitset of active patterns and iterating over the set bits in that bitset to iterate over active patterns?

The original though behind the filtering was that it might get the patterns into the CPU cache, but since they're references and they probably won't all be on the same page.

ansoncfit:
Closed by #708.

Issue #144, opened 2016-05-02
Modifications: change trips field to patterns


abyrd:
We decided that instead of exhaustively listing out all affected trips, we would instead have a "patterns" field. This is a list of trip IDs. All trips on any pattern containing any of the listed trip IDs will be modified. This avoids problems where only some of the trips on a pattern are modified, creating a situation where new patterns would need to be created. It will only be possible to select entire patterns or routes until we have a need to do otherwise, at which time we may re-introduce the trips field.

abyrd:
In the end, I changed "trips" to "patterns" for all modifications where it would be disastrous to affect only some trips on a pattern, and for all others I added a patterns field so they accept any one of routes, patterns, or trips. Here's a summary:

AddTrips: Only documentation was updated.
AdjustDwellTime: Added patterns field. Now accepts routes, patterns, or trips.
AdjustFrequency remains unchanged.
AdjustSpeed: Added patterns field. Now accepts routes, patterns, or trips.
RemoveStops: Added patterns field. Now accepts both routes and patterns (not trips).
RemoveTrips: Added patterns field. Now accepts routes, patterns, or trips.
Reroute: Added patterns field. Now accepts both routes and patterns (not trips).

abyrd:
This issue already contains a description of how it was resolved over five years ago. However, we still don't make efficient use of this patterns field (we often send all the trip IDs instead of just one). This seems like low hanging fruit to improve database access latency over the network and generally lower bandwidth requirements. For now I'll leave this open until we make a conscious decision to use or not use this feature in requests from the front end.

abyrd:
It seems like there's still potential to make bandwidth and storage constant-size (instead of variable and often large) and make some code simpler and more maintainable by storing/sending individual trip IDs to reference patterns instead of full lists of every trip on the pattern. This would eliminate any misunderstandings or assumptions over whether all trips must be present to select a pattern, whether their order matters, what to do when some trips are present but not others etc. It also simplifies looking up patterns from trips, which may be significant: in some places comparing the full lists seems to require linear scans over every trip in a GTFS feed and may have quadratic or cubic complexity.

As far as I know switching to single trip IDs requires no changes on the backend/workers. There may have been some misunderstanding before that this requires subtle or difficult work, but I think it just requires a few lines changed here and there, albeit across several projects so it requires careful coordination.

Issue #20, opened 2015-11-13
Filtering trips and patterns unexpectedly changes results


abyrd:
Given a search day, com.conveyal.r5.analyst.scenario.InactiveTripsFilter removes all trips and patterns that are not running on that day during the search time window. It should have no effect on the analysis results, but we see a small change in areas farther away from the origin point: travel time increases by 2-5 minutes when the filter is used.

This implies that the basic search is making use of some routes that the filter is removing. Since none of these routes should be active, this implies an oversight in the basic search process.

One consideration is that searches may continue up to two hours past the end of the departure time window, but that is taken into account when filtering the trips. Even deactivating the time window filtering and full-pattern filtering (filtering only individual trips based on their service ID) the discrepancy is still visible.

abyrd:
Interestingly the filtered network is faster in some places (the problem causes bidirectional changes in travel time). I am now considering whether removing trips or patterns is changing trip indexes and making some index invalid, and that index is not being rebuilt.

abyrd:
This issue is over six years old, and recent work has overhauled the trip and pattern filtering logic to fix all known problems.

PR #708, opened 2021-03-04
Trip and pattern filtering

ansoncfit:
We have relied on the assumption that no trip overtakes another trip on the same pattern.

Recently, a regional analysis with path recording failed on one task because negative transfer times were being calculated. Closer inspection uncovered negative transfer times and other path failures related to overtaking. 

[Specifically, the September 2019 MBTA GTFS implies overtaking on Route 22: the 6:35 AM departure from Ashmont overtakes the 6:33 departure from Ashmont around Talbot Ave. @ Spencer St. For stops down the line, the correct trip would potentially not be boarded because of this upstream overtaking.]

This PR:
- Checks for negative transfer times when requesting single-point path results for inspection in the browser console (https://github.com/conveyal/r5/commit/caa5db1a9ad19a91273034477795354ec4ed605b). Previously, this check was only in place for regional analyses of paths
- Refactors how trips are filtered in the scheduled search, perhaps a marginal performance gain (https://github.com/conveyal/r5/commit/607bc16ea1a29fbf2247a521d8196d6ed15e73de)
- Checks the order of departure times at a boarding stop before breaking out of iteration, relaxing the assumption about overtaking for scheduled trips. This may reduce performance (https://github.com/conveyal/r5/commit/607bc16ea1a29fbf2247a521d8196d6ed15e73de)

A few comments: 
- The `0` initial value `boardTime` was apparently not used, so changing this to `Integer.MAX_VALUE` to help with one of the conditions should be safe, but it's worth double checking this.
- We should discuss whether to handle overtaking for frequency-based trips
- We should not spend time on handling overtaking downstream of boarding stops -- for a given boarding stop and pattern, passenger behavior is arguably to board the earliest feasible trip, rather than looking ahead at the schedule for later departures that arrive earlier at downstream stops.


ansoncfit:
Previously, patterns were being filtered at the start of each `FastRaptorWorker.route()` call; this required a loop over all patterns at the start of each transit routing task. Furthermore, trips on patterns were not filtered, so while iterating over patterns within routing, there were multiple statements similar to:
```
if (!servicesActive.get(candidateSchedule.serviceCode) || candidateSchedule.headwaySeconds != null) {
  continue;
}
```

This PR has been updated with changes to the pre-filtering by analysis date and transit modes:
- Trips are filtered by date, with the selected ones stored in separate lists for schedule-based and frequency-based trips. This change and a pre-check for overtaking allows for the elimination of the `continue` blocks mentioned above, and better handling of overtaking scheduled trips (the original motivation)
- These trips, along with details of their parent patterns relevant for routing, are cached with the transit layer, eliminating the filtering loop over patterns when successive requests have the same date and transit modes. Testing with a small regional analysis of MBTA GTFS (which tends to have many different trips on a pattern running for different days), the revised pre-filtering saves about 30 milliseconds per origin on average (substantial computational savings on large regional analyses).

ansoncfit:
Commits since the previous comment fix a small oversight (that could lead the first trip in each pattern to be ignored in the reboarding checks), add a unit test for overtaking, and clarify how enabling range raptor may change results (due to look-ahead behavior related to overtaking).

src/main/java/com/conveyal/r5/analyst/cluster/PathResult.java
```diff
@@ -155,13 +155,15 @@ public void setTarget(int targetIndex, Multimap<PatternSequence, Iteration> patt
             this.egress = pathTemplate.stopSequence.egress == null ? null : pathTemplate.stopSequence.egress.toString();
             this.transitLegs = pathTemplate.transitLegs(transitLayer);
             this.iterations = iterations.stream().map(HumanReadableIteration::new).collect(Collectors.toList());
+            iterations.forEach(pathTemplate.stopSequence::transferTime); // Sense check that travel time components
+            // do not exceed total travel time TODO report transfer times in single-point responses?
```
trevorgerhardt, line 159: Was this comment auto-formatted? I'm a bit confused by it.

src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
```diff
@@ -239,25 +231,21 @@ private void dumpAllTimesToFile(int[][] arrivalTimesAtStopsPerIteration, int max
...
-                // At least one trip on this pattern is relevant, based on the profile request's date and modes.
-                if (pattern.hasFrequencies) {
-                    runningFrequencyPatterns.set(patternIndex);
-                }
-                // Schedule case is not an "else" clause because we support patterns with both frequency and schedule.
-                if (pattern.hasSchedules) {
-                    runningScheduledPatterns.set(patternIndex);
-                }
-            }
+    private void filterPatternsAndTripsIfNeeded() {
```
trevorgerhardt, line 238: Would this method be more appropriately part of the TransitLayer? 

`transit.filterPatternsAndTripsIfNeeded(request.transitModes, servicesActive);`

abyrd: It does make sense to put the method on the TransitLayer since it's modifying data within that class. However, the TransitLayer is shared between all threads, so must either be treated as read-only or any modifications must be synchronized. This modification of a single set of filtered patterns is stateful, and the test-then-filter action here, as well as their subsequent use in routing, appears to create a race condition. Specifically, another thread can change the set of filtered patterns while they're still being used, or even while or before they're filtered.

It's not necessarily a problem to concurrently read and write a single reference field serving as a cache, because reference reads are atomic (Java prevents "read tearing" on references). But the reading thread would need to hold on to its own copy of the filtered patterns for a consistent view.

abyrd (commented):
I have come back to working on this over the last few days. In addition to a few comments inline in the code, I'll summarize our last round of discussion and update my reactions: 

Cacheing filtered patterns
- TransitLayer is roughly the right place to store filtered patterns, however as a data structure shared accross all threads TransitLayer is arguably the wrong scope.
- Wherever they're stored, filtered pattern cacheing would need to be revised for thread safety.
- A more appropriate location may be the TransportNetworkCache, which already provides thread-aware handling of various scenarios built on top of the base network - seems reasonable to treat filtered patterns as an additional stage here: Base network -> scenario network -> filtered network.
- I will try moving the cacheing over to TransportNetworkCache.

Departure Search
- This PR correctly disables the early exit from that search. We can't break out of the departure search where trips in a pattern overtake. 
- However, to handle all cases we need to search the entire list of trips in the pattern, not just ones with lower indexes. We should treat them as totally unsorted and scan over the whole list of trips.
- In a way this makes the re-boarding departure search with overtaking more similar to the initial-boarding departure search. There may be an opportunity to merge the four distinct cases (initial vs. reboarding, sorted vs. unsorted) but it will be easier to achieve correctness by revising the code incrementally, keeping the various conditional clauses at first.
- I will revise the departure search to handle all four cases.

I had some intuition that we should mark overtaking patterns when we build the network, then update that flag on any pattern affected by a scenario or by trip filtering. Thinking through that intuition and why we may want to ignore it, as well as how this could be efficiently implemented, here are my observations:
- Early breakout of departure searches is an optimization. The code works fine if we conservatively assume any pattern can contain an overtaking pattern, then always scan every trip for re-boarding at every stop. But this is very slow: organizing trips into sorted patterns is key to the speed of the raptor algorithm.
- Therefore, false positives when detecting overtaking patterns are harmless for correctness. They only incur a speed penalty proportionate to the number of false positives.
- Applying a scenario can toggle a pattern's overtaking status - overtaking patterns can become non-overtaking and vice versa. Filtering trips out of a pattern based on service day and mode can only make overtaking patterns non-overtaking, but not the reverse.
- Marking patterns as overtaking or not involves looking at every stop on every trip in the pattern. This is a potentially time consuming operation we don't want to apply unnecessarily.
- Unfiltered patterns in the baseline network could be pre-marked as overtaking or not, and upon filtering we'd only have to re-evaluate the overtaking status of patterns changed by filtering which contained overtaking in the baseline network or were affected by the scenario. The amortized effectiveness of such an optimization depends on how many baseline patterns have overtaking, how many are typically filtered or modified in a given request and scenario, and whether that proportion varies significantly from one network to another.
- Tracking and revising the overtaking status of patterns adds some complexity, but this could probably be done by repeatedly calling the same method, not adding much new code. 
- However, cacheing the filtered result with overtaking detected means we incur that cost only once for a particular scenario, date and time, and reuse the result many times thereafter. So measured over many requests, the advantage of cascading overtaking status from baseline to scenario network to filtered patterns may be very small.

Summarizing the situation (and I think this reflects our last discussion on the subject), we have three independent changes here: 
1. changing re-boarding searches to scan all trips in case there is overtaking
2. flagging overtaking patterns, so that departure scanning can be conditionally optimized
3. filtering patterns before flagging them, because overtaking patterns are often non-overtaking on any one day.

Those three changes could be developed separately and sequentially, but are combined in the current PR. It contains a cacheing mechanism, so seems to follow the reasoning above, keeping code simple and always incurring the full cost of filtering and overtaking detection because the amortized cost over large numbers of repeated requests is low. We then gain something by implementing all three of these changes at once, avoiding writing the intermediate code to cascade overtaking status from baseline to sceanrio to filtered pattern.

## Reformulate the problem and transform the data
One could argue that the complexity here is just a result of ill-adapted data structures mixing trips from different days. A different approach is to change the definition (unique key) of Pattern such that all trips in a pattern are active on the same set of service days. I believe it's always algorithmically valid to split patterns on additional criteria. Given realistic schedule data all patterns should then be non-overtaking. We'd still get hit with the occasional bad input data, which we could either drop or handle. If we decided to gracefully handle such schedules and flag the patterns that contain overtaking, we know that overtaking status will not change when filtering for date, because only entire patterns will be dropped. The effectiveness of this approach will depend on how many patterns exist across different input data containing trips active on intersecting sets of two or more service days. I'd guess this is rare, but we could log it and observe, or pre-analyze a sample of input data. 

## Filtering trips on both date and time
If we do continue filtering patterns at the individual trip level, we could filter on time window as well as date. In typical usage this would remove the majority of the trips, probably speeding up the search or at least reducing the memory footprint of filtered patterns. However those extra trips probably have little impact on usual non-overtaking operation where we only perform short partial scans of the trips in a pattern. The impact of large trip lists appears only when we handle overtaking by doing full linear searches over all trips in a pattern.

## Tests
It's great to have added tests to demonstrate the original problem and verify that the PR resolves it. Some extensions are made to the gridded network testing framework to enable this particular test. This framework may not be ideally suited for testing behavior on a single pattern, and it might make the test grid construction API more confusing in the long run if we introduce extensions that see limited reuse. My sense is that we should either go in and directly alter the timetables after they're generated for this specific test case, or somehow simplify or generalize the way routes are specified. I'm happy to refactor that now or later, and add some more documentation to that test framework. The new tests are directly examining paths and comparing them to literal arrays of expected travel times. I have a patch reworking them to actually test the distributions using the Distribution mechanism. I should add some documentation to clarify how the distribution builder is meant to be used.

I like the use of setDestination on the task in this PR. The previous grid test pattern retains travel time histograms at every destination which uses a lot more memory and time. I might optimize the framework to retain travel time histograms only when destinations are small freeform pointsets or individual points, and interconnect with your setDestination method.

abyrd:
Next steps: I can proceed with the revisions described in my last review, but we should first discuss the alternative described there of redefining trip patterns to contain only trips running on the same combination of service codes. That would be quite different, but could turn out to be much simpler and might work as well or better, depending on what typical input data looks like.

ansoncfit:
Related issues:
- https://github.com/conveyal/r5/issues/144 and https://github.com/conveyal/r5/issues/478, which suggest moving in opposite directions
- https://github.com/conveyal/r5/issues/20
- https://github.com/conveyal/r5/issues/256

src/test/java/com/conveyal/r5/analyst/network/SimpsonDesertTests.java
```diff
@@ -149,12 +137,81 @@ public void testGridFrequencyAlternatives () throws Exception {
...
+     * ahead of Trip B until stop 20, is passed around stop 21, and runs 9 minutes behind Trip B for the remaining
+     * stops. This example is tested with 3 cases for a rider traveling to stop 42 (well downstream of the overtaking
+     * point):
+     *
+     * 1. Standard rider, departing stop 30 between 7:00 and 7:05, who always rides Trip B (the earliest feasible
+     * departure from that stop)
+     * 2. Naive rider, departing stop 10 between 7:00 and 7:05, who always rides Trip A (the earliest feasible
+     * departure from that stop)
+     * 3. Savvy rider, departing stop 10 between 7:13 and 7:18, who always rides Trip B (avoiding boarding the
+     * slower Trip A thanks to the "look-ahead" abilities when ENABLE_OPTIMIZATION_RANGE_RAPTOR is true)
```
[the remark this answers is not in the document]
abyrd: Discussed with @ansoncfit, it does appear that in the presence of overtaking, results will be changed by ENABLE_OPTIMIZATION_RANGE_RAPTOR=true. This PR gets "reasonable" results for overtaking trips in the sense that a passenger will discover and ride the first trip departing from any stop, but range-raptor will actually find the trip that allows them to alight earliest. An incumbent later-departing, earlier-arriving trip found at a later departure minute wins out over an earlier-departing, later-arriving trip. In other situations range raptor should remain a pure optimization, so we just need to document this and make sure any test for correctness of optimizations isn't applied to schedules with overtaking trips.

abyrd:
@ansoncfit I just added a few more commits factoring out the departure search methods as I mentioned in Slack. The R5 tests passed locally, including the overtaking tests. Please take a look at the new logic in FastRaptorWorker#doScheduledSearchForRound and tell me if it looks correct to you, considering your overtaking changes.

src/test/java/com/conveyal/r5/analyst/network/GridGtfsGenerator.java
```diff
@@ -34,21 +34,21 @@
 
     private final boolean mergeStops;
 
-    public GridGtfsGenerator (GridLayout gridLayout) {
+    public GridGtfsGenerator(GridLayout gridLayout) {
```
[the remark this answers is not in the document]
ansoncfit: Updated IDE settings:

image

Then, Code → Reformat Code

abyrd:
Fixed a problem with caching filtered scenarios. The only Cypress tests failing now are the ones related to the changes in feed naming in the activity-api branch.

ansoncfit (commented):
I used `v6.2-32-g329da6c` for a regional analysis with 130,735 origins (see https://analysis.conveyal.com/regions/a58996ff-a377-4277-9201-43b355817cd7/regional)

As expected, increased access is apparent around Route 22, which has overtaking as described above.
image

In terms of total computation time with auto-scaling, v6.2 took ~1.07 ms per origin and this branch took ~1.03 ms per origin. This relative 3.7% savings should be larger for network bundles with more complex service/pattern combinations, and the absolute savings for analyses with more origins could be substantial.

After that testing, I merged the latest dev and pushed a no-op commit to remove unused variables.

I now consider this approved for merge (though as the user who originally opened this PR, I cannot submit a formal Approve review).

abyrd (approved):

PR #711, opened 2021-03-18
Adjustable zoom level for grids

ansoncfit:


src/main/java/com/conveyal/analysis/controllers/OpportunityDatasetController.java
```diff
@@ -148,6 +148,8 @@ private OpportunityDatasetUploadStatus downloadLODES(Request req, Response res)
         }
 
         final String regionId = req.params("regionId");
+        final int zoom = req.params("zoom") != null ? Integer.parseInt(req.params("zoom")) : ZOOM;
```
ansoncfit, line 151: @trevorgerhardt what do you think about this approach? 

Zoom level should be adjustable for:
* analysis (previously implemented)
* LODES import (here)
* spatial dataset and aggregation area upload (soon)
* prebuilt and saved linkages/egress cost tables (longer-term)

One approach would be setting the zoom once at the region level, then reading it from the region model object in the database. 

Alternatively, the front-end could include a zoom selector for each of the uses above (maybe defaulting to a value set at the region level, but allowing customization).

ansoncfit:
Notes from today's discussion:

Let's proceed with adding an adjustable zoom parameter to the spatial dataset creation. Should we jump straight to our long-planned change of keeping the source files (e.g. .shp) available and allowing conversion to grids on-demand, or just force re-upload? Probably the latter as a starting point.

Note that `zoom` and `extents` apply to the destination grid for single-point analyses, but to the origin grid for regional analyses. We need to document that more clearly.

Rather than trying to do downstream filtering (e.g. so that the destination layer selector only show grids that match the zoom level of the analysis task), we should just show errors when resolutions do not match.

The automatic upsampling changes in #712 are mostly useful for using destination grids with different zoom levels in the same regional analysis.

src/main/java/com/conveyal/analysis/controllers/AggregationAreaController.java
```diff
@@ -132,6 +132,8 @@ private FileStorageKey getStoragePath (AggregationArea area) {
 
         boolean unionRequested = Boolean.parseBoolean(query.get("union").get(0).getString());
 
+        int zoom = req.attribute("zoom") != null ? Integer.parseInt(req.attribute("zoom")) : DEFAULT_ZOOM;
```
trevorgerhardt, line 135: Just realized you're using `req.attribute` here. It should probably be sent via the request body, so here you would be able to get it via `Integer.parseInt(query.get("zoom").get(0).getString())`

src/main/java/com/conveyal/analysis/controllers/OpportunityDatasetController.java
```diff
@@ -148,6 +148,8 @@ private OpportunityDatasetUploadStatus downloadLODES(Request req, Response res)
         }
 
         final String regionId = req.params("regionId");
+        final int zoom = req.params("zoom") != null ? Integer.parseInt(req.params("zoom")) : DEFAULT_ZOOM;
```
trevorgerhardt, line 151: Here we probably don't want to use `req.params` either, otherwise it would have to be part of the URL. We could put as a query parameter fairly easily. So `req.queryParams("zoom")` would work here.

src/main/java/com/conveyal/analysis/controllers/OpportunityDatasetController.java
```diff
@@ -440,6 +442,7 @@ private void verifyBaseNamesSame (List<FileItem> fileItems) {
     private OpportunityDatasetUploadStatus createOpportunityDataset(Request req, Response res) {
         final String accessGroup = req.attribute("accessGroup");
         final String email = req.attribute("email");
+        final int zoom = req.attribute("zoom") != null ? Integer.parseInt(req.attribute("zoom")) : DEFAULT_ZOOM;
```
trevorgerhardt, line 445: Same as above, we can use the helper `getFormField` here. So 
```java
Integer.parseInt(getFormField(formFields, "zoom", false))
```

trevorgerhardt (changes requested):
I didn't notice this before, but it looks like we're using the wrong `req.` methods to get the zoom level. I added some comments inline suggesting the appropriate ones to use for the way that endpoint is used.

trevorgerhardt (approved):
Looks good to me 👍

Issue #606, opened 2020-10-08
Adjustable web mercator zoom level for opportunity datasets


ansoncfit:
As of #586, the resolution of isochrones can be adjusted with a `zoom` parameter in the request. As the discussion there notes, the tricky part will be allowing adjustable zoom levels for the opportunity datasets.

Our initial approach will likely be having users specify a zoom at the region level, and using that as the default when rasterizing opportunity datasets and running analyses. We could check at region creation that the number of pixels is reasonable.

We should also address #566 before enabling.

ansoncfit:
We should address this as part of #732. When converting a previously uploaded and validated resource to grid layers, users should specify a desired zoom level.

PR #712, opened 2021-03-19
Upsample opportunity grids to higher zoom levels

abyrd:
This PR updates the GridTransformWrapper to allow transforming grids to higher zoom levels. Downsampling (aggregation) is also possible, but it's not clear whether we'd want to do that in our ideal workflow (we might always resample to the highest resolution in use). Because this kind of rescaling is so computationally cheap we should seriously consider the possibilities over the long term. Related to #606 (adjustable zoom levels). 

Similar transforms are possible in regional results display (when comparing regional results within the UI) or anywhere else we use grids. Unlike counts, accessibility indicators do not "dilute" across smaller cells so may be even simpler.

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/GridTransformWrapper.java
+++ b/src/main/java/com/conveyal/r5/analyst/GridTransformWrapper.java
@@ -3,6 +3,7 @@ package com.conveyal.r5.analyst;
 import org.locationtech.jts.geom.Envelope;
 
 import static com.google.common.base.Preconditions.checkArgument;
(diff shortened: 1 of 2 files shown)


abyrd:
This PR served as a proof of concept and an example of dynamically shifting between zoom levels. It prompted a conversation about how we will address zoom level changes. We now plan to manually or automatically generate grids at each different zoom level needed rather than even reusing a low resolution grid at higher resolutions. Instead of keeping this logic, which is a little more complicated and harder to follow, we'll require zoom levels to match which tightens requirements and should be safer (fail-fast).

Issue #698, opened 2021-02-25
Corrupt OSM fails silently


abyrd:
If you create a new bundle in the UI, choosing or uploading a good GTFS file but a bad OSM file, the UI eventually shows the newly created bundle without the ERROR indicator that we usually see when uploading bad GTFS. The upload fails silently until you try to use the OSM in an analysis.

This is apparently because in `OSM.readPbf()`, osm-lib doesn't throw exceptions, it logs them and stops quietly. I just reproduced this problem, uploading a GTFS zip file that had been renamed to `.osm.pbf`. The frontend shows no error at all on upload, it appears that a new bundle has been successfully created. But the backend logs show: 

```
08:50:31.942 [pool-3-thread-1] INFO  com.conveyal.osmlib.OSM - OSM will be stored in a temporary file.
08:50:32.242 [pool-3-thread-1] INFO  com.conveyal.osmlib.OSM - No OSM tables exist yet, they will be created.
08:50:32.250 [pool-3-thread-1] ERROR com.conveyal.osmlib.OSM - Error occurred while parsing VEX stream.
org.openstreetmap.osmosis.osmbinary.file.FileFormatException: Unexpectedly long header 65536 bytes. Possibly corrupt file.
        at org.openstreetmap.osmosis.osmbinary.file.FileBlockHead.readHead(FileBlockHead.java:50)
        at org.openstreetmap.osmosis.osmbinary.file.FileBlock.process(FileBlock.java:130)
        at org.openstreetmap.osmosis.osmbinary.file.BlockInputStream.process(BlockInputStream.java:34)
        at com.conveyal.osmlib.PBFInput.copyTo(PBFInput.java:237)
        at com.conveyal.osmlib.OSM.readPbf(OSM.java:247)
        at com.conveyal.analysis.controllers.BundleController.lambda$create$1(BundleController.java:156)
        at com.conveyal.analysis.components.TaskScheduler$ErrorTrap.run(TaskScheduler.java:95)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
        at java.base/java.lang.Thread.run(Thread.java:829)
```

[...]

Issue #624, opened 2017-02-13
Report invalid GTFS to UI


abyrd:
I am uploading a ZIP of 18 GTFS feeds and OSM data. The upload never seems to complete (based on the UI spinner). Looking at the server logs, I see that the GTFS feed is missing all required tables.

Oddly, it also reports that the feed does not contain a feed ID and it's using the file name 1b376653-c553-4d28-a752-3753d607952c. This ZIP file does not contain any file by that name, all GTFS files have been given the names of their respective agencies.

```
06:50:48.916 [Thread-16] INFO  c.conveyal.gtfs.model.Entity$Loader - Table feed_info was missing but it is not required.
06:50:48.916 [Thread-16] INFO  com.conveyal.gtfs.GTFSFeed - Feed ID is undefined, pester maintainers to include a feed ID. Using file name 1b376653-c553-4d28-a752-3753d607952c.
06:50:48.916 [Thread-16] INFO  c.conveyal.gtfs.model.Entity$Loader - Table calendar_dates was missing but it is not required.
06:50:48.916 [Thread-16] INFO  c.conveyal.gtfs.model.Entity$Loader - Table fare_attributes was missing but it is not required.
06:50:48.916 [Thread-16] INFO  c.conveyal.gtfs.model.Entity$Loader - Table fare_rules was missing but it is not required.
06:50:48.916 [Thread-16] INFO  c.conveyal.gtfs.model.Entity$Loader - Table shapes was missing but it is not required.
06:50:48.917 [Thread-16] INFO  c.conveyal.gtfs.model.Entity$Loader - Table transfers was missing but it is not required.
06:50:48.917 [Thread-16] INFO  c.conveyal.gtfs.model.Entity$Loader - Table frequencies was missing but it is not required.
06:50:48.917 [Thread-16] INFO  com.conveyal.gtfs.GTFSFeed - 6 errors
06:50:48.917 [Thread-16] INFO  com.conveyal.gtfs.GTFSFeed - GTFSError: agency line 0: This table is required by the GTFS specification but is missing.
06:50:48.917 [Thread-16] INFO  com.conveyal.gtfs.GTFSFeed - GTFSError: calendar line 0: This table is required by the GTFS specification but is missing.
06:50:48.917 [Thread-16] INFO  com.conveyal.gtfs.GTFSFeed - GTFSError: routes line 0: This table is required by the GTFS specification but is missing.
[...]

Issue #608, opened 2020-10-07
Mark bundle state ERROR when OSM loading fails


abyrd:
Uploading bad OSM fails silently until you try to use the OSM in an analysis. If I upload a bad OSM file with a good GTFS file, the UI eventually shows the newly created bundle as if there were no problems, without the ERROR suffix we see on GTFS load failures. You don't realize anything's wrong until you kick off a single point analysis, you then get "Error communicating with the server" with a NullPointerException at TransportNetworkCache.java:111.

This is because in OSM.readPbf(), OsmLib doesn't throw exceptions, it logs them and exits cleanly.

This OSM processing is being wrapped in an asynchronous task which has its own exception handling and can report errors. The exception is hidden and cascades another error that is successfully caught by the background task's ErrorTrap. But the original exception it should be caught farther down, in the heavyTask's catch clause.

Related to conveyal/analysis-backend#15. For both OSM and GTFS we should try to expose more detailed errors to the UI.

ansoncfit:
https://github.com/conveyal/r5/pull/713 closes.

Issue #570, opened 2020-01-07
Better error reporting during network build


abyrd:
Many errors can happen during network build which are logged to the console. For example see the errors in `com.conveyal.r5.transit.TripSchedule#create`. This is left over from early days of R5 use as a point to point trip planner, and is not adequate for use as a library in a modular/distributed system. Errors and warnings should have an object representation, be accumulated, and be reported to the caller in a structured way. If the build does recover from some problems, it should still accumulate warnings and return them. There should be a path for these to bubble up from  analysis workers to a web UI.

PR #713, opened 2021-03-19
Propagate errors from GTFS/OSM/other caches out to UI

abyrd:
This ensures that serious errors from uploading bad GTFS or OSM, as well as any errors that occur when building or loading transport networks on workers, are all revealed to the UI. Problems during bundle creation are recorded in the bundle metadata in Mongo. Problems on the worker are reported via scenarioApplicationWarnings. Addresses #570 #608 #624 #698.

Gallery of new errors:

Bad OSM:
Screen Shot 2021-03-20 at 00 24 54

Missing GTFS:
Screen Shot 2021-03-20 at 00 11 56

Invalid GTFS route type (fails on worker):
Screen Shot 2021-03-20 at 00 05 16

GraphQL chain of causes:
GraphQL


abyrd:
The failing test seems to be the probabilistic Simpson Desert test again. I can't get it to fail locally, maybe I can make it deterministic by seeding the random number generators, or just increase the number of MC draws.

abyrd:
@trevorgerhardt asked whether these changes might cause unforeseen problems on workers - we may have been swallowing errors on workers in order to prevent them from crashing and stalling. 

To my knowledge we were not hiding exceptions to avoid crashes. Most or all of the exception-catching with null return values was an intentional (but in retrospect misguided) aspect of API design when gtfs-lib and osm-lib were being drafted as libraries. It turns out it's much more useful for these libraries to signal _why_ they failed, via an exception or some compound return type, but that's not how we originally did it. 

So now that exceptions can escape can they crash anything? All uses should be within either the handler threads of a web framework, or in runnables managed by an asynchronous ExecutorService. Our web framework definitely catches exceptions and errors and transforms them into HTTP responses so I don't have much concern there. The potential problem is with async background tasks. If a Throwable (not just an Exception, but also an Error and a few other classes) is not caught it will crash a thread of the executor and can stall it indefinitely, especially with fixed size thread pools. We definitely don't want that. Our newer `TaskScheduler` component is designed to handle this by wrapping every runnable in a TaskScheduler.ErrorTrap instance that reports these errors without crashing the thread. 

I have updated several things to catch and retain Throwable instead of Exception to make this more hermetic.

So, I think our HTTP endpoints and TaskScheduler async tasks are all safe. But we've got a surprising number of other ExecutorServices scattered around that need to be protected in the same way (this is a concern even without the recent changes to gtfs-lib and osm-lib making them throw more exceptions). I see Executors in AsyncLoader (traps Throwable), S3Util (unused, traps Exceptions not Throwable), Jobs (weirdly unused and contains _only_ an ExecutorService), and ShapeDataStore (creates an ExecutorService in a local variable just for the duration of its upload operation, and doesn't trap any errors). 

To avoid this crashed thread problem, but also to eventually allow better tracking and reporting on background tasks, they should eventually all be run through the TaskScheduler component. I will delete the unused executors, and add a comment to the ShapeDataStore about why it's a special case.

Finally, AnalysisWorker also has its own ExecutorService for regional tasks. This one needs to be improved. Every task it submits is the method handleOneRegionalTask, which is mostly but not entirely guarded by try/catch blocks. The catch block swallows all Exceptions (not Throwables) and ends with: `TODO communicate regional analysis errors to the backend (in workResults)`. First, this entire method execution needs to be inside a `try`/`catch(Throwable)` block (ideally by wrapping the function call one layer up the stack); and second it needs to actually report those errors, causing regional jobs on the backend to be canceled gracefully.

abyrd:
Latest commits expose worker and backend result assembler errors to the UI. The UI hits the `running` endpoint and gets a JSON array of JobStatus objects from the backend. These now include an `errors` property, which contains a list of unique errors on that analysis from workers or the backend result assembler. Here's an example:

- [{jobId: "605869f78580f1000c23ae4c", graphId: "5f6b0549a835e54ba2c0ddc4", workerCommit: "v6.0.1",…},…]
    - 0: {jobId: "605869f78580f1000c23ae4c", graphId: "5f6b0549a835e54ba2c0ddc4", workerCommit: "v6.0.1",…}
    - 1: {jobId: "605867d58580f1000c23ae4a", graphId: "5f6b0549a835e54ba2c0ddc4", workerCommit: "v6.0.1",…}
        - activeWorkers: 0
        - complete: 0
        - deliveries: 10
        - deliveryPass: 0
        - errors: [,…]
            - 0: "UnsupportedOperationException: Path results are limited to 5000 destinations↵[detail follows]↵java.lang.UnsupportedOperationException: Path results are limited to 5000 destinations↵	at com.conveyal.r5.analyst.cluster.PathResult.<init>(PathResult.java:66)↵	at com.conveyal.r5.analyst.TravelTimeReducer.<init>(TravelTimeReducer.java:141)↵	at com.conveyal.r5.analyst.TravelTimeComputer.computeTravelTimes(TravelTimeComputer.java:76)↵	at com.conveyal.r5.analyst.cluster.AnalysisWorker.handleOneRegionalTask(AnalysisWorker.java:584)↵	at com.conveyal.r5.analyst.cluster.AnalysisWorker.lambda$run$2(AnalysisWorker.java:392)↵	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)↵	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)↵	at java.base/java.lang.Thread.run(Thread.java:834)↵"
        - graphId: "5f6b0549a835e54ba2c0ddc4"
        - incomplete: 10
        - jobId: "605867d58580f1000c23ae4a"
        - regionalAnalysis: {_id: "605867d58580f1000c23ae4a", name: "FAIL", nonce: "605867d58580f1000c23ae4b",…}
        - total: 10
        - workerCommit: "v6.0.1"
```

trevorgerhardt (approved):

PR #715, opened 2021-03-23
Record GTFS validation errors

abyrd:
Ten of each type of error are nested in the feed sub-object of the bundle object in Mongo. The full list of errors is serialized to JSON and stored in a file alongside the feed and MapDB databases, in the local cache and synced to cloud storage. The error table is emptied in the GtfsFeed object before closing it, which should save a lot of space in some cases. The format of the bulk-dumped JSON file (shown below) leaves something to be desired, but we can improve the validation and quality of information in the fields over time. The first step is to start recording it at all.

```
    {
        "affectedEntityId": null,
        "affectedEntityType": null,
        "badReference": "57463092",
        "errorCode": "ReferentialIntegrityError",
        "errorType": "ReferentialIntegrityError",
        "field": "from_trip_id",
        "file": "transfers",
        "line": 54202,
        "message": "57463092",
        "messageWithContext": "transfers line 54202, field 'from_trip_id': 57463092",
        "priority": "UNKNOWN"
    },
    {
        "affectedEntityId": null,
        "affectedEntityType": null,
        "badReference": "57462720",
        "errorCode": "ReferentialIntegrityError",
        "errorType": "ReferentialIntegrityError",
        "field": "from_trip_id",
        "file": "transfers",
        "line": 54203,
        "message": "57462720",
        "messageWithContext": "transfers line 54203, field 'from_trip_id': 57462720",
        "priority": "UNKNOWN"
    },
```

PR #719, opened 2021-04-07
Revised Activity API

abyrd:
This is a revised version of the new user activity API created in #706. This takes in to account the discussion on #716 and in subsequent meetings. The API is now more clearly defined via model classes within `UserActivityController.java`. The title of tasks is now separate from their description, with the latter changing as work is done. Fine grained progress reporting on OSM and GTFS will be shown separately, reporting the percentage of each file that's been loaded.

The endpoint will now stop reporting on tasks after their DONE or ERROR state has been delivered to a UI at least once, and one minute has passed.

Any exceptions will be attached to the Task as well as the Mongo. A common code path will be factored out for this purpose (tied in with the workProduct information), but it works in the current form.

Work product API model classes are present but not yet set. I can take care of that tomorrow, probably during the work session.

This should also be integrated with #715 so we don't give the illusion of successful GTFS loading by not showing feed validation errors. I just tried merging them and the only conflicts were imports.

### DIFF
--- a/src/main/java/com/conveyal/analysis/components/TaskScheduler.java
+++ b/src/main/java/com/conveyal/analysis/components/TaskScheduler.java
@@ -1,22 +1,18 @@
 package com.conveyal.analysis.components;
 
 import com.conveyal.analysis.UserPermissions;
-import com.conveyal.analysis.controllers.UserActivityController;
-import com.conveyal.analysis.controllers.UserActivityController.ApiTask;
+import com.conveyal.r5.analyst.progress.ApiTask;
 import com.conveyal.r5.analyst.progress.Task;
 import com.conveyal.r5.analyst.progress.TaskAction;
 import com.google.common.collect.HashMultimap;
-import com.google.common.collect.Lists;
 import com.google.common.collect.Multimaps;
 import com.google.common.collect.SetMultimap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
-import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.ExecutorService;
--- a/src/main/java/com/conveyal/analysis/components/TaskScheduler.java
+++ b/src/main/java/com/conveyal/analysis/components/TaskScheduler.java
@@ -23,6 +23,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
@@ -201,6 +202,11 @@ public class TaskScheduler implements Component {
                 }
             }))
        );
+    }
+
+    /** Get the number of slower "heavy" tasks that are queued for execution and not yet being processed. */
+    public int getBacklog() {
+        return ((ThreadPoolExecutor) heavyExecutor).getQueue().size();
     }
 
 }
--- a/src/main/java/com/conveyal/analysis/components/TaskScheduler.java
+++ b/src/main/java/com/conveyal/analysis/components/TaskScheduler.java
@@ -162,23 +162,11 @@ public class TaskScheduler implements Component {
             Set<Task> tasks = tasksForUser.get(userEmail);
             if (tasks == null) return Collections.emptyList();
             List<ApiTask> apiTaskList = tasks.stream()
-                    .map(TaskScheduler::toApiTask)
+                    .map(Task::toApiTask)
                     .collect(Collectors.toUnmodifiableList());
             tasks.removeIf(t -> t.durationComplete().getSeconds() > 60);
             return apiTaskList;
         }
-    }
-
-    /** Convert a single internal Task object to its representation for JSON serialization and return to the UI. */
-    private static ApiTask toApiTask (Task task) {
-        ApiTask apiTask = new ApiTask();
-        apiTask.id = task.id; // This can be the same as the workProduct ID except for cases with no Mongo document
-        apiTask.title = task.description;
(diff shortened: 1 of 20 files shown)


src/main/java/com/conveyal/analysis/components/TaskScheduler.java
```diff
@@ -143,19 +147,34 @@ public final void run () {
...
+     * be updated at any time. So there is a risk of sending a partially updated Task out to the UI. If this ever causes
+     * problems we'll need to lock each Task independently.
+     * Returns an empty list even when no tasks have been recorded for the user (return is always non-null).
+     */
+    public @Nonnull List<ApiTask> getTasksForUser (String userEmail) {
+        synchronized (tasksForUser) {
+            Set<Task> tasks = tasksForUser.get(userEmail);
+            if (tasks == null) return Collections.emptyList();
+            List<ApiTask> apiTaskList = tasks.stream()
+                    .map(Task::toApiTask)
```
trevorgerhardt, line 165: Should this method live in the ApiTask? `ApiTask::fromTask`.

src/main/java/com/conveyal/analysis/controllers/BundleController.java
```diff
@@ -215,38 +218,32 @@ private Bundle create (Request req, Response res) {
...
+                throw t;
               } finally {
+                // ID and create/update times are assigned here when we push into Mongo.
                 Persistence.bundles.modifiyWithoutUpdatingLock(bundle);
               }
         }));
-
+        // TODO do we really want to return the bundle here? It should not be needed until the background work is done.
+        // We could instead return the WorkProduct instance (or BaseModel instance or null) from TaskActions.
         return bundle;
```
trevorgerhardt, line 247: Should we return the `Task` (or `ApiTask`)? We wouldn't currently use that in the UI, but that might be a common pattern that we could use in the future. "The progress for the action you just took will be tracked via this task"

src/main/java/com/conveyal/analysis/models/Bundle.java
```diff
@@ -104,6 +104,7 @@ public FeedSummary clone () {
         }
     }
 
+    // The first two PROCESSING_* values are essentially deprecated in favor of Task. Consider eliminating this field.
     public enum Status {
```
trevorgerhardt, line 108: I think we can eliminate `Status` here, and we should consider changing `statusText` to something it will more correctly represent -- `errors`.

trevorgerhardt: Or something similar

src/main/java/com/conveyal/analysis/util/ProgressInputStream.java
```diff
@@ -0,0 +1,57 @@
+package com.conveyal.analysis.util;
```
trevorgerhardt, line 1: Should we create a `progress` package with the corresponding files created specifically for task management and progress reporting? Having them circularly referenced across packages like this is not ideal to me.

trevorgerhardt:
I think my biggest concerns is the location of the various files here. Like:
- `r5.analyst.progress.Task` referencing static classes from a HTTP controller.
- `ProgressInputStream` could just be in the same package as the `ProgressListener` interface since they are tied together right?

Long term, we probably should both consolidate our `utils` and consider co-locating similar `utils` into more specific folders/packages.

src/main/java/com/conveyal/analysis/util/ProgressInputStream.java
abyrd: Discussed in last call - I agree but we'll do this as part of a bigger reshuffle pulling analysis backend out above the r5 package tree.

src/main/java/com/conveyal/analysis/controllers/BundleController.java
abyrd: It seems sufficient to return only an HTTP status code "Accepted". We could also return the task and potentially even a workProduct (it turns out we have to create them and assign IDs before processing a task to allow exception handling one level up). But unless we've got a specific need for those return values, I think HTTP 202 is best.

src/main/java/com/conveyal/analysis/components/TaskScheduler.java
abyrd: I tried it both ways. Both have shortcomings. Ideally classes for the API would not be accessing non-final fields of internal data structures, and the internal data structures are in packages that ideally shouldn't even know the API exists. Respecting both constraints yields lots of verbose Java repetition and boilerplate. The real solution is language support for properties and immutable data classes (port to Swift)?

src/main/java/com/conveyal/analysis/models/Bundle.java
abyrd: I agree, I've been progressively removing updates to this field but didn't want to leave it out entirely in case UI or other components need it.

src/main/java/com/conveyal/r5/analyst/progress/Task.java
```diff
@@ -217,38 +215,69 @@ public Duration durationExecuting () {
...
+
     public Task withAction (TaskAction action) {
         this.action = action;
         return this;
     }
 
+    // We can't return the WorkProduct from TaskAction, that would be disrupted by throwing exceptions.
+    // It is also awkward to make a method to set it on ProgressListener - it's not really progress.
+    // So we set it directly on the task before submitting it. Requires pre-setting (not necessarily storing) Model._id.
+    public Task withWorkProduct (Model model) {
```
trevorgerhardt, line 253: Why don't we just pass the `WorkProduct` itself here?

src/main/java/com/conveyal/analysis/components/TaskScheduler.java
trevorgerhardt: Hmmm. Putting getters on the non-final fields of `Task` certainly seems more ideal than having `BackgroundTask` import a subclass from a Controller. What happens if we want to have another type of `ApiTask`? For example, for admin purposes?

trevorgerhardt (approved):
Approving functionality. Clean up can be done later.

trevorgerhardt (approved):

PR #721, opened 2021-04-14
Build with NEXT_PUBLIC_CYPRESS: true

trevorgerhardt:
This disables deduping UI requests that come from multiple components (the tests run faster than real world usage).

Specifically, Cypress tests failing with a "button still exists but expected not to" type error should now pass. On the region page it would load the regions, delete the region, confirm deletion, and reload the regions page very fast. If it was less than 2 seconds, the request would be deduplicated and the region would still appear.


ansoncfit (approved):
For future reference, relevant UI configuration is at https://github.com/conveyal/analysis-ui/blob/e5a1c1189994db2aa29f78159661523382d15e3c/lib/config/swr.tsx#L7

PR #723, opened 2021-04-20
Add priority to GTFS error summaries

trevorgerhardt:
Includes the priority of the GTFS error in the error type summary. This is useful for showing emphasis in the user interface depending on the priority.

I included in the error type summary because it appears that all of the errors of the same type have the same priority, if that is wrong I will move it into the error summaries.


abyrd (approved):

ansoncfit (approved):
Just discussed -- this should work with our current use of `GTFSError` and the pending `NewGTFSErrorType`.

Also, we should pull the descriptions from `Priority` into the user manual.

PR #724, opened 2021-04-23
Add Priority to each GTFSError subclass

trevorgerhardt:
Add a specific `Priority` to each `GTFSError` subclass enforced through the constructor. This affects all of the "old" GTFS errors. I selected each error's `Priority` from the files themselves or from the `NewGTFSErrorType` enum that looked most closely related to it. 

If there are desired changes to the priority levels used, please feel free to change directly or to submit a suggested change inline.

Additionally, I've removed a thrown exception during pattern naming for when pattern has a stop that doesn't exist. A error will have already been recorded, so removing that exception allows the feed creation to complete and for those recorded errors to be displayed. If there's an alternative way of achieving the same behavior, I'm open to suggestions.

### DIFF
--- a/src/main/java/com/conveyal/gtfs/error/GTFSError.java
+++ b/src/main/java/com/conveyal/gtfs/error/GTFSError.java
@@ -18,17 +18,19 @@ public abstract class GTFSError implements Comparable<GTFSError>, Serializable {
     public final String field;
     public final String affectedEntityId;
     public final String errorType;
+    public final Priority priority;
 
-    public GTFSError(String file, long line, String field) {
-        this(file, line, field, null);
+    public GTFSError(String file, long line, String field, Priority priority) {
+        this(file, line, field, priority, null);
     }
 
-    public GTFSError(String file, long line, String field, String affectedEntityId) {
+    public GTFSError(String file, long line, String field, Priority priority, String affectedEntityId) {
         this.file  = file;
         this.line  = line;
         this.field = field;
         this.affectedEntityId = affectedEntityId;
         this.errorType = this.getClass().getSimpleName();
+        this.priority = priority;
     }
 
     /**
@@ -40,6 +42,7 @@ public abstract class GTFSError implements Comparable<GTFSError>, Serializable {
         this.field = null;
         this.errorType = null;
         this.affectedEntityId = entityId;
+        this.priority = Priority.UNKNOWN;
     }
 
     /**
@@ -48,10 +51,6 @@ public abstract class GTFSError implements Comparable<GTFSError>, Serializable {
      */
     public final String getErrorCode () {
         return this.getClass().getSimpleName();
-    }
-
-    public Priority getPriority () {
-        return Priority.UNKNOWN;
     }
 
     /**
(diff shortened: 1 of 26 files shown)


src/main/java/com/conveyal/gtfs/error/MissingShapeError.java
```diff
@@ -11,10 +11,8 @@
 public class MissingShapeError extends GTFSError implements Serializable {
     public static final long serialVersionUID = 1L;
 
-    public final Priority priority = Priority.MEDIUM;
-
     public MissingShapeError(Trip trip) {
-        super("trips", trip.sourceFileLine, "shape_id", trip.trip_id);
+        super("trips", trip.sourceFileLine, "shape_id", Priority.MEDIUM, trip.trip_id);
```
ansoncfit, line 15:
```suggestion
        super("trips", trip.sourceFileLine, "shape_id", Priority.LOW, trip.trip_id);
```

src/main/java/com/conveyal/gtfs/error/RangeError.java
```diff
@@ -9,7 +11,7 @@
     final double min, max, actual;
 
     public RangeError(String file, long line, String field, double min, double max, double actual) {
-        super(file, line, field);
+        super(file, line, field, Priority.MEDIUM);
```
ansoncfit, line 14: Maybe low? I see this often (maybe with stop_times.pickup_type?)
```suggestion
        super(file, line, field, Priority.LOW);
```

src/main/java/com/conveyal/gtfs/error/URLParseError.java
```diff
@@ -1,13 +1,15 @@
...
+
 import java.io.Serializable;
 
 /** Represents a problem parsing a URL field from a GTFS feed. */
 public class URLParseError extends GTFSError implements Serializable {
     public static final long serialVersionUID = 1L;
 
     public URLParseError(String file, long line, String field) {
-        super(file, line, field);
+        super(file, line, field, Priority.MEDIUM);
```
ansoncfit, line 12:
```suggestion
        super(file, line, field, Priority.LOW);
```

ansoncfit (dismissed):
Looks good overall. I added a few small suggestions.

src/main/java/com/conveyal/gtfs/error/GTFSError.java
```diff
@@ -50,10 +53,6 @@ public final String getErrorCode () {
         return this.getClass().getSimpleName();
     }
 
-    public Priority getPriority () {
-        return Priority.UNKNOWN;
-    }
```
abyrd, line 55: The original design was for priorities to be returned by this method, which would be overridden on each subclass. This makes it clear that the priority is a characteristic of the class rather than each instance of the class. It also eliminates one field from each instance, which is a minor benefit from a storage perspective but still simpler. If we are depending on priorities being associated with classes rather than instances (in the UI for example) then I'd recommend we use the original approach.

abyrd (commented):
It seems preferable to me to use a method rather than a field to expose the priorities. If they're named or annotated in a standard way, they should be serialized to JSON fields just as fields would be.

src/main/java/com/conveyal/gtfs/error/GTFSError.java
trevorgerhardt: Updated to reflect the original design.

abyrd (approved):
Looks good. I think we could recover from the bad stop ID and complete the naming process but if we are going to forbid use of the resulting feed, returning early is sufficient.

Issue #696, opened 2021-02-25
Unify CSV rasterization


abyrd:
We currently have two different ways of converting CSV files of points into grids. Our original CSV rasterization automatically detects all numeric fields and makes one layer per numeric CSV column. Our newer process of making freeform pointsets from CSV requires one specific count field to be chosen, and converts the resulting freeform pointset into a grid. This freeform-to-grid conversion was originally intended as a stopgap measure to allow visualization of points in the map UI, and previewing accessibility results in a single point analysis (which can't display freeform points) before running a regional analysis with the true freeform layer. 

However, thinking through the UI for these various features we realized it's confusing to have two different code paths with different behavior and different parameters for rasterizing CSV. We should ensure CSV points are always converted to raster by a single code path, and try to align its parameters with the CSV-to-Freeform parameters.

In the short term, to facilitate unification we could sacrifice the ability to automatically generate multiple layers at once from a single CSV file. In fact we find automatic selection of multiple columns generates junk layers as often as not. We would not be bothered by uploading a CSV multiple times to create layers from multiple columns.

Note also that automatic column selection and multiple column selection are two independent features. I think the ideal is manual multiple column selection, and a bonus would be selection from known column names based on a pre-scan of the CSV header row (and potentially full type validation of the columns) on the UI side.

This could also tie in to a UI change where when uploading CSV, there are two independent switches for creating a raster and creating a freeform layer, both always using the same parameter fields (count and id columns, lat and lon columns etc.)

We also need to consider our existing special functionality for uploading matched OD pairs as CSV for one-to-one freeform-to-freeform analyses.

ansoncfit:
Closing for now. We'll flesh out the `CsvDataSourceIngester` added in #732 when we get to #761

Issue #603, opened 2020-10-08
Make aggregation area upload a background task


ansoncfit:
From @trevorgerhardt: If an aggregation area takes more than a few minutes to build the HTTP request is likely to time out. We should move it into a background task.

ansoncfit:
Closed in #732

PR #732, opened 2021-06-04
Spatial DataSource upload and conversion to AggregationArea

ansoncfit:
The intended flow when uploading a spatial dataset is:
- Files saved via FileStorage
- Geometries and attributes validated
- Source, with geometry type and attributes, saved
- From source page, users can convert to layers appropriate for specific uses, (multi-)selecting from the attributes with valid numeric values for certain fields (e.g. opportunity datasets), and selecting from the attributes with valid text values for others (e.g. to label multiple aggregation areas). See "manual multiple column selection" comment in https://github.com/conveyal/r5/issues/696

image

Related: https://github.com/conveyal/analysis-ui/issues/1364, https://github.com/conveyal/r5/issues/603


ansoncfit:
In terms of sequencing, the plan is first to transition shapefile polygon upload/validation and processing as an aggregation area to use the newer database and file storage components, as well as background task management.

GeoJSON polygons for use as modification polygons are probably next, as the changes from "Resource" management (currently dev-only) should be minimally disruptive. Grid upload should also be straightforward (and rarely used by most end users).

Transitioning the CSV and SHP uploads out of the existing OpportunityDatasetController will be the most disruptive, and we could implement those changes in subsequent PRs (perhaps after revising network file storage).

ansoncfit:
Questions for discussion:

* Do we prefer Controllers to have code focused on high-level passing of data between user actions and components, with implementation methods moved to model classes and `@JsonIgnore`d (or to static utility classes, or somewhere else)? For example, the current `OpprotunityDatasetController` feels unwieldy, and some of its methods could be moved elsewhere (see the attempts at modularization in `SpatialDatasetController`).
* How to mark a `SpatialDatasetSource` as valid and ready to conversion to Conveyal layers? An explicit `validated` flag, or whether or not `attributes` is set?
* With planned file storage changes, if we validate a shapefile once, can we just register the `.shp` then assume all associated files (`.dbf`, `.prj`, and optionally `.shx`) are present?

ansoncfit:
* Yes, controllers should just be HTTP glue
* Classic enterprise approach: API model classes (empty classes that only hold data) vs. internal domain model classes; compromise is to use static utility methods (equivalent to have instance of class as first parameter)
* Use explicit `attributesValid` flag, or a list of errors (if empty, then valid)
* Supply basename, instead of `.shp`

src/main/java/com/conveyal/analysis/persistence/AnalysisDB.java
```diff
@@ -31,7 +32,11 @@ public AnalysisDB (Config config) {
         // Create codec registry for POJOs
         CodecRegistry pojoCodecRegistry = fromRegistries(
                 MongoClientSettings.getDefaultCodecRegistry(),
-                fromProviders(PojoCodecProvider.builder().automatic(true).build()));
+                fromProviders(PojoCodecProvider.builder()
+                        .conventions(Conventions.DEFAULT_CONVENTIONS)
```
ansoncfit, line 36: We should discuss this change and the related `@Bson` annotations

ansoncfit:
With the recent changes on this branch, the process of aggregation area creation now uses:
* The new two-step upload/convert workflow. In https://github.com/conveyal/analysis-ui/pull/1575, you can upload a shapefile. Then from the resulting resource, you can create an aggregation area from any polygon shapefiles (dissolving to a single area by default, or keeping features separate and naming them by a selected text attribute of the polygon).
* Task tracking and activity reporting
* Components
* Database patterns (e.g. extending `BaseModel` instead of the legacy `Model`)

Instead of `SpatialDatasetSource` as proposed above, I now prefer `SpatialResource`

src/main/java/com/conveyal/analysis/controllers/OpportunityDatasetController.java
```diff
@@ -140,29 +141,31 @@ private OpportunityDatasetUploadStatus downloadLODES(Request req, Response res)
...
-                List<Grid> grids = extractor.censusDataForBounds(region.bounds, zoom);
-                createDatasetsFromPointSets(
-                        email, accessGroup, extractor.sourceName, downloadBatchId, regionId, status, grids
-                );
-            } catch (IOException e) {
-                status.completeWithError(e);
-                LOG.error("Exception processing LODES data: " + ExceptionUtils.stackTraceString(e));
-            }
-        });
+        SpatialResource source = SpatialResource.create(userPermissions, extractor.sourceName)
```
abyrd, line 152: This will create a new distinct SpatialResource entity each time someone imports from LODES. If I'm not mistaken we will end up with one entity per region, and multiple entities per region if for whatever reason the import operation is repeated (e.g. failure or deletion of resulting layers, then re-import). It doesn't seem right to have multiple entities all referencing the same single data source (the seamless census dataset for a particular year). Also, the name field in the new entity contains the human-readable name of the LODES data set, while on other entities of this type it seems like it might contain a filename.

src/main/java/com/conveyal/analysis/controllers/OpportunityDatasetController.java
```diff
@@ -140,29 +141,31 @@ private OpportunityDatasetUploadStatus downloadLODES(Request req, Response res)
...
+                .withRegion(regionId);
+
+        taskScheduler.enqueue(Task.create("Extracting LODES data")
+                .forUser(userPermissions)
+                .setHeavy(true)
+                .withWorkProduct(source)
+                .withAction((progressListener) -> {
+                    try {
+                        status.message = "Extracting census data for region";
+                        List<Grid> grids = extractor.censusDataForBounds(region.bounds, zoom);
```
abyrd, line 162: Building all the grids in memory and then handing them off to the next method is a red flag for scalability. We may want to do this streaming (with a callback to process each grid in turn).

src/main/java/com/conveyal/analysis/controllers/OpportunityDatasetController.java
```diff
@@ -140,29 +141,31 @@ private OpportunityDatasetUploadStatus downloadLODES(Request req, Response res)
...
+                        updateAndStoreDatasets(source, status, grids);
+                    } catch (IOException e) {
+                        status.completeWithError(e);
```
abyrd, line 165: Catching the exception here bypasses the background task mechanism. The endgame here is to remove the upload status object entirely and track progress purely through our new background task mechanism.

abyrd (commented):
Leaving one first round of comments - I plan to continue reviewing later. Many of these describe changes I would plan to make myself when we hand off this PR.

src/main/java/com/conveyal/analysis/controllers/OpportunityDatasetController.java
```diff
@@ -140,29 +141,31 @@ private OpportunityDatasetUploadStatus downloadLODES(Request req, Response res)
...
-                LOG.error("Exception processing LODES data: " + ExceptionUtils.stackTraceString(e));
-            }
-        });
+        SpatialResource source = SpatialResource.create(userPermissions, extractor.sourceName)
+                .withRegion(regionId);
+
+        taskScheduler.enqueue(Task.create("Extracting LODES data")
+                .forUser(userPermissions)
+                .setHeavy(true)
+                .withWorkProduct(source)
```
abyrd, line 158: As above when processing aggregation areas, the source dataset is not the work product -- we're potentially producing many work products that are not known in advance, a situation not covered by the current taskScheduler programming API. We'll need to come up with a way to cover this situation.

src/main/java/com/conveyal/analysis/controllers/OpportunityDatasetController.java
```diff
@@ -171,47 +174,28 @@ private OpportunityDatasetUploadStatus downloadLODES(Request req, Response res)
...
-
-            // Make new PointSet metadata objects.
-            // Unfortunately we can't pull this step out into a method because there are so many parameters.
-            // Some of that metadata could be consolidated e.g. user email and access group.
             OpportunityDataset dataset = new OpportunityDataset();
-            dataset.sourceName = sourceName;
-            dataset.sourceId = sourceId;
+            dataset.sourceName = source.name;
+            dataset.sourceId = source._id.toString();
+            dataset.createdBy = source.createdBy;
```
abyrd, line 188: If we allow uploading data sources, then creating opportunity data sets from them later, then the createdBy may not be the same person as on the source right?

src/main/java/com/conveyal/analysis/controllers/SpatialResourceController.java
```diff
@@ -0,0 +1,190 @@
...
+        return spatialResourceCollection.findPermitted(
+                and(eq("regionId", request.params("regionId"))),
+                request.attribute("accessGroup")
+        );
+    }
+
+    @Override
+    public void registerEndpoints (spark.Service sparkService) {
+        sparkService.path("/api/spatial", () -> {
+            sparkService.post("", this::handleUpload, toJson);
```
abyrd, line 183: Should region be included in all URLS?

src/main/java/com/conveyal/analysis/models/SpatialResource.java
```diff
@@ -0,0 +1,106 @@
...
+
+import static com.conveyal.file.FileCategory.RESOURCES;
+import static com.conveyal.r5.analyst.Grid.checkWgsEnvelopeSize;
+
+/**
+ * Record of a spatial source file (e.g. shapefile, CSV) that has been validated and can be processed into specific
+ * Conveyal formats (e.g. grids and other spatial layers). Eventually a general Resource class could be extended by
+ * SpatialResource, GtfsResource, and OsmResource?
+ */
+public class SpatialResource extends BaseModel {
```
abyrd, line 30: Do we really want a specifically SpatialResource, or a more general Resource with the sourceFormat implying that it's spatial? I guess this one needs geometry type and attribute descriptors, so it's specifically spatial. I'm wondering though where our GTFS and OSM file records fit in.

src/main/java/com/conveyal/analysis/persistence/AnalysisCollection.java
```diff
@@ -73,6 +73,14 @@ public T create(T newModel, String accessGroup, String creatorEmail) {
         return newModel;
     }
 
+    /**
+     * Note that if the supplied model has _id = null, the Mongo insertOne method will overwrite it with a new
+     * ObjectId(). We consider it good practice to set the _id for any model object ourselves, avoiding this behavior.
```
abyrd, line 78: Should we add an assertion here to require ourselves to follow this convention? _id must already exist before insert?

src/main/java/com/conveyal/analysis/persistence/AnalysisDB.java
abyrd: Yes, I would like to learn about this change and the Bson annotations.

src/main/java/com/conveyal/r5/analyst/progress/WorkProductType.java
```diff
@@ -13,13 +13,14 @@
...
+    BUNDLE, REGIONAL_ANALYSIS, AGGREGATION_AREA, OPPORTUNITY_DATASET, RESOURCE;
 
-    // Currently we have two base classes for db objects so may need to use Object instead of BaseModel parameter
-    public static WorkProductType forModel (Model model) {
+    public static WorkProductType forModel (Object model) {
         if (model instanceof Bundle) return BUNDLE;
         if (model instanceof OpportunityDataset) return OPPORTUNITY_DATASET;
         if (model instanceof RegionalAnalysis) return REGIONAL_ANALYSIS;
+        if (model instanceof AggregationArea) return AGGREGATION_AREA;
+        if (model instanceof SpatialResource) return RESOURCE; // TODO switch to spatial dataset source
```
abyrd, line 23: We should discuss making terminology more uniform here.

src/main/java/com/conveyal/analysis/spatial/Polygons.java
```diff
@@ -0,0 +1,9 @@
+package com.conveyal.analysis.spatial;
+
+public class Polygons {
```
abyrd, line 3: What is the intended use of these classes (Polygons, Points, Lines)? Or are they just left over from a draft / prototype?

abyrd (commented):
Posting this second round of commentary in case you have any feedback @ansoncfit. I will probably do one more round of review tomorrow before starting to refactor.

src/main/java/com/conveyal/analysis/controllers/AggregationAreaController.java
```diff
@@ -72,113 +91,102 @@ private FileStorageKey getStoragePath (AggregationArea area) {
...
-                throw new AnalysisServerException("Error processing/uploading aggregation area");
-            }
-
-            tempDir.delete();
-        });
+        List<SimpleFeature> finalFeatures = features;
+        taskScheduler.enqueue(Task.create("Aggregation area creation: " + source.name)
+                .forUser(userPermissions)
+                .setHeavy(true)
+                .withWorkProduct(source)
```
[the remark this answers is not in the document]
ansoncfit: I like the suggestion here. I agree it's awkward to try and use a source as a collection of aggregation areas.

src/main/java/com/conveyal/analysis/models/SpatialResource.java
ansoncfit: I've gone in circles on this. The original concept of resources would eventually include all user uploads. I then started developing this as standalone spatial database management, and by the end realized (as reflected in the Javadoc highlighted) that a common Resource superclass could make sense.

src/main/java/com/conveyal/analysis/spatial/Polygons.java
ansoncfit: Leftovers, possible container for static utility methods.

src/main/java/com/conveyal/analysis/controllers/OpportunityDatasetController.java
abyrd: We decided in today's call that we will introduce an intermediate "group of uploaded file derivatives" table in the database. Processing one uploaded file into one set of derivatives corresponds to exactly one work product in this new table.

src/main/java/com/conveyal/analysis/controllers/OpportunityDatasetController.java
abyrd: In today's call we decided to create one stand-in instance of each Conveyal-managed data source per region, and re-use that instance if multiple import operations occur in the same region. This will have the desired effect of grouping all LODES layers from the same year together, with sub-groupings for each import operation, using only general purpose code and avoiding some special case where a single application-wide entity (lacking region ID and group affiliation) represents the Conveyal-managed data source.

src/main/java/com/conveyal/analysis/persistence/AnalysisDB.java
ansoncfit: See PojoCodecProvider Builder 
Thinking back on this, I added the @bsonignore annotations in FileStorageFormat to resolve an error, which required registering the Annotation convention.

https://mongodb.github.io/mongo-java-driver/3.11/bson/pojos/#pojos-plain-old-java-objects

https://mongodb.github.io/mongo-java-driver/3.11/javadoc/org/bson/codecs/pojo/Conventions.html#ANNOTATION_CONVENTION

src/main/java/com/conveyal/analysis/persistence/AnalysisCollection.java
abyrd: It looks like we could remove the OBJECT_ID_GENERATORS convention to force explicit ID creation. https://mongodb.github.io/mongo-java-driver/3.11/bson/pojos/#conventions

src/main/java/com/conveyal/analysis/persistence/AnalysisDB.java
abyrd: We looked at this together today. FileStorageFormat is an enum with fields. Apparently the Mongo driver was trying to serialize these like normal objects, including their field values instead of just serializing the enum value name. There is a canonical instance of the enum class for each enum value, so all references to that enum value have the same values for their fields - it makes no sense for an object codec to save, then try to set the values of these fields. The Mongo documentation linked to above imples this should not be happening: "Enums are fully supported. The POJO Codec uses the name of the enum constant as the property value. This is then converted back into an Enum value by the codec using the static Enum.valueOf method." Maybe it's confused by the fact that these are public fields? But then it would be equally confused by the corresponding property accessor methods on a private field. Maybe something about this POJO Codec Registry configuration are causing it to generate codecs for enums?

abyrd: Digging a bit deeper: PojoCodecProvider.conventions is only used by method createClassModel, which will only overwrite ClassModelBuilder.conventions when non-null;  ClassModelBuilder.conventions is initialized to DEFAULT_CONVENTIONS. So I think `.conventions(Conventions.DEFAULT_CONVENTIONS)` is a no-op. I removed this (reverted the line back to the pre-change state) and when I process an upload, the FileStorageFormat appears to be serialized to Mongo as a simple string representation of the enum's name. I will be on the lookout for some other source of this problem.

abyrd:
Summarizing the current situation after this morning's group discussion: 
- Important previous observations and decisions are in our Notion work session notes from 2021-07-01.
- The corresponding UI branch/PR is conveyal/analysis-ui#1579. It has been merged so is available in dev. The testing UI for spatial data upload is in the "resources" tab of the UI (a hidden admin-group-only tab).
- Broadly, the raw file uploaded by a user is called an "upload", and our Mongo cataloged/tagged representation of that upload with metadata is our "resource" (or maybe data source), which has a subtype called "spatial resource" or "spatial data source" (feature collection in GIS terminology), which can be processed into things like "layers" and grids, usually corresponding to a single attribute of the features, maybe at a partucular zoom level.
- Our terminology is still a bit jumbled though. See July 1st document for ideas on terminology.
- As of this PR we are maintaining other old upload API enpoints, and adding this new one only to handle shapefile upload for aggregation areas. The existing upload channel for aggregation area upload was problematic and has been stripped out of the UI (the plan is to replace it entirely in the upcoming release). 
- You could say that two connecting lines in the first diagram on this PR should be highlighted in a different color: we are first building (shapefile upload -> polygon layer resource -> modification polygon) as a working prototype, which will then be expanded to handle all kinds of spatial data uploads in the diagram, then probably expanded to all uploads (though OSM / GTFS data could also be considered spatial). Due to the ambiguity of some file extensions and formats, there will need to be an enum parameter to specify the file type.
- We also require some grouping functionality. Derived entities (layers) belong to groups, which have a source (resource), which have source upload (file). We decided we want a stand alone data group table - we don't want all derived products from a single source (resource) to be in one giant implied group. For example we want a group for a single layer at N zoom levels, separate from other such groups built from the same data source (resource).

abyrd:
I have merged my most recent work to this PR. It's not complete but getting there. The API endpoints provided are those described at https://github.com/conveyal/analysis-ui/pull/1586#issue-984621711, though DELETE still needs work.

Using the DataSource upload mechanism in that UI branch, you should be able to upload GeoJson, GeoPackage, or Shapefile and they should all appear in the list of DataSources with metadata about their geometry type and attributes. So far only SHP can be converted to aggregation areas though.

In using the common GeoTools feature source interface for all these formats, some validation has been difficult to achieve. GeoTools seems to prefer silently ignoring or correcting discrepancies in somewhat arbitrary ways. For example, QGIS will happily export GeoJSON with a non-WGS84 CRS, though this is not supported by the GeoJSON spec. GeoTools will then read it as WGS84 without checking for the presence of a `crs` property and we will only report a failure due to the coordinate range checking I added.

I hope to improve the data validation in the future but don't want to let that delay things any further - for now it should at least work with well-formed inputs.

abyrd:
I just pushed a few updates to group data sets derived from DataSources. Now when you make AggregationAreas from a shapefile, when the operation completes a DataGroup will be added to mongo, and all the AggregationArea records will have a dataGroupId field referencing that new group. The status API endpoint will then report a workProduct like this:
```
"workProduct": {
                "type": "AGGREGATION_AREA",
                "id": "6130df91c9e4c841fe4a14ff",
                "regionId": "610272ee4cf3e69f54ef1151",
                "isGroup": true
            }
```
When isGroup is true, the id field is the groupId of all the created entities of the given type (rather than the id of a single entity). I tried creating a new workProduct type DATA_GROUP, but this opens a can of worms of a multi-level type system (the type is actually something like `DATA_GROUP<AGGREGATION_AREA>`). I find group/single flag much cleaner to work with.

Based on that work product object, you can then hit `http://localhost:7070/api/region/610272ee4cf3e69f54ef1151/aggregationArea?regionId=610272ee4cf3e69f54ef1151&dataGroupId=6130da5a386fed022073ce5e` and get a list of all the work products in that region, filtered down to the ones in the new group, each with their own ID. You can then hit one of them to fetch an individual aggregationArea if needed:
`http://localhost:7070/api/region/610272ee4cf3e69f54ef1151/aggregationArea/6130da5f386fed022073ce7c`

Possible next steps:

The region is repeated twice in that first URL - let's discuss simplifying these API paths tomorrow, potentially making regionId always a query parameter.)

We may not actually want or need to insert documents in the Mongo dataGroups collection - currently they only contain an auto-generated note on how the group was created. The data groups could just be implicit. We previously decided not to use the dataSourceId to implicitly group derived products, since multiple batches could be made from the same dataSource, but once dataGroupIds are introduced it's not strictly necessary to materialize them as dataGroup records in the database.

Known issues:

You may need to drop or transform any existing aggregationAreas collection for this to work - otherwise it will try to deserialize some fields from older database records that do not have setters in Java.

**The aggregationAreaController is currently hard-wired to use nameProperty = "dist_name"** (from the LA public council districts shapefile). This was just a workaround while I couldn't get the frontend to send the nameProperty.

src/main/java/com/conveyal/analysis/datasource/DataSourceIngester.java
```diff
@@ -0,0 +1,74 @@
...
+    public static DataSourceIngester forFormat (FileStorageFormat format) {
+        if (format == SHP) {
+            return new ShapefileDataSourceIngester();
+        } else if (format == GEOJSON) {
+            return new GeoJsonDataSourceIngester();
+        } else if (format == TIFF) { // really this enum value should be GEOTIFF rather than just TIFF.
+            return new GeoTiffDataSourceIngester();
+        } else {
+            return new GeoPackageDataSourceIngester();
+        }
```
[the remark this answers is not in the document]
abyrd: Changed.

src/main/java/com/conveyal/analysis/datasource/DataSourceUploadAction.java
```diff
@@ -0,0 +1,143 @@
...
+    // Using named and well-defined classes for these background actions makes data flow and depdendencies clearer.
+    private final FileStorage fileStorage;
+    private final AnalysisCollection<DataSource> dataSourceCollection;
+
+    /** The files provided in the HTTP post form. These will be moved into storage. */
+    private final List<FileItem> fileItems;
+
+    /**
+     * This DataSourceIngester provides encapsulated loading and validation logic for a single format, by composition
+     * rather than subclassing. Format ingestion does not require access to the fileStorage or the database collection.
```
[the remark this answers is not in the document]
abyrd: No, this refers to the fact that a DataSourceIngester handling a specific format plugs into the DataSourceUploadAction, rather than having separate subclasses of DataSourceUploadAction for each format (which is how it originally worked). But then this "plugin" DataSourceIngester ended up having subclasses, so we've got both composition and inheritance here. I plan to factor a bunch of common logic out of the DataSourceIngester subclasses which may again reduce inheritance.

src/main/java/com/conveyal/analysis/datasource/DataSourceUploadAction.java
```diff
@@ -0,0 +1,143 @@
...
+        final String sourceName = getFormField(formFields, "sourceName", true);
+        final String regionId = getFormField(formFields, "regionId", true);
+        final List<FileItem> fileItems = formFields.get("sourceFiles");
+
+        FileStorageFormat format = detectUploadFormatAndValidate(fileItems);
+        DataSourceIngester ingester = DataSourceIngester.forFormat(format);
+
+        String description = "From uploaded files: " + fileItems.stream()
+                .map(FileItem::getName).collect(Collectors.joining(", "));
+        ingester.initializeDataSource(sourceName, description, regionId, userPermissions);
```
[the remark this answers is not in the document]
abyrd: I think we want both - I already added an originalFileName field to DataSource but am not setting it yet. I just updated it to set both originalFileName and description.

src/main/java/com/conveyal/analysis/datasource/GeoJsonDataSourceIngester.java
```diff
@@ -0,0 +1,179 @@
...
+ * further nested into a top-level object of type FeatureCollection. We only support GeoJSON whose top level object is
+ * a FeatureCollection (not a single Feature or a single Geometry), and where every geometry is of the same type.
+ *
+ * Section 4 of the GeoJSON RFC at https://datatracker.ietf.org/doc/html/rfc7946#section-4 defines the only acceptable
+ * coordinate reference system as WGS84. You may notice some versions of the GeoTools GeoJSON handler have CRS parsing
+ * capabilities. This is just support for an obsolete feature and should not be invoked. We instead range check all
+ * incoming coordinates (via a total bounding box check) to ensure they look reasonable in WGS84.
+ *
+ * Current stable Geotools documentation (version 25?) shows a GeoJSONFeatureSource and CSVFeatureSource.
+ * We're using 21.2 which does not have these.
```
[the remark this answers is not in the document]
abyrd: This whole Javadoc section needs to be cleaned up, with some of it merged into my GeoJSON handling notes and put in notion. I hope to do that while getting this merged. I just removed these few clearly-incorrect lines.

src/main/java/com/conveyal/analysis/models/SpatialDataSource.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static com.conveyal.file.FileCategory.DATASOURCES;
+import static com.conveyal.r5.analyst.Grid.checkWgsEnvelopeSize;
+
+/**
+ * A SpatialDataSource is metadata about a user-uploaded file containing geospatial features (e.g. shapefile, GeoJSON,
+ * or CSV containing point coordinates) that has been validated and is ready to be processed into specific Conveyal
```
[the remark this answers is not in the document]
abyrd: For background tasks we have two failure modes: action throws exception which is caught by the taskScheduler, and errors (in this case DataSourceValidationIssues) placed on the model object stored in Mongo. The former are "hard failures" and can still be seen in the UI. The latter are soft failures that allow accumulating and reviewing a lot of validation issues.

I have started storing the new work product in Mongo and reporting the WorkProduct to the task scheduler near the end of the TaskAction logic. This way, hard failures do not leave orphan objects in Mongo.

src/main/java/com/conveyal/analysis/controllers/AggregationAreaController.java
```diff
@@ -72,113 +91,102 @@ private FileStorageKey getStoragePath (AggregationArea area) {
...
+
+        File sourceFile;
+        List<SimpleFeature> features = null;
+
+        if (SHP.equals(source.sourceFormat)) {
+            sourceFile = fileStorage.getFile(source.storageKey());
+            ShapefileReader reader = null;
+            try {
+                reader = new ShapefileReader(sourceFile);
+                features = reader.wgs84Stream().collect(Collectors.toList());
```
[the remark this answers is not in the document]
abyrd: I moved these comments into the source code.

src/main/java/com/conveyal/analysis/controllers/AggregationAreaController.java
```diff
@@ -72,113 +91,102 @@ private FileStorageKey getStoragePath (AggregationArea area) {
...
-            UnaryUnionOp union = new UnaryUnionOp(geometries);
-            // Name the area using the name in the request directly
-            String maskName = query.get("name").get(0).getString("UTF-8");
-            areas.put(maskName, union.union());
-        } else {
-            // Don't union. Name each area by looking up its value for the name property in the request.
-            String nameProperty = query.get("nameProperty").get(0).getString("UTF-8");
-            features.forEach(f -> areas.put(readProperty(f, nameProperty), (Geometry) f.getDefaultGeometry()));
+        if (GEOJSON.equals(source.sourceFormat)) {
+            // TODO implement
```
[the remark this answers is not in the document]
abyrd: Fixed.

src/main/java/com/conveyal/analysis/controllers/AggregationAreaController.java
```diff
@@ -72,113 +91,102 @@ private FileStorageKey getStoragePath (AggregationArea area) {
...
-            }
-
-            tempDir.delete();
-        });
+        List<SimpleFeature> finalFeatures = features;
+        taskScheduler.enqueue(Task.create("Aggregation area creation: " + source.name)
+                .forUser(userPermissions)
+                .setHeavy(true)
+                .withWorkProduct(source)
+                .withAction(progressListener -> {
```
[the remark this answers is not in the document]
abyrd: Factored out into `AggregationAreaDerivation implements DataDerivation<SpatialDataSource, AggregationArea>`

src/main/java/com/conveyal/analysis/models/AggregationArea.java
```diff
@@ -1,14 +1,33 @@
...
+import com.conveyal.analysis.UserPermissions;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 
 /**
  * An aggregation area defines a set of origin points to be averaged together to produce an aggregate accessibility figure.
  * It is defined by a geometry that is rasterized and stored as a grid, with pixels with values between 0 and 100,000
  * depending on how much of that pixel is overlapped by the mask.
  */
-public class AggregationArea extends Model {
+public class AggregationArea extends BaseModel {
```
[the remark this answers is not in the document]
abyrd: This has all been clarified - when polygons are not fused we get many aggregation areas from a single spatial data source. These are then grouped into DataGroups, because we might produce one group per zoom level for example, all from the same source. I updated the javadoc as well.

src/main/java/com/conveyal/analysis/models/SpatialDataSource.java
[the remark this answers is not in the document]
abyrd: My original reply was kind of hard to understand, I rephrased it for future reference.

src/main/java/com/conveyal/analysis/controllers/OpportunityDatasetController.java
```diff
@@ -297,8 +295,9 @@ private static FileStorageFormat getFormatCode (PointSet pointSet) {
     /**
      * Get the specified field from a map representing a multipart/form-data POST request, as a UTF-8 String.
      * FileItems represent any form item that was received within a multipart/form-data POST request, not just files.
+     * This is a static utility method that should be reusable across different HttpControllers.
      */
-    private String getFormField(Map<String, List<FileItem>> formFields, String fieldName, boolean required) {
+    public static String getFormField(Map<String, List<FileItem>> formFields, String fieldName, boolean required) {
```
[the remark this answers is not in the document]
abyrd: It has general uses outside either class - I moved it to HttpUtils as a utility function alongside getRequestFiles.

src/main/java/com/conveyal/analysis/controllers/HttpController.java
```diff
@@ -1,5 +1,8 @@
 package com.conveyal.analysis.controllers;
 
+import com.conveyal.analysis.UserPermissions;
+import spark.Request;
+
```
ansoncfit, line 5:
```suggestion
```

src/main/java/com/conveyal/analysis/controllers/OpportunityDatasetController.java
```diff
@@ -171,47 +170,28 @@ private OpportunityDatasetUploadStatus downloadLODES(Request req, Response res)
...
-                                                                       OpportunityDatasetUploadStatus status,
-                                                                       List<? extends PointSet> pointSets) {
+    private void updateAndStoreDatasets (SpatialDataSource source,
+                                         OpportunityDatasetUploadStatus status,
+                                         List<? extends PointSet> pointSets) {
         status.status = Status.UPLOADING;
         status.totalGrids = pointSets.size();
-
         // Create an OpportunityDataset holding some metadata about each PointSet (Grid or FreeForm).
         final List<OpportunityDataset> datasets = new ArrayList<>();
```
ansoncfit, line 179: We no longer return `datasets`, so no need to create or populate it.

src/main/java/com/conveyal/analysis/datasource/GeoJsonDataSourceIngester.java
```diff
@@ -0,0 +1,183 @@
...
+import com.conveyal.file.FileStorageFormat;
+import com.conveyal.r5.analyst.progress.ProgressListener;
+import com.conveyal.r5.util.ShapefileReader;
+import com.conveyal.r5.util.ShapefileReader.GeometryType;
+import org.geotools.data.Query;
+import org.geotools.data.geojson.GeoJSONDataStore;
+import org.geotools.data.simple.SimpleFeatureSource;
+import org.geotools.feature.FeatureCollection;
+import org.geotools.feature.FeatureIterator;
+import org.geotools.geometry.jts.ReferencedEnvelope;
```
ansoncfit, line 15:
```suggestion
import com.conveyal.r5.util.ShapefileReader.GeometryType;
import org.geotools.data.Query;
import org.geotools.data.simple.SimpleFeatureSource;
import org.geotools.feature.FeatureCollection;
import org.geotools.feature.FeatureIterator;
```

src/main/java/com/conveyal/analysis/datasource/GeoTiffDataSourceIngester.java
```diff
@@ -0,0 +1,92 @@
...
+        } catch (IOException e) {
+            throw new DataSourceException("Could not read GeoTiff.", e);
+        }
+        // Transform to WGS84 to ensure this will not trigger any errors downstream.
+        CoordinateReferenceSystem coverageCrs = coverage.getCoordinateReferenceSystem2D();
+        MathTransform wgsToCoverage, coverageToWgs;
+        ReferencedEnvelope wgsEnvelope;
+        try {
+            wgsToCoverage = CRS.findMathTransform(DefaultGeographicCRS.WGS84, coverageCrs);
+            coverageToWgs = wgsToCoverage.inverse();
```
ansoncfit, line 67: `coverageToWgs` appears to be unused.

src/main/java/com/conveyal/analysis/datasource/SpatialAttribute.java
```diff
@@ -0,0 +1,63 @@
+package com.conveyal.analysis.datasource;
+
+import org.locationtech.jts.geom.Geometry;
+import org.opengis.coverage.SampleDimensionType;
+import org.opengis.feature.type.AttributeDescriptor;
+import org.opengis.feature.type.AttributeType;
+
+/**
+ * In OpenGIS terminology, SpatialResources contain features, each of which has attributes. This class represents a
```
ansoncfit, line 9:
```suggestion
 * In OpenGIS terminology, SpatialDataSources contain features, each of which has attributes. This class represents a
```

src/main/java/com/conveyal/analysis/datasource/SpatialLayers.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Utility class with common static methods for validating and processing uploaded spatial data files.
+ */
+public abstract class SpatialLayers {
+
+    /**
+     * FIXME Originally used for OpportunityDataset upload, moved to SpatialLayers but should be named DataSource
```
ansoncfit, line 20: Should we rename this class?

src/main/java/com/conveyal/analysis/datasource/derivation/AggregationAreaDerivation.java
```diff
@@ -0,0 +1,225 @@
...
+     */
+    private AggregationAreaDerivation (FileStorage fileStorage, AnalysisDB database, Request req) {
+
+        // Before kicking off asynchronous processing, range check inputs to fail fast on obvious problems.
+        userPermissions = UserPermissions.from(req);
+        dataSourceId = req.queryParams("dataSourceId");
+        nameProperty = req.queryParams("nameProperty"); //"dist_name"; //
+        zoom = parseZoom(req.queryParams("zoom")); 
+        checkNotNull(dataSourceId);
+        checkNotNull(nameProperty);
```
ansoncfit, line 84: Should nulls be allowed, to indicate the area should be the union of the input features?

src/main/java/com/conveyal/analysis/models/SpatialDataSource.java
```diff
@@ -0,0 +1,59 @@
...
+import org.opengis.referencing.operation.TransformException;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static com.conveyal.file.FileCategory.DATASOURCES;
+import static com.conveyal.r5.analyst.Grid.checkWgsEnvelopeSize;
```
ansoncfit, line 23:
```suggestion
import com.conveyal.analysis.UserPermissions;
import com.conveyal.analysis.datasource.SpatialAttribute;
import com.conveyal.file.FileStorageKey;
import com.conveyal.r5.util.ShapefileReader;
import org.bson.codecs.pojo.annotations.BsonDiscriminator;

import java.util.List;

import static com.conveyal.file.FileCategory.DATASOURCES;
```

src/main/java/com/conveyal/analysis/persistence/AnalysisDB.java
```diff
@@ -1,18 +1,24 @@
 package com.conveyal.analysis.persistence;
 
+import com.conveyal.analysis.models.BaseModel;
+import com.conveyal.analysis.models.SpatialDataSource;
 import com.mongodb.MongoClientSettings;
```
ansoncfit, line 5:
```suggestion
```

src/main/java/com/conveyal/file/FileStorageFormat.java
```diff
@@ -1,22 +1,39 @@
...
+    OSMPBF("pbf", "application/octet-stream"),
+    // Also can be application/geo+json, see https://www.iana.org/assignments/media-types/application/geo+json
+    // The extension used to be defined as .json TODO ensure that changing it to .geojson hasn't broken anything.
+    GEOJSON("geojson", "application/json"),
+    // See requirement 3 http://www.geopackage.org/spec130/#_file_extension_name
+    GEOPACKAGE("gpkg", "application/geopackage+sqlite3");
 
+    // These should not be serialized into Mongo. Default Enum codec uses String name() and valueOf(String).
+    // TODO clarify whether the extension is used for backend storage, or for detecting type up uploaded files.
+    // TODO array of file extensions, with the first one used canonically in FileStorage and the others for detection.
```
ansoncfit, line 32: 👍

src/main/java/com/conveyal/r5/analyst/scenario/IndexedPolygonCollection.java
```diff
@@ -104,9 +107,11 @@ public IndexedPolygonCollection (
     }
 
     public void loadFromS3GeoJson() throws Exception {
-        InputStream polygonInputStream = WorkerComponents.fileStorage.getInputStream(POLYGONS, polygonLayer);
-        FeatureJSON featureJSON = new FeatureJSON();
-        FeatureCollection featureCollection = featureJSON.readFeatureCollection(polygonInputStream);
+        // FIXME this needs to be adapted to new SpatialDataSource. How will we handle .gz data?
```
ansoncfit, line 110: Are there end users who rely on .gzipped congestion area polygons?

src/main/java/com/conveyal/r5/util/ShapefileReader.java
```diff
@@ -34,14 +41,18 @@
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
+import static com.conveyal.r5.util.ShapefileReader.GeometryType.LINE;
+import static com.conveyal.r5.util.ShapefileReader.GeometryType.POINT;
+import static com.conveyal.r5.util.ShapefileReader.GeometryType.POLYGON;
```
ansoncfit, line 46: Unused?

ansoncfit (commented):
Feel free to merge after addressing my in-line comments (as the original author of this PR, I cannot formally approve or request changes). We expect patches may be needed in our upcoming test cycle.

src/main/java/com/conveyal/analysis/datasource/GeoTiffDataSourceIngester.java
abyrd: I wanted to validate that both transforms can be created to detect any problems early. Transforms in both directions will be used - for example, below the wgsEnvelope is performing this transform. So I guess this call is redundant since creating the wgsEnvelope would fail if GeoTools doesn't know how to convert the CRS to WGS84.

src/main/java/com/conveyal/r5/analyst/scenario/IndexedPolygonCollection.java
abyrd: We do have some pre-existing gzipped congestion areas already on S3. However they're in a different bucket (polygons) and this has been explained to users as a new feature subject to change. So if we get gzip support added in time, we still need to move all those polygon files over to the datasources bucket on S3 and add database records for them. Probably the easiest way to do that is to just re-upload them via the UI. If we re-up them without gzipping them, I believe most of the layers involved (client to backend, backend to s3) will automatically compress the data stream so it's not that wasteful - only the storage space on s3 will be larger. Gzip support is only tricky because it alters the recognizable extensions of the files, and is typically implemented as an InputStream wrapping the FileInputStream, but GeoTools reads from Files rather than InputStreams.

I kind of suspect we're going to eventually end up implementing our own GeoJson support that works in a completely streaming manner, to allow for gzipped files and very large files, and to validate the five or six important constraints we're counting on when reading GeoJson. But for the time being we will be leaning on GeoTools so have to work within the limitations of its GeoJson extension.

src/main/java/com/conveyal/analysis/datasource/SpatialLayers.java
abyrd: I have renamed it and done some other refactoring in 014db0689035d259d2443be86f8ee731ed87a75e.

src/main/java/com/conveyal/analysis/datasource/GeoTiffDataSourceIngester.java
abyrd: I added a comment to explain why these unused transforms are looked up.

src/main/java/com/conveyal/r5/util/ShapefileReader.java
abyrd: I cleaned up unused imports.

src/main/java/com/conveyal/analysis/datasource/derivation/AggregationAreaDerivation.java
abyrd: I have changed this to allow nulls when merging will not happen, but have also added an explicit mergePolygons parameter. It seemed strange to infer something as specific as merging all polygons into one from the simple fact that a particular parameter was not supplied.

abyrd:
Thanks for the review @ansoncfit. I think I've responded to or fixed all your points and will now merge this - this is getting pretty big, and we can follow up with more changes if needed. Note that a few things have changed in the API:

- `GET /api/aggregationArea?regionId=X&dataGroupId=Y` now returns the aggregation area mongo documents
- `GET /api/aggregationArea/:_id/gridUrl` now returns a download URL for the mask grid for a single aggregation area
- `POST /api/aggregationArea` is now used to create new aggregation areas from dataSources

This final post method now takes the following parameters:
- `dataSourceId` the ID of a polygon shapefile data source
- `mergePolygons` if supplied and true, merge all polygons into one nameless polygon
- `nameProperty` the shapefile attribute containing the name of each polygon (required if mergePolygons is not true)
- `zoom` the zoom level of the mask grid, now range-checked and constrained to the range 9-12

PR #736, opened 2021-07-01
File and logging tweaks

abyrd:
This PR makes two small changes: 
- When files are moved into storage they are set read-only at the filesystem level, precluding accidental corruption or deletion, for example of MapDB files (which after a recent change are also opened in read-only mode for good measure).
- Removes some repetitive code from HTTP server exception handling code, allowing FORBIDDEN and UNAUTHORIZED exceptions to be handled by the same method as all other exceptions, which in turn causes them to fire error events, making it possible to log or otherwise broadcast messages about these errors.
- Before review and merge, we should follow up with other planned changes to respondToException and ErrorEvent, to include the HTTP request path in the ErrorEvent and log it. Authentication code (cluster-specific) can then be changed back to use Type.FORBIDDEN.

### DIFF
--- a/.github/workflows/cypress-integration.yml
+++ b/.github/workflows/cypress-integration.yml
@@ -16,7 +16,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           repository: conveyal/analysis-ui
-          ref: a869cd11919343a163e110e812b5d27f3a4ad4c8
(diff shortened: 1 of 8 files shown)


abyrd:
This has been tested on staging and works as intended.

trevorgerhardt (approved):

Issue #740, opened 2021-08-02
Frequency offset lookup uses wrong tripScheduleIndex


ansoncfit:
The changes to trip pattern filtering in https://github.com/conveyal/r5/pull/708 (included in v6.4) introduced a bug related to frequency offsets.

Before the changes, inactive and frequency-based services were skipped within the loop over trips, allowing for simple incrementing of `tripScheduleIndex`:
https://github.com/conveyal/r5/blob/v6.2/src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java#L630-L633

But with the changes to iterate over only the active frequency patterns, `tripScheduleIndex` is no longer incremented for inactive services, so an incorrect index (too low) may be used at https://github.com/conveyal/r5/blob/v6.4/src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java#L674

This bug was discovered using the STM (Montreal) GTFS, which has frequency-based trips on the same pattern with different numbers of frequency entries.

We could resolve this by generating frequency offsets only for active services.

abyrd:
We should also add some more assertions that array/list lengths are all equal, and probably some "desert tests" with more strongly varying frequency schedule entries.

PR #741, opened 2021-08-06
Look up random offsets using TripSchedule object

abyrd:
This is a minimal fix for #740, where the running frequency trip indexes from filtered patterns were being used as if they were global trip indexes from unfiltered patterns.

I tried several other approaches, but they all run into situations that require extensive refactoring. I'll describe these below as next steps.

This minimum viable fix builds a secondary map of the offsets. The offsets are randomized and phasing is solved as before, yielding a map keyed on (pattern index, trip index, freqEntryIndex), which is then copied into a map keyed on (TripSchedule, freqEntryIndex) , avoiding translation from filtered trip indexes to unfiltered ones by using a direct object reference.


src/main/java/com/conveyal/r5/profile/FrequencyRandomOffsets.java
```diff
@@ -179,10 +201,20 @@ public void randomize () {
...
             }
+            // Copy results of randomization to a Map keyed on TripSchedules (instead of TripPattern index ints)
+            offsetsForTripSchedule.clear();
+            for (TIntObjectIterator<int[][]> it = offsets.iterator(); it.hasNext(); ) {
+                it.advance();
+                TripPattern tripPattern = data.tripPatterns.get(it.key());
+                int[][] offsetsForTrip = it.value();
+                for (int ts = 0; ts < tripPattern.tripSchedules.size(); ts++) {
+                    TripSchedule tripSchedule = tripPattern.tripSchedules.get(ts);
+                    offsetsForTripSchedule.put(tripSchedule, offsetsForTrip[ts]);
```
abyrd, line 215: Looking at this again, I think `offsetsForTrip[ts]` can be null when it has no frequency patterns and there's no point in storing those.

abyrd:
Problems encountered while working on this PR:
- Existing tests are tightly coupled to implementation. At first I was passing the FilteredTripPatterns into the FrequencyRandomOffsets, but the tests did not filter patterns, and even trivial "keep everything" filtering could not be performed because the test networks lacked route IDs and other things accessed by the filtering logic. Outputs were accessed by index in the internal offsets array of FrequencyRandomOffsets.

abyrd:
Problems encountered while working on this PR:
- The multi-criteria raptor router and existing FrequencyRandomOffsets tests do not filter patterns or trips. At first I was passing the FilteredTripPatterns into the FrequencyRandomOffsets, but this required adding filtering in tests and McRaptor. Even trivial "keep everything" filtering could not be performed because the test networks lacked things like route IDs accessed by the filtering logic. The solution adopted in this PR is to always randomize the full unfiltered set of frequency trips. This is in theory less efficient but in practice I don't expect it to have much of an impact. In fact, this is what we have always been doing, but it necessitates translation from unfiltered to filtered pattern indexes (which we sidestep here by using TripSchedule object references instead of integer indexes).
- Phasing of one frequency entry could in theory depend on a different trip pattern or trip that is filtered out on certain days. If we ever switch to randomizing only the active (running) frequency entries, we'd want to detect this situation and skip the phasing lookup on these entries.
- Frequency entries are not materialized as objects, so they can't be referenced. This takes a toll when associating them with IDs and with phase source information like stops that must be looked up by ID. They are represented as parallel arrays (a column store) probably mirroring the EdgeStore. But there are so few frequency entries that I'm not sure this optimization creates any observable improvement (I'm not sure it even helps much with edges since we now have so many fields in them).
Logic to process them ends up buried in other loops and methods. It's awkward to process frequency routes in stages. We should be looking up the phasing target stops and frequency entries during scenario application, at the modification resolution stage.
- The information that is looked up in advance is stored in untyped int arrays of ints in TransitLayer: `Map<String, int[]> frequencyEntryIndexForId`. These should be replaced with object references. But currently there are no frequency entry objects to reference.
- This `frequencyEntryIndexForId` is initialized in TransitLayer.rebuildTransientIndexes, as is patternsForStop. Neither of these is marked transient, so presumably they're being serialized into the network files and overwritten. This should be changed, but will change the file format (possibly in a way that is backward compatible since the new network class will not need to read the transient fields - but this needs to be verified). This probably deserves to be tracked in a separate issue.

Other observations:
- FilteredPatterns never have scheduled and frequency trips mixed together. Unfiltered TripPatterns do mix them together.  Most of the logic becomes much simpler if we pass in the list of filtered patterns and only look at the running frequency trips, rather than all trips on all patterns.
- However, we still need the unfiltered frequency patterns for one specific thing: we use them to look up stop indexes for phased frequency entries. We should not let this single narrow case have such a strong effect on the design of the monte carlo randomization class.
- Filtered patterns do not hold references to their unfiltered counterparts, nor to their stop patterns. This is fine, but affects what we need to resolve or store in advance of performing the randomization.
- Existing tests are tightly coupled to implementation. Outputs were accessed by index in the internal offsets array of FrequencyRandomOffsets. We should migrate all such uses to encapsulated methods.
- We may want to revise the phase solving loop. Currently it seems to be designed around the case where many or most routes use phasing. Rather than setting everything to -1 and looping over the whole list until everything is solved, we may want to iterate once over all the frequency routes, creating a set of the (usually very few) phased routes or a full dependency tree and concentrate on only those. This is purely about optimization and readability though - not necessary.

Proposed refactoring:
- Create FrequencyEntry class holding fields from (this will require a new network file version number)
- Create PhaseSourceInfo class to store resolved phase FrequencyEntry, together with stop numbers etc.
- Pull phase source lookup out to scenario resolution, or at least to FrequencyRandomOffsets construction/setup

abyrd:
I have updated this branch, adding more checkState assertions, then revising the grid test framework to create a failing test at fa62067, then applying the fix to make the test pass after 6b1c218.

src/main/java/com/conveyal/r5/profile/FrequencyRandomOffsets.java
```diff
@@ -179,10 +224,27 @@ public void randomize () {
...
-
-            if (remainingAfterPreviousRound == remaining && remaining > 0) {
+            if (remainingAfterPreviousPass == remaining && remaining > 0) {
                 throw new IllegalArgumentException("Cannot solve phasing, you may have a circular reference!");
             }
+            // Copy results of randomization to a Map keyed on TripSchedules (instead of TripPattern index ints). This
+            // allows looking up offsets in a context where we only have a filtered list of running frequency trips and
+            // don't know the original unfiltered index of the trip within the pattern. Ideally we'd just build the
+            // original map keyed on TripSchedules (or hypothetical FreqEntries) but that reqires a lot of refactoring.
+            offsetsForTripSchedule.clear();
```
ansoncfit, line 234: What effect does this have? Is it just a safeguard, or is it ever expected that this map will have mappings at this point?

src/main/java/com/conveyal/r5/profile/FrequencyRandomOffsets.java
```diff
@@ -179,10 +224,27 @@ public void randomize () {
...
+            for (TIntObjectIterator<int[][]> it = offsets.iterator(); it.hasNext(); ) {
+                it.advance();
+                TripPattern tripPattern = data.tripPatterns.get(it.key());
+                int[][] offsetsForTrip = it.value();
+                for (int ts = 0; ts < tripPattern.tripSchedules.size(); ts++) {
+                    TripSchedule tripSchedule = tripPattern.tripSchedules.get(ts);
+                    // On patterns with mixed scheduled and frequency trips, scheduled trip slots will be null.
+                    // Maps can store null values, but there's no point in storing them. We only store non-null arrays.
```
ansoncfit, line 242: Note to self: reads from this map occur only in `getOffsetSeconds (TripSchedule tripSchedule, int freqEntryIndex)`, which is only called when iterating over trips that have frequency entries. So indeed, there is no point in storing a `null` value that will never be read.

ansoncfit (approved):
Thanks for the fix.

src/main/java/com/conveyal/r5/profile/FrequencyRandomOffsets.java
abyrd: Randomize() is called repeatedly, once at each iteration of the raptor search. the first time it's called, the offsets map is indeed empty. But all subsequent times, it contains the offsets from the last time randomize() was called. Clearing an existing map is slightly more efficient than constructing a new one, if only because it already contains an array of exactly the right size to hold the mappings.

PR #742, opened 2021-08-11
Scenarios, controller cleanup, and updated GTFS endpoints

trevorgerhardt:
This change set is required by and depends on https://github.com/conveyal/analysis-ui/pull/1572. I would welcome any and all kinds of feedback on individual changes here!

There are three main parts of this change set: 
1. Changes the single point and regional analysis from looking up modification IDs by Project and Variant Index to receiving the list of modification IDs along with the request itself.
2. A new, simpler `GTFSController` that replaces the `GTFSGraphQLController`.
3. Removal of controller endpoints that have been migrated to the Next.js API

# Scenarios
The related files for this changed in small, but important ways. 

### `AnalysisRequest.java`
The  `AnalysisRequest` model has three new parameters (`regionId`, `scenarioId`, and `modificationIds`) and removes one previous (`variantIndex`). Additionally, it takes a new `List<Modification>` parameter in it's `populateTask` method, allowing us to separate out the retrieval of modifications into the Broker and RegionalAnalysis controllers themselves. By doing a stricter bounds check it also allows removal of any database interaction within the model. 

The `regionId`, `projectId` and `scenarioId` are all entirely for reference now. The `bundleId` and `modificationIds` are all that is needed to make the request object and send it to a worker.

### `BrokerController.java` & `RegionalAnalysisController.java`
Now looks up the modifications and transforms them to their R5 counterpart from the incoming request itself. Importantly this allows for (in the future) us to use a stricter incoming version check by passing in all of the modification nonce's in a separate array. 

> @abyrd and I spoke about this awhile back. The version checking has not been implemented here yet, but it allows for us to do it in a more relaxed way (for single point requests) and stricter way (for regional analyses). It should only be a few line change so it could be done along with this PR or in a future PR.

### `SinglePointEvent.java` and WorkerTags
Have been modified to take new IDs. With this new paradigm, a `projectId` is very much diminished in importance as the `scenarioId` indicates the collection of modifications used and the `bundleId` points us to the underlying feed and street data.

# GTFS Controller
I originally intended to add endpoints to the `GTFSGraphQLController`, but as soon as I began down that path it became clear that accessing the data I needed directly would be much, much easier. Getting stops, patterns, trips, and routes via their own endpoints should reduce overall data transfer and allow for much better caching. 

I also added individual API response transforms for each type. Admittedly, I'm not sure if this is the best way to go about it. For the stops response, we're creating thousands of Java objects to convert them to their JSON representation as a string. 

Stops In MapDB -> List of stops in API Response format -> JSON string

Creating the JSON directly may be a faster way to go. Also, `ProtocolBuffers` may also offer savings both in the creation overhead and the size sent over the wire. The formats may be simple enough to pin to a version already. For example, stops as JSON are just:

```js
[{_id: 'stopId', name: 'Stop name', lat: 50.145, lon: 75. 234}, ...]
```
And as a protocol buffer:

```proto
message Stop {
  required string _id = 1;
  required string name = 2;
  required double lat = 3;
  required double lon = 4;
}

message Stops {
  repeated Stop stops = 1;
}
```

I'm fairly certain this is no worse than what was happening before in the GraphQL controller. But it could be a bottleneck and an area of improvement to be aware of.

With this change, we can remove the `GTFSGraphQLController` and any references to `ApiMain` (and eventually the entire `src/main/java/com.conveyal/gtfs/api` folder).

# Other controller removal
The `ProjectController`, `ModificationController` and `TimetableController` have all been fully removed. Their functionality has been replicated and updated in corresponding Next.js API endpoints. Additionally, the `Project` model and persistence methods have been removed. 

Note: a `Scenario` model has not been added to the backend. The backend has everything it needs from getting passed the `modificationIds` directly.


src/main/java/com/conveyal/analysis/models/AnalysisRequest.java
```diff
@@ -178,55 +162,34 @@
...
         // FIXME Job IDs need to be unique. Why are we setting this to the project and variant?
         //       This only works because the job ID is overwritten when the job is enqueued.
         //       Its main effect is to cause the scenario ID to have this same pattern!
         //       We should probably leave the JobID null on single point tasks. Needed: polymorphic task initialization.
-        task.jobId = String.format("%s-%s-%s", projectId, variantIndex, crcValue);
-        task.scenario.id = task.scenarioId = task.jobId;
-        task.scenario.modifications = modifications;
-        task.scenario.description = scenarioName;
-        task.graphId = project.bundleId;
+        task.jobId = String.format("%s-%s", bundleId, crcValue);
```
abyrd, line 178: While we're changing this, we should fix the strangeness mentioned in the FIXME above about jobIds. I think jobId can be nulled out here, and in subsequent lines we can directly set `scenario.id = task.scenarioId = String.format("%s-%s", this.scenarioId, crcValue);` In the PR code the first part of the scenario name is the bundleId. Is this intentional? Should it instead be the scenarioId?

The code in this PR has the first part of the generated (CRC-versioned) scenarioId as the bundleId instead of scenarioId, is this intentional?

abyrd (commented):
Thanks for the write-up. I made some line comments in the code, more general comments are below.

## GTFS API efficiency and binary responses

I think that as you said, in terms of object creation the new stops query is no worse than what GraphQL was already doing, but it's a possible area of improvement to be aware of. We should avoid premature optimization though - contrary to popular belief, creating objects in a modern JVM can be very low cost, as long as the objects are short-lived and don't get promoted out of the eden space. The JVM can even optimize out object creation and convert instance fields to variables (though probably not in a case like this where instances escape and are placed in a collection). Exact garbage collection behavior will vary over time but it is already highly optimized for this kind of case. That said, the Jackson streaming API is also pretty straightforward to use and has to be more efficient. My main concern with either the object mapper or streaming Jackson APIs is that the JSON output is probably being written into an in-memory buffer instead of streamed out the network socket - that's something we could probably improve in a lot of places but shouldn't bother unless we can actually measure the scale of the "problem".

I'm generally open to binary data exchange. But for this purpose I am not a big fan of protocol buffers as opposed to other compact serialization libraries or just writing raw binary data. In any case binary is not necessarily much more compact than text (especially with compression), it might just map more readily into numeric types without conversions. Deflate is gong to turn all the repeated JSON keys into tiny 2-3 bit strings. For example, a latitude to six digits of precision is about 9 characters, and the first three or four of those will be identical for all coordinates. So it might be 6 bytes or less compressed; a double is going to take 8 bytes and may compress poorly unless specially coded to favor compression.

I also wouldn't want to change one part of the API to binary responses and leave the rest in JSON. 

## Order of modifications

The order of the modifications we send to the workers seems to be nondeterministic. It will depend on the order of the Modification records returned by Mongo, which is probably not fixed. The previously used method modificationsForProject also did not place the modifications in any particular order before sending them to the worker.

The modifications are eventually sorted just before they are used on the worker, at Scenario.java L101. From one perspective this is good: the final stage (worker) places them in the right order, so no change in upsteam handling can cause problems. However, this sort order is still not fully deterministic because all modifications of the same type will have the same sort order number and will be "tied" for that position.

Additionally, on the backend before we send the scenario to the workers, we compute checksums of these serialized modifications to create unique scenario version identifiers. If we don't deterministically sort the modifications here before we send them, this checksum (and scenario version) could in theory change every time the same scenario is sent. 

This is not a regression - apparently these modifications were already not being sorted before being checksummed. But we do need to resolve this problem, and now may be a good time since we're changing so much Modification code in this release cycle. 

One possible solution is to treat the arrays of modification IDs as ordered - the record in Mongo actually determines the order, which could in theory be changed by the end user (though in the near-term we'd just apply some deterministic order when storing them to Mongo). The backend could still look up the modifications in a single Mongo bulk query, but would need additional logic to put them back in the order of the original IDs array.

An alternate solution is to automatically put the modifications in some deterministic order before checksumming them, and treat that order as completely distinct from the sort order imposed by the workers. This is probably the better approach - we could just sort them by ID  before computing the digest and be done with it.

## Scenario version naming

However we decide to order the modifications, a CRC32 may not be ideal. With 3000 scenarios, we have a 1 in 1000 chance of a collision (https://preshing.com/20110504/hash-collision-probabilities/). This problem is mitigated by prefixing the bundleId or scenarioId but we should probably use a longer 64 or 128 bit hash. If we do send and check the modification nonces, we could hash or iteratively xor the nonces of all the modifications which is more efficient and deterministic than serializing them all to JSON and calculating a digest as we currently do.

Should we just immediately begin sending an array of nonces for the modifications and calculating scenario version identifiers this way? I guess we'd send two parallel arrays, modificationIds and modificationNonces?

src/main/java/com/conveyal/analysis/BackendMain.java
```diff
@@ -46,7 +46,6 @@ private static void startServerInternal (BackendComponents components, TaskActio
         // TODO remove the static ApiMain abstraction layer. We do not use it anywhere but in handling GraphQL queries.
         // TODO we could move this to something like BackendComponents.initialize()
         Persistence.initializeStatically(components.config);
-        ApiMain.initialize(components.gtfsCache);
```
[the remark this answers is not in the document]
trevorgerhardt: Removed all related GraphQL code and the Gradle dependency.

src/main/java/com/conveyal/analysis/controllers/BrokerController.java
```diff
@@ -170,7 +179,7 @@ private Object singlePoint(Request request, Response response) {
         String address = broker.getWorkerAddress(workerCategory);
         if (address == null) {
             // There are no workers that can handle this request. Request some.
-            WorkerTags workerTags = new WorkerTags(accessGroup, userEmail, project._id, project.regionId);
+            WorkerTags workerTags = new WorkerTags(accessGroup, userEmail, analysisRequest.projectId, analysisRequest.regionId);
```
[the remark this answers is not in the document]
trevorgerhardt: As per discussion, keeping it as project ID for now. The worker's tag themselves automatically with the network ID.

src/main/java/com/conveyal/analysis/persistence/Persistence.java
```diff
@@ -36,7 +35,6 @@
     private static DB db;
 
     public static MongoMap<Modification> modifications;
-    public static MongoMap<Project> projects;
```
abyrd, line 39: I discovered this is still used by the CacheWarmer. Might be better to keep the Project class and this `MongoMap<Projects>` - removing them will break the cluster build and require refactors there.

src/main/java/com/conveyal/analysis/models/AnalysisRequest.java
```diff
@@ -150,20 +152,20 @@
...
-        return Persistence.modifications
-                .findPermitted(QueryBuilder.start("projectId").is(projectId).get(), userPermissions)
-                .stream()
-                .filter(m -> variantIndex < m.variants.length && m.variants[variantIndex])
-                .map(com.conveyal.analysis.models.Modification::toR5)
-                .collect(Collectors.toList());
+    public Scenario createScenario (UserPermissions userPermissions) {
+        QueryBuilder query = "all".equals(scenarioId)
+                ? QueryBuilder.start("projectId").is(projectId)
+                : QueryBuilder.start("_id").in(modificationIds);
```
abyrd, line 160: What is the special significance of scenarioId == "all"?

abyrd (commented):
Just a couple of questions / things to resolve, besides these points everything looks good.

src/main/java/com/conveyal/analysis/models/AnalysisRequest.java
trevorgerhardt: It's the ID that the UI sends for the "All modifications" scenario. Which includes all modifications within a project.

src/main/java/com/conveyal/analysis/persistence/Persistence.java
trevorgerhardt: Added it back 👍

src/main/java/com/conveyal/analysis/models/AnalysisRequest.java
abyrd: Clarified in Slack chat: we now have two implicit scenarios, "baseline" and "all". The "all" scenario always contains every modification for a project, even as that list grows or shrinks. Theoretically there's a potential timing issue here where the backend's view of the database is out of sync with the UI's, where a modification is added or changed in the database between the time the UI saw it and the backend receives this request. We will think through ways to mitigate these risks, but for the time being the use of "all" is certainly no worse than existing mechanisms.

abyrd (approved):
We have reviewed the code and done a few rounds of updates, Java CI tests are passing, and after the UI version update most Cypress tests are passing. The remaining UI test failures can be investigated later - this is ready to merge to dev.

PR #746, opened 2021-08-23
Apply egress table preload logic in regional tasks

abyrd:
In common use cases this should cause egress tables to be calculated before we fire the HandleRegionalEvent and begin routing, allowing better tracking of whether workers are still in a slow preparation phase.

This patched version is currently the r5 dependency for the head of cluster. It was necessary to run a TAUI site and a regional analysis for the whole Netherlands using bicycle egress. Without this patch, the workers will all shut down except for the single point worker and regional job progress will be very slow.

Note that some proportion of workers may survive without this patch; the ones that shut down seem to be those that handle at least one task that never reaches transit and therefore wait to build egress tables until after they've handled a few regional tasks and lowered their shutdown timers.


ansoncfit (approved):
We've seen this work on production for a (non-Taui) regional analysis.

PR #748, opened 2021-09-17
Patches for v6.5 release

abyrd:
I am accumulating any patches here leading up to the September 2021 v6.5 release.

In addition to the items described in detail below, this PR contains various small but important fixes to:
- normalize file extensions on DataSources in FileStorage
- change some API endpoint paths and parameters to match UI usage
- add range and null checking to some API parameters
- not tag workers with projectId (workers can migrate between projects)
- ensure all sidecar files of a shapefile are pulled from S3 before reading it

## In-memory upload FileItems
I ran into a problem creating new opportunity datasets. On upload, a checkState that all uploaded items are readable files failed. This is because unlike Bundle and DataSource uploads, OpportunityDataset uploads were not using the method HttpUtils.getRequestFiles() which customizes the FileUploadItemFactory to ensure even tiny files (like Shapefile projection sidecars) are actually put in a disk file instead of memory. In the case of creating opportunity dataset grids from shapefiles, this wasn't hurting anything because the conversion process explicitly writes the shapefile FileItems out to a temporary directory. Rather than remove the checkState (which actually does check required conditions for other code paths handling uploaded files) I just changed all such code paths to use the same HttpUtils utility method.

Long term it would probably be better to remove the zero-threshold factory, allowing small FileItems (including non-file form fields) to remain in memory, then revise and use moveFilesIntoStorage to explicitly FileItem.write() to a temporary location or directly into FileStorage.

Confusion was caused here by the fact that contrary to its documentation, FileItem does not return null when getStoreLocation is called on an in-memory FileItem. It returns the name of a temporary file where it _would_ save the data if the file were to exceed the size threshold.

## LODES extract progress and DataGroups
We were extracting LODES data in a background task that shows up on `/api/activity`, but not reporting any progress, not putting the resulting layers into a datagroup, and not reporting the WorkProduct. Because we still have the "classic" LODES upload status reports, I considered just bypassing the background task activity reporting, but didn't want to move backward. So with this change we are now reporting both kinds of progress. The new one shows much better incremental progress, so I believe we could now just stop displaying the classic progress updates at the top left of the UI. Observations: I noticed seeing the two side by side that the new progress text transitions from green to blue upon completion, while in the classic progress the entire background flips from blue to green when it finishes. The new progress reporting is clearly better, but the "flip to green" effect of the old progress reporting is more satisfying, and I always feel more instantly aware that the task has finished. This is now reporting a work product of OPPORTUNITY_DATASET with group=true, which the UI does not know how to link to. Suggested approach for the release is to leave both progress mechanisms enabled, but temporarily stop setting the workProduct on the backend so the UI doesn't try to display a broken link.

Note also that OpportunityDatasets have a `sourceId` field which should probably be `dataSourceId` to match AggregationArea, but I'm not changing this now because it might require a database migration.


src/main/java/com/conveyal/file/LocalFileStorage.java
```diff
@@ -62,7 +61,15 @@ public void moveIntoStorage(FileStorageKey key, File sourceFile) {
                         sourceFile.getName());
             }
             // Set the file to be read-only and accessible only by the current user.
-            Files.setPosixFilePermissions(storedFile.toPath(), Set.of(OWNER_READ));
+            try {
+                Files.setPosixFilePermissions(storedFile.toPath(), Set.of(OWNER_READ));
+            } catch (UnsupportedOperationException e) {
+                LOG.info("Could not restrict permissions on {} to POSIX OWNER_READ (probably because OS is not POSIX " +
+                        "compatible)", sourceFile.getName());
+                if (storedFile.setReadOnly()) LOG.info("Marked {} read-only via alternative setReadOnly() method",
```
abyrd, line 69: We should probably just call `storedFile.setReadOnly()` in all cases, and remove the call to `setPosixFilePermissions`. It looks like setReadOnly will adapt to different platforms via the `FileSystem` abstraction.

abyrd: The javadoc does not make it clear whether setReadOnly applies to only the user or everyone. However there is a setWritable (bool, bool) method that allows you to explicitly specify it should not be writable by anyone.

ansoncfit (approved):
Approving for merge to dev

PR #759, opened 2021-11-15
Add error messages for `checkNotNull` in the BrokerController

trevorgerhardt:
This errors will bubble up to the user so it's useful to have specific error messages and not just "NullPointerException"s.


abyrd (changes requested):
I'm definitely supportive of adding clearer error messages. In a situation like this I think constant strings are the usual choice, but maybe the best option here is a reusable range checker that ensures they all look like UUIDs and only generates error messages when something looks wrong. I might push another commit to give an example of what I mean.

abyrd:
I just pushed one commit containing the changes described in my review - this has been tested on normal requests from the UI (which work) and cURL command line requests, which fail when the parameters deviate from UUID or ObjectId format. We'll want to check whether we ever have IDs that don't fit these patterns.

trevorgerhardt:
Changes look good to me 👍

abyrd (approved):

Issue #619, opened 2020-10-09
Methods return parameters they have mutated


ansoncfit:
From @abyrd: We have a few methods that take an object reference as a parameter, have side effects on (mutate) that object, then return the same object reference from the method. The two examples I saw recently were:

regionalAnalysis = Persistence.regionalAnalyses.create(regionalAnalysis);

The create method sets the _id, nonce, updatedAt, createdAt, and updatedBy fields on the instance of a Model subclass passed in. The fact that the method is used to assign to a variable gave me the (incorrect) impression that it doesn't mutate the object passed in.

Second, the method AnalysisRequest#populateTask takes an AnalysisWorkerTask as a parameter, and returns the exact same AnalysisWorkerTask after initializing it. Again this can give the impression that it's copying the object rather than mutating it.

In my opinion the fact that a method has side effects on its parameters (is not a pure function) is a very important thing to know, so I'm adding Javadoc explaining that to both methods. I'd also like to make both of them have void return types and remove the return statements. But first I just wanted to ask if there is a reason or meaning behind this, maybe some convention I don't know.

abyrd:
We  should also add something to our code style documents about this.

PR #763, opened 2021-11-20
Retain additional transfer stop

### LINKED ISSUE #699: Transfers to add-trip modifications at re-used stops
We noticed that transfers from loop routes (e.g. the CTA Orange Line) to add-trip modifications only work in certain directions, apparently determined by which baseline platform the new scenario route stops closest to.
image

Let's check:
* Pattern reboard restrictions
* Transfers to closest stop
* Index of patterns through stops that is not updated by scenarios
* Changes introduced by https://github.com/conveyal/r5/pull/640

ansoncfit:
(Mostly) fixes #699. 

As explained in the updated Javadoc for `retainClosestStopsOnPatterns()`, there may still be cases where optimal transfers are not retained, but this update should correctly handle common cases similar to the ones we have observed. We discussed eliminating the filtering entirely, but there were tractability concerns. The proposed approach here retains at most 1 additional target stop for each transfer source stop, which should have minimal performance impact.

Testing with the Route 350 + added trip example described in #699, the proposed changes indeed allow the intended transfer and have no noticeable differences in other parts of the region or discernible difference in computation time for a regional analysis.
image

Working on this, because transfers are serialized in the network's `transitLayer.transfersForStop`, I realized we need a convention for testing changes that require a new transport network version. I propose using the hash of the R5 commit with the change to test as the `NETWORK_FORMAT_VERSION`. Once a batch of changes is consolidated, in advance of a release we can increment to something more standard (e.g. `nv2`)


src/main/java/com/conveyal/r5/transit/TransferFinder.java
```diff
@@ -181,33 +179,71 @@ public void findTransfers () {
...
 
     /**
-     * Filter down a map from target stop indexes to distances so it only includes those stops that are the
-     * closest on some pattern. This is technically incorrect (think of transfers to a U shaped metro from a bus line
-     * running across the legs of the U, a situation which actually exists in Washington, DC with the
-     * red line and the Q4) but anecdotally it speeds up computation by up to 40 percent. We may want to look into
-     * other ways to optimize transfers (or why the transfers are making routing so much slower) if this turns out to
-     * affect results.
+     * Filters down a map of potential transfer target stops so that, for each pattern, only the closest target stop
+     * is retained; if the closest stop is the source stop itself, the next closest stop is also retained. This
```
abyrd, line 183: Based on the code I think this is actually what happens:
```suggestion
     * is retained; if the closest stop is the source stop itself, the next closest stop is retained instead. This
```

src/main/java/com/conveyal/r5/transit/TransferFinder.java
```diff
@@ -181,33 +179,71 @@ public void findTransfers () {
...
+     * Transferring from the Brown Line southbound at Belmont, the closest stop on the Purple Line Express pattern would
+     * be the same platform. But for this journey, the desired transfer would actually imply switching platforms at
+     * Belmont. Ignoring the source transfer in the search for the nearest stop, then adding it back at the end,
+     * should handle this case.
+     *
+     * So, this filtering can be incorrect in certain rare situations, but anecdotally it speeds up computation by up
+     * to 40 percent. TODO confirm this performance gain with more systematic measurement, and look into other ways to
+     * optimize transfers (or why the transfers make routing so slow).
+     *
+     * @param timesToReachedStops map from target stop index to distance (along the street network) to it
```
abyrd, line 216: We should call attention to the fact that this method has side effects on one of its parameters, which is usually assumed not to be the case.

ansoncfit: https://github.com/conveyal/r5/issues/619

ansoncfit:
We were retaining the source stop, but it was essentially ignored by the router (because it's faster to make a same-stop transfer by just standing at the stop rather than walking to the linked street vertex and back). It's actually important to remove the source stop, to handle cases with certain loop routes -- for example, a journey from Paulina (Brown) to Howard (Purple) on the CTA El. Assume the CTA codes Purple Line Express service as a single pattern (south, around the Loop, and back north), with separate stops for the northbound and southbound boarding platforms at Belmont. Transferring from the Brown Line southbound at Belmont, the closest stop on the Purple Line Express pattern would be the same platform. But for this journey, the desired transfer would actually imply switching platforms at Belmont. Excluding the source stop from the search for the nearest stop should handle this case.

trevorgerhardt:
> I propose using the hash of the R5 commit with the change to test as the `NETWORK_FORMAT_VERSION`. Once a batch of changes is consolidated, in advance of a release we can increment to something more standard (e.g. `nv2`)

Would using a date (or date + commit hash) provide some value by showing us when we updated it last?

abyrd (approved):

Issue #768, opened 2021-11-26
Regional task redelivery failure


abyrd:
In a recent regional analysis, the job was stuck in an incomplete state. Examining the details of the job showed it was on its first retry attempt, but no workers were active to claim and process any tasks. 

These particular tasks were slow to compute. The broker delivered the last batch of tasks, and decided some had been lost when 2 minutes after the last batch was delivered to a worker it hadn't received all the responses. In the mean time the workers had apparently all shut down.

It's possible a few tasks were genuinely lost (e.g. connection lost when worker polling). In this case, maybe the backend's wait to retry is too long (from another point of view, worker shutdown time too short), and all workers to shut down before it retries. 

We should also be aware if the "lost" tasks are among the final ones to be delivered (within N of the end of the list). This is very different than when you've just delivered the last N tasks, but are missing some from much earlier on in the job.

Issue #517, opened 2019-05-05
Report and record errors from regional jobs


abyrd:
Currently if something goes wrong with a regional job, the worker often just logs some errors and fails to return results to the backend. We should report the error immediately and stop the analysis, allowing the user to see what went wrong and fix it.

The errors should be returned with the work result to the broker (without any actual results, since they'd be invalid). The backend would then cancel the job immediately or after it receives a certain number of such errors. It would save the errors with the analysis object so the user (and we) can see them.

abyrd:
Our current "solution" to this is to require people to perform a successful single point analysis before launching a regional analysis with the same settings. However that prevents people from launching many regional analyses for different scenarios in rapid succession, which is a common need. Better error reporting and auto-cancellation for regional jobs would allow us to remove the requirement to first run a single point.

abyrd:
Error display has been added to the current UI, but we need a way to test it. We will add a parameter that causes workers to throw an exception on task N with some specified string as the message. This is in the same vein as com.conveyal.r5.analyst.cluster.AnalysisWorker#TESTING_FAILURE_RATE_PERCENT - we should also check if that's still effective in testing job retries.

PR #769, opened 2021-11-26
Fault Injection ("Chaos Engineering")

abyrd:
This is motivated by testing #517 and #768.  This allows intentionally requesting workers to fail during regional processing by throwing exceptions, dropping tasks, or completely exiting the JVM. It would benefit from some refinements described in the Javadoc, but is definitely useable as-is. For now this only applies to regional analyses but would also make sense for testing single point analysis (with the difference that the starting task and failure percent should probably be ignored).

For example, to test reporting of exceptions on workers, in the request JSON you can add:
```
{
  "injectFault": {
    "failureMode": "THROW_EXCEPTION",
    "message": "This is an exception.",
    "failurePercent": 50,
    "startingAtTask": 100
  }
```
After handling 100 tasks the job should fail, reporting an exception with the supplied message.

The failureMode can be DROP_TASK, THROW_EXCEPTION, or EXIT_JVM. To test job redelivery:
```
{
  "injectFault": {
    "failureMode": "DROP_TASK",
    "failurePercent": 10,
    "startingAtTask": 200
  }
```

This should cause 10% of tasks above task 200 to be dropped, leading to several retries before all results are returned.

Unfortunately these tests are not deterministic and may (with low probability) not even trigger any faults. Reasons and possible refinements are described in the Javadoc.

### DIFF
--- a/src/main/java/com/conveyal/analysis/models/AnalysisRequest.java
+++ b/src/main/java/com/conveyal/analysis/models/AnalysisRequest.java
@@ -5,6 +5,7 @@ import com.conveyal.analysis.UserPermissions;
 import com.conveyal.analysis.persistence.Persistence;
 import com.conveyal.r5.analyst.WebMercatorExtents;
 import com.conveyal.r5.analyst.cluster.AnalysisWorkerTask;
+import com.conveyal.r5.analyst.cluster.ChaosParameters;
 import com.conveyal.r5.analyst.decay.DecayFunction;
 import com.conveyal.r5.analyst.decay.StepDecayFunction;
 import com.conveyal.r5.analyst.fare.InRoutingFareCalculator;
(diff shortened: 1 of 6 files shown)


src/main/java/com/conveyal/analysis/models/AnalysisRequest.java
```diff
@@ -253,6 +260,8 @@ public void populateTask (AnalysisWorkerTask task, UserPermissions userPermissio
         if (task.decayFunction == null) {
             task.decayFunction = new StepDecayFunction();
         }
+        // Intentionally introduce errors for testing purposes.
+        task.injectFault = injectFault;
```
trevorgerhardt, line 264: Should we only allow administrative users to set these?

abyrd: I did think about that. We could strip them out or throw an authorization exception etc. for non-admin users on the backend before the request gets sent to the worker. The alternate point of view is that the system is supposed to be able to handle such errors so people are free to inject problems into their jobs if they actually find this possibility in the source code and want to try it :)

abyrd:
Next task here is to add a manual test to our test document explaining how to check this. This can't be reliably automated yet because the intentional failures are random and nondeterministic.

abyrd:
I have confirmed that when logged in as a user from a non-conveyal (non-admin) group requests including fault injection parameters will be rejected with an exception message. I also confirmed that when logged in as a conveyal user, even if the group is set to something different via `/admin/set-access-group` the user retains admin status and is able to set fault injection parameters. I have created a Notion placeholder test for checking that failed analyses retry or show an error message; that test description will be expanded when I actually perform that test.

trevorgerhardt (approved):
Faults thrown 👍

Issue #765, opened 2021-11-24
RegionalAnalysisController#getScenarioJsonURL is broken. Scenario ID does not exist on embedded request.


trevorgerhardt:
Newly created `AnalysisWorkerTask`s populate their scenario ID at `task.scenario.id`, old tasks populate it at `task.scenarioId`. We still use the old version in the `getScenarioJsonURL` method. This will always fails at the `checkNotNull` assertion with new regional analyses. See here:

https://github.com/conveyal/r5/blob/ebbd53b3079858a7ac7fbb32b0119350a15730b2/src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java#L536-L538

This `scenarioId` is generated when the regional analysis is created and should not be confused with the `Scenario` model's `_id`. See here:

https://github.com/conveyal/r5/blob/ebbd53b3079858a7ac7fbb32b0119350a15730b2/src/main/java/com/conveyal/analysis/models/AnalysisRequest.java#L169-L172

1. We should determine if the old style (`task.scenarioId`) is used anywhere else. 
2. If it is not, we should update the `getScenarioJsonURL` method to use the new style (`task.scenario.id`) and 
3. Run a migration on old regional analyses and put the scenario ID there.
4. Add an end-to-end test that downloads the scenario JSON.

Other related code that should be commented on or updated / deprecated:

https://github.com/conveyal/r5/blob/ebbd53b3079858a7ac7fbb32b0119350a15730b2/src/main/java/com/conveyal/r5/profile/ProfileRequest.java#L171-L179

abyrd:
I just traced through the related code in hopes of recording some more context for posterity, and ensuring we don't introduce any new problems when patching this. 

This change was part of #742 that shifted to sending the backend a list of modifications instead of using integer "variants". The first thing to note is that we're specifically concerned here with RegionalTask, the subclass of AnalysisWorkerTask used in regional analyses (the other subclass being TravelTimeSurfaceTask used for single point analysis). The broken API endpoint is attempting to fetch the scenario for a completed regional analysis, and needs to look up the identifier/filename for that scenario on the RegionalTask within a RegionalAnalysis in the database.

[...]

PR #772, opened 2021-12-07
Fix Download Scenario JSON

trevorgerhardt:
Set the `task.scenarioId` properly. The immediate bug this fixes is that regional analysis scenario JSON files can be downloaded again. The details about why we went with this change can be found here: #765 

Also:
1. Add back the FileStorageFormat JSON.
2. Add suggestion about renaming the `RegionalAnalysis#request` parameter in the subsequent change.

### DIFF
--- a/src/main/java/com/conveyal/file/FileStorageFormat.java
+++ b/src/main/java/com/conveyal/file/FileStorageFormat.java
@@ -15,12 +15,11 @@ public enum FileStorageFormat {
     // SHP implies .dbf and .prj, and optionally .shx
     SHP("shp", "application/octet-stream"),
 
-    // Some of these are not yet used.
+    // Some of these are not yet used. Any file type that can be downloaded through `FileStorage` is required here.
     // In our internal storage, we may want to force less ambiguous .gtfs.zip .osm.pbf and .geojson.
     GTFS("zip", "application/zip"),
     OSMPBF("pbf", "application/octet-stream"),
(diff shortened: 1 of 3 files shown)


src/main/java/com/conveyal/file/FileStorageFormat.java
```diff
@@ -24,6 +22,7 @@
     // Also can be application/geo+json, see https://www.iana.org/assignments/media-types/application/geo+json
     // The extension used to be defined as .json TODO ensure that changing it to .geojson hasn't broken anything.
     GEOJSON("geojson", "application/json"),
+    JSON("json", "application/json"),
```
abyrd, line 25: Should we defer adding this value until we actually use it?

trevorgerhardt: We use it for downloading scenarios in offline mode.

trevorgerhardt: I just noticed the comment above, I will update them accordingly.

abyrd: For future reference we do actually use this. Not the enum value by name in Java code, but com.conveyal.file.FileStorageFormat#fromFilename uses file names to look up the enum values. For now it's only used to set the correct MIME type and only affects local usage. It may break with certain file types where the enum value does not match the extension field, so we'll follow up on this in another issue.

PR #774, opened 2021-12-10
Accept extended route_type codes and map them to modes enum

abyrd:
We used to record warnings for any route_type code over 7 seen in a GTFS feed. The values 11 and 12 have been officially added and we already can interpret TPEG codes up to 1500. Logic has been changed so this will all work with no errors.

It seemed potentially disruptive (and maybe unnecessary) to add separate enum values in our domain model for trolleybus and monorail, so I mapped them to BUS and RAIL.

### DIFF
--- a/src/main/java/com/conveyal/r5/transit/TransitLayer.java
+++ b/src/main/java/com/conveyal/r5/transit/TransitLayer.java
@@ -676,35 +676,33 @@ public class TransitLayer implements Serializable, Cloneable {
 
     public static TransitModes getTransitModes(int routeType) {
         /* TPEG Extension  https://groups.google.com/d/msg/gtfs-changes/keT5rTPS7Y0/71uMz2l6ke0J */
-        if (routeType >= 100 && routeType < 200){ // Railway Service
+        if (routeType >= 100 && routeType < 200) { // Railway Service
             return TransitModes.RAIL;
-        }else if (routeType >= 200 && routeType < 300){ //Coach Service
+        } else if (routeType >= 200 && routeType < 300) { //Coach Service
             return TransitModes.BUS;
-        }else if (routeType >= 300 && routeType < 500){ //Suburban Railway Service and Urban Railway service
+        } else if (routeType >= 300 && routeType < 500) { //Suburban Railway Service and Urban Railway service
             return TransitModes.RAIL;
-        }else if (routeType >= 500 && routeType < 700){ //Metro Service and Underground Service
+        } else if (routeType >= 500 && routeType < 700) { //Metro Service and Underground Service
             return TransitModes.SUBWAY;
-        }else if (routeType >= 700 && routeType < 900){ //Bus Service and Trolleybus service
+        } else if (routeType >= 700 && routeType < 900) { //Bus Service and Trolleybus service
             return TransitModes.BUS;
-        }else if (routeType >= 900 && routeType < 1000){ //Tram service
+        } else if (routeType >= 900 && routeType < 1000) { //Tram service
             return TransitModes.TRAM;
-        }else if (routeType >= 1000 && routeType < 1100){ //Water Transport Service
+        } else if (routeType >= 1000 && routeType < 1100) { //Water Transport Service
             return TransitModes.FERRY;
-        }else if (routeType >= 1100 && routeType < 1200){ //Air Service
+        } else if (routeType >= 1100 && routeType < 1200) { //Air Service
             return TransitModes.AIR;
-        }else if (routeType >= 1200 && routeType < 1300){ //Ferry Service
+        } else if (routeType >= 1200 && routeType < 1300) { //Ferry Service
             return TransitModes.FERRY;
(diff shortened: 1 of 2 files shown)


abyrd:
This has been tested on staging, via branch staging-2012-12 where it is merged with all other outstanding pull requests. I modified a Wichita GTFS feed making most of the routes monorails and trolley busses, with a few TPEG ferries and funiculars thrown in. The bundle passes upload and validation. I again modified the feed and added some out of range supra-TPEG numbers and non-numbers in the fields. These are caught on upload and displayed in the bundle UI (Number Parse Error and Range Error).

The range errors are now displayed as info level (blue) in the UI, probably because we used to generate spurious range errors for these larger values and didn't want that to invalidate the feed. Range errors can probably now be switched back to hard failures - they generally indicate something truly wrong with the input GTFS.

Curiously, a project made with both the high-numbered feed _and_  the bad feed can be analyzed in single-point with no errors. Not sure why yet.

src/main/java/com/conveyal/r5/transit/TransitLayer.java
```diff
@@ -723,8 +721,14 @@ public static TransitModes getTransitModes(int routeType) {
             return TransitModes.GONDOLA;
         case 7:
             return TransitModes.FUNICULAR;
+        // Codes 8, 9, and 10 are unused.
+        // Codes 11 and 12 were officially added to the spec to bring it in line with internal Google usage.
+        case 11:
+            return TransitModes.BUS;  // Trolleybus
+        case 12:
+            return TransitModes.RAIL; // Monorail
```
ansoncfit, line 729: Let's discuss - I'd be inclined to group this with METRO instead of RAIL (which we typically use for long-distance/regional rail).

abyrd: `SUBWAY` could work. I didn't think through the RAIL choice much, believing the monorail code wouldn't be used much and wouldn't really influence our output. Looking at the GTFS spec, subway/metro is defined as being underground. Another possible option is `0 - Tram, Streetcar, Light rail. Any light rail or street level system within a metropolitan area` but I'm not sure if this is strictly for at-grade systems. "Light rail" is sometimes elevated e.g. in Singapore, and the formulation "any light rail _or_ street level system" implies that light rail is often  _not_ at street level. All monorails I'm familiar with serve a comparable function to the Singpore light rail (e.g. Punggol LRT).

ansoncfit: I think the GTFS spec definition for `SUBWAY` should be revised grade-separated rather than underground, but given the current definition, I'd vote for `TRAM`.

abyrd: Good examples, those are definitely monorails as part of a metro system. I should have thought of the Chongqing case. The main problem here is that "monorail" is a guideway or traction technology, not a mode of transportation in the sense "metro" is (grade-separated high capacity). This is a bit like having a separate mode code for "rubber tires" which could then apply to a subset of the Paris Metro lines. 

We should suggest the change from "underground" to "grade-separated" in the GTFS spec if it hasn't already been proposed and/or rejected.

trevorgerhardt: Hmmm. Maybe I forgot the original reasoning and conversation around this, but it looks like we're essentially mapping 11 to 3 and 12 to 2 here, which does not feel right at all...

Why wouldn't we allow trolleybus and monorail to be their own types? 
If it's a usability issue, shouldn't this be addressed in the UI?

ansoncfit: Some background:

R5 has always (since 2015) included `TransitMode` enums corresponding to GTFS `route_type` values 0-7 (see history of https://github.com/conveyal/r5/blob/v6.5/src/main/java/com/conveyal/r5/api/util/TransitModes.java). Filtering for these modes was added to R5 in https://github.com/conveyal/r5/pull/355 and (in 2017), and to the UI in 529 (in 2017, for `route_type` 0-3) and 776 (in 2018, for `route_type` 4-7).

In 2020, the GTFS spec was amended to add `route_type` values 11 and 12 (via https://github.com/google/transit/pull/174). My initial thought was to update the `TransitMode` enums in R5 accordingly, but after reviewing the timeline, I agree with @abyrd's assessment about the potential disruptiveness of doing so. While the mapping approach could require additional debugging effort in the future, we already do the mapping of TPEG modes, and already lack a 1:1 mapping of `TransitMode` to GTFS `route_type` (thanks to `AIR`, though that was admittedly a more recent change)

ansoncfit:
```suggestion
            return TransitModes.TRAM; // Monorail
```

ansoncfit (changes requested):
I think we have consensus around the general approach here, mapping trolleybus to bus and monorail to ~~rail~~ tram

ansoncfit (approved):

Issue #779, opened 2022-01-19
large GTFS feed extents may be innocuous


abyrd:
In #738 and subsequent work for v6.5, we added some methods that check the size of envelopes and throw exceptions if they contain suspect lat/lon values, are too large, or span the antimeridian. These are broadly applied to data source uploads, OSM and GTFS uploads etc. to prevent accidental use of data sets with very large extents (e.g. containing points at 0,0 or on small outlying islands on the opposite side of 180 degrees) which can lead to automatic creation of very large grids.

Once these checks were in place we realized that some GTFS fails the test because it contains a few long distance trains running outside the study area to distant cities. It's possible to manually crop these GTFS files down to a smaller area, but that's an extra step that must be done carefully, and the large-extent GTFS feeds are not necessarily harmful. Transit routing is pretty fast and not sensitive to physical distance, so a few extra stops don't really hurt anything. The real performance hit comes from propagating times to grid cells, which won't happen if those transit stops are outside the study area. We derive our default grid from the OSM data, not the GTFS data (as evidenced by the fact that we sometimes harmlessly upload dummy GTFS in completely different towns when creating test regions). 

There are a couple of ways we could address this problem, not necessarily mutually exclusive:
1. In the case of GTFS (but not OSM or GIS uploads), translate these exceptions into soft GTFSErrors that will be visible in the UI but sill allow people to make bundles out of the feed.
2. Crop the GTFS data down to the analysis area on the fly. This would probably mean removing all stops outside a bounding rectangle or polygon.

The second option would introduce a certain lag as the geographic location of each stop is tested and the full set of stops copied so we can update either disable pickup/dropoff at each one or completely remove them from the network. It might make sense to treat it as a new modification type so it could produce persistent modified networks that are reused on each request.

It's not clear though that the speedup we would obtain is actually worth any increase in complexity. Simply not linking these stops to the grid is enough to eliminate most of their impact on runtime.

PR #780, opened 2022-01-19
Tolerate large GTFS geographic extents

abyrd:
This is an initial response to #779. Certain exceptions are translated to GtfsErrors to prevent them from making the feed unusable.

### DIFF
--- a/src/main/java/com/conveyal/analysis/controllers/BundleController.java
+++ b/src/main/java/com/conveyal/analysis/controllers/BundleController.java
@@ -13,6 +13,8 @@ import com.conveyal.file.FileStorageKey;
 import com.conveyal.file.FileUtils;
 import com.conveyal.gtfs.GTFSCache;
 import com.conveyal.gtfs.GTFSFeed;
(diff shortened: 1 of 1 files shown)


trevorgerhardt:
`checkWgsEnvelopeSize` is also used for OSM extents. Should we continue to throw an error for OSM?

abyrd:
> `checkWgsEnvelopeSize` is also used for OSM extents. Should we continue to throw an error for OSM?

Yes, the strict extents checking was originally intended for OSM data since that automatically determines the default linking grid. For example, upload OSM containing islands on either side of 180 degrees longitude, and it thinks you want a grid spanning the whole Earth. 

The same checks were then reused for GTFS, where the problem isn't as important because GTFS stops outside the OSM area are just not linked to the road. They cause only a marginal performance hit and it can be more trouble than it's worth to manually modify the GTFS.

trevorgerhardt (dismissed):

trevorgerhardt (approved):

Issue #644, opened 2020-11-01
Write user-specified options to TransportNetworkConfig


ansoncfit:
Users should be able to configure a range of options when creating bundles (then have these options read when a worker builds the corresponding network):
* #862 
* #114 
* #641
* Save GTFS shapes (#751)
* Set custom traversal permission labelers (e.g., for #643)
* Add modifications to "bake in" to the network (added in #782)
* Override the default `SpeedConfig` values
* Handle left-hand-drive regions

**Update: The paragraphs here are largely superseded by changes in #862. See more recent comments in this issue.**
It might be worth the effort to allow some of these options to be set in analysis requests, but it's probably sufficient for at least some of them to be set once at the network level.

But note that "network bundle creation" (on the backend) is not the same as network building (on workers). It's not enough to specify assumptions about streets when uploading OSM -- those assumptions need to be saved somewhere and passed along to the worker that actually builds the network.

One idea is to save a new `NetworkBuilderConfig` in the bundle manifest json saved to S3. With that approach, we would no longer rely on passing a default `TNBuilderConfig` in `TransportNetwork.fromFiles()`. 

At first glance, a TNBuilderConfig file is read when building a network...
https://github.com/conveyal/r5/blob/bc69f806a796e4f515106665ad171fccb6c70e42/src/main/java/com/conveyal/r5/transit/TransportNetwork.java#L209

But in practice, we don't provide that config file, so R5 reverts to the default in this block...
https://github.com/conveyal/r5/blob/bc69f806a796e4f515106665ad171fccb6c70e42/src/main/java/com/conveyal/r5/transit/TransportNetwork.java#L254-L256

This current implementation seems subpar, and many of the fields in TNBuilderConfig are for old point-to-point routing instead of analysis.

We could strip out the fallback to a default config, and instead throw an exception if a bundle manifest does not have the config. There's not a backwards compatibility concern, because old workers will continue to have the fallback.

ansoncfit:
The configuration options at hand could just be in the top level object, rather than in a nested config within the manifest. And actually this flat top-level object would be a networkBuildConfig rather than a manifest. We can safely eliminate the concept of bundle manifests; in storage, these are just `[networkId].json` without the terms bundle or manifest hardcoded.

ansoncfit:
Just discussed the idea of allowing network .dat files to be rewritten -- it might be useful for people to be able to save car egress tables, for example, after they have built a network.

ansoncfit:
[...]

PR #782, opened 2022-01-24
Cost rasters and LTS from shapefiles

abyrd:
This combines preliminary work from Q3 2021 and rebases it onto the current development branch. This branch will add the following features:
- Map matching a shapefile to OSM, using specified attribute of the shapefile as biking LTS
- Draping the road network over a raster and assigning costs from the resulting profiles along edges

If we store these to files, then this will require a change to the Network Version of the serialized TransportNetworks.

To facilitate testing and experimentation, these new types of data may be added with the following custom modification types. The source files may be uploaded via the Data Sources tab, and their IDs (copied from the URL) substituted in to the dataSourceId field. 

```
{
  "name": "LTS from Shapefile",
  "r5type": "shapefile-lts",
  "dataSourceId": "61f277140e46af79fd6d5367",
  "ltsAttribute": "lts_ov"
}
```

```
{
  "name": "Elevation from GeoTIFF Raster",
  "r5type": "raster-cost",
  "costFunction": "TOBLER",
  "dataSourceId": "61f277140e46af79fd6d5367"
}
```

```
{
  "name": "Sun cost from GeoTIFF Raster",
  "r5type": "raster-cost",
  "costFunction": "SUN",
  "dataSourceId": "61f277140e46af79fd6d5367"
}
```


abyrd:
This should be ready to merge. The build failure on the last commit is just because this was a fast-forward merge and the artifact already exists in GH Packages.

ansoncfit (commented):
Elevation results look good on staging.

`"outputScale": 1.5`
image

`"outputScale": 3`
image

A few minor/cleanup comments are below. Let me know if you want me to jump in on any of the suggestions.

ansoncfit:
One note when manually copying modifications into a network config .json file to pre-apply costs: the key `r5type` needs to be changed to `type`. 

This note could be added to the Javadoc for the `modifications` field in `TransportNetworkConfig`.

src/main/java/com/conveyal/r5/analyst/scenario/RasterCost.java
```diff
@@ -0,0 +1,120 @@
...
+
+    /** @see com.conveyal.r5.rastercost.RasterDataSourceSampler#setEastShiftMeters(double) */
+    public double eastShiftMeters = 0;
+
+    private CostField.Loader loader;
+
+    @Override
+    public boolean resolve (TransportNetwork network) {
+        if (costFunction == TOBLER || costFunction == MINETTI) {
+            loader = new ElevationLoader(dataSourceId);
```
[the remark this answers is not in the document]
abyrd: This means passing an enum value that has other non-elevation values (SUN) and unfortunately Java doesn't allow defining enums as unions of other enums (so CostFunction could be the union of ElevationCostFunction and SUN). We do want to make this configurable, so I passed in an instance of the function itself rather than the CostFunction enum. This has future thread safety implications, but I documented those.

src/main/java/com/conveyal/r5/rastercost/CostField.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.io.File;
+import java.util.List;
+
+/**
+ * Subclasses or plugin functions will provide costs for hills, sun, noise, pollution etc.
+ *
+ * Interface-wise, a CostField is like a TraversalTimeCalculator with no turn costs, but it can transform an existing
+ * cost. So TraversalTimeCalculator is like a special case of CostField that is always called first with a
+ * parameter of zero, and has the added ability to calculate turn costs.
```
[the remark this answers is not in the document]
abyrd: The comment is a little more abstract, referring to a generalization of both CostFields and our existing TraversalTimeCalculator as functions from a time domain to a time range. This generalization is conceptual but was not actually carried through into code. It's just a way of thinking about applying arbitrarily many costs each time an edge is traversed. From this point of view, we calculate our final traversal times by chaining N functions of this same type: first the traversal time, then the cost fields. Essentially `traversalTime = costFieldN(costField2(costField1(streetTraversalTime(0)))`. The parameter I'm referring to is the input time, which is zero for the first function called in the chain. In practice there are other parameters, but I was thinking of the outputScale factors as being premultiplied into these functions, in practice the functions also have a parameter for which edge is being traversed. And this is the main reason I didn't give them all the same interface in practice: TraversalTimeCalculator also calculates turn costs, and the vague "elegance" of all the functions having the same type was not worth adding turn cost considerations to every cost field like elevation or sun.

src/main/java/com/conveyal/r5/rastercost/ElevationCostField.java
```diff
@@ -0,0 +1,177 @@
...
+ *
+ * In this case tree shade might be able to share the storage model. The value change elements would all be +/- 1 which
+ * would waste some space, but the code reuse might be worth it. And of course this would also allow greyscale shade
+ * instead of binary shade inputs.
+ */
+public class ElevationCostField implements CostField, Serializable {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ElevationCostField.class);
+
+    public static final double DECIMETERS_PER_METER = 10;
```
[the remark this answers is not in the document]
abyrd: This is actually independent of `inputScale`. It's an attempt to keep the storage of elevation data down to a reasonable size since we are keeping a profile for every edge just to facilitate splitting them. Note that vertexElevationsDecimeters is a TShortList and elevationProfilesDecimeters contains short arrays, so all our elevation measurements take half as much space as 32 bit ints or a quarter as much as double precision floats. 2^15/10 meters is enough to store all elevations outside very high mountain passes. This may be considered premature or misguided optimization, its effectiveness should be evaluated.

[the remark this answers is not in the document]
abyrd: I replicated my above comment in the code.

src/main/java/com/conveyal/r5/rastercost/SunCostField.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Override
+    public String getDisplayKey () {
+        return "sun";
+    }
+
+    @Override
+    public double getDisplayValue (int edgeIndex) {
+        return sunProportions.get(edgeIndex / 2);
```
ansoncfit, line 66: For elevation costs, `getDisplayValue` includes the `outputScale`. We could revise sun costs for consistency:

```suggestion
        float sunProportion = sunProportions.get(edgeIndex / 2);
        return (1.0D - sunProportion) + (sunProportion * sunPenalty)
```

src/main/java/com/conveyal/r5/streets/EdgeStore.java
```diff
@@ -1113,6 +1145,30 @@ public void setWalkTimeFactor (double walkTimeFactor) {
...
+            map.put("lengthM", getLengthM());
+            // FIXME we should employ a method like com.conveyal.r5.labeling.LevelOfTrafficStressLabeler.ltsToInt for this
+            int lts =  getFlag(EdgeFlag.BIKE_LTS_1) ? 1 :
+                        getFlag(EdgeFlag.BIKE_LTS_2) ? 2 :
+                        getFlag(EdgeFlag.BIKE_LTS_3) ? 3 : 4;
+            map.put("lts", lts);
+            map.put("pedestrian", getFlag(EdgeFlag.ALLOWS_PEDESTRIAN));
+            map.put("bike", getFlag(EdgeFlag.ALLOWS_BIKE));
+            map.put("car", getFlag(EdgeFlag.ALLOWS_CAR));
+            if (costFields != null) {
```
ansoncfit, line 1164: Here, it would also be helpful to display the `perceivedLengthMultipler` values for walk and bike from the `edgeTraversalTimes`.

Even if we move away from the legacy LADOT generalized costs, other users have set their own custom impedance values for walking and biking (e.g. via #653), and they might want to inspect them.

src/main/java/com/conveyal/r5/rastercost/SunCostField.java
abyrd: I factored the formula out into a private method so it's reused for display and for actually transforming traversal times.

src/main/java/com/conveyal/r5/rastercost/CostField.java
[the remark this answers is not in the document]
abyrd: Coming back to this today, I think I see what you mean. There's a potential flaw in chaining functions this way: the sun factor scales the cost of elevation, not just the cost of the road. Each function could return a factor to be applied to the original traversal time, which could be applied and summed in the caller (the edge traversal function). To make it a bit more general, each function could receive the original traversal time and return an amount to be added to that original time.

src/main/java/com/conveyal/r5/streets/EdgeStore.java
abyrd: I added a `walkTimeFactor` and `bikeTimeFactor` to the attributes for display when the `edgeTraversalTimes` are present on the `EdgeStore`.

abyrd:
Note that this PR addresses #644 - we have started putting more options that affect network build directly in the bundle JSON file which is now called TransportNetworkConfig.

abyrd:
I just gave this another look and updated the comment we discussed in today's meeting. No problems have been encountered testing or demoing these features, so I think it's ready for review/merge.

One thing I noticed while reading through: the older method to load generalized costs (EdgeTraversalTimes) from OSM should be enabled through TransportNetworkConfig, as noted in comments around StreetLayer L287. As it is, it is always enabled but is automatically disabled through a soft-fail when the special OSM tags are not present. This change can be made after this PR is merged though, it's not pressing.

abyrd:
I just added a commit to address an issue @trevorgerhardt pointed out, where due to line thickness some clipped geometries were leaving artifacts on the edge of tiles. In testing this, I realized some other clipping problems exist. Line segments that pass through a tile but have both endpoints outside the tile are not drawn. You see this mostly with transit lines that have no attached shapes and widely spaced stops, and mostly when zoomed very far in. Logic to handle this case can be added later - in its current form we still have good looking tiles in most cases.

abyrd:
A few observations on tile generation: when you zoom in, you can readily see the smoothness of the shapes improve as it fetches more detailed map tiles. This is intended behavior - lower zoom tiles are more simplified since they contain more features. The backend memory use increases while it's building tiles, but this seems ephemeral. After triggering a GC, I see very little ongoing memory impact even with spatial indexes built for 4 feeds. We'll want to check this out with some really big feeds like NL or CH.

abyrd:
Now using JTS intersection operation to clip, and handling cases that produce MultiLineStrings. Cases like the DC blue line exiting and re-entering a tile, and the long straight segments of the LA metrolink lines converging on Union Station now display properly.

src/main/java/com/conveyal/analysis/controllers/WorkerProxyController.java
```diff
@@ -41,59 +48,93 @@ public void registerEndpoints (Service sparkService) {
...
      */
-    private Object proxyGet (Request request, Response response) {
-
-        final long startTimeMsec = System.currentTimeMillis();
-
-        final String bundleId = request.params("bundleId");
-        final String workerVersion = request.params("workerVersion");
-
-        WorkerCategory workerCategory = new WorkerCategory(bundleId, workerVersion);
+    private String getOrStartWorker (String bundleId, String workerVersion, UserPermissions userPermissions) {
```
trevorgerhardt, line 60: This code is mostly a duplicate of the code in the `BrokerController` (in fact it was originally copied from there). Should we abstract this out as a helper method in the broker itself?

```java
   /**
     * Get or start worker.
     */
    public synchronized String getOrStartWorker(WorkerCategory workerCategory, WorkerTags workerTags) {
        String address = getWorkerAddress(workerCategory);
        if (address == null) {
            // There are no workers that can handle this request. Request some.
            createOnDemandWorkerInCategory(workerCategory, workerTags);
        } else {
            // Workers exist in this category, clear out any record that we're waiting for one to start up.
            // FIXME the tracking of which workers are starting up should really be encapsulated using a "start up if needed" method.
            recentlyRequestedWorkers.remove(workerCategory);
        }
        return address;
    }
```

At the very least, we shouldn't leave the replicated `FIXME` in two different places.

src/main/java/com/conveyal/analysis/controllers/GtfsVectorTileMaker.java
```diff
@@ -0,0 +1,293 @@
+package com.conveyal.analysis.controllers;
```
trevorgerhardt, line 1: Should the GtfsVectorTileMaker be moved into the com.conveyal.gtfs package since it's not a controller?

src/main/java/com/conveyal/analysis/controllers/GtfsVectorTileMaker.java
```diff
@@ -0,0 +1,293 @@
...
+
+    /** A cache of spatial indexes of the trip pattern shapes, keyed on the BundleScopedFeedId. */
+    private final LoadingCache<String, STRtree> shapeIndexCache = Caffeine.newBuilder()
+            .maximumSize(MAX_SPATIAL_INDEXES)
+            .expireAfterAccess(EXPIRE_AFTER_ACCESS)
+            .removalListener(this::logCacheEviction)
+            .build(this::buildShapesIndex);
+
+    /** A cache of spatial indexes of the transit stops, keyed on the BundleScopedFeedId. */
+    private final LoadingCache<String, STRtree> stopIndexCache = Caffeine.newBuilder()
```
trevorgerhardt, line 89: Should the shape and stop indexes be moved into the GTFSCache? If they shouldn't, should they at least be extracted into their own file in case they ever want to be used outside of the GtfsVectorTileMaker? Additionally, is there a better way to distinguish between the type of cache data they contain vs what is currently in the GTFSCache? Possibly renaming `GTFSCache` -> `GtfsMapDbCache` and extracting the shape and stop indexes into a `GtfsGeoSpatialInMemoryCache`? I don't think these names distinguish between the two well enough, but I wanted to give an example.

src/main/java/com/conveyal/analysis/controllers/GtfsVectorTileMaker.java
```diff
@@ -0,0 +1,293 @@
...
+
+    /** Produce a single Mapbox vector tile for the given unique GTFS feed ID and tile coordinates. */
+    public byte[] getTile (String bundleScopedFeedId, int zTile, int xTile, int yTile) {
+
+        // It might make sense to unify these two values into a single compound object.
+        STRtree shapesIndex = shapeIndexCache.get(bundleScopedFeedId);
+        STRtree stopsIndex = stopIndexCache.get(bundleScopedFeedId);
+
+        final long startTimeMs = System.currentTimeMillis();
+        final int tileExtent = 4096; // Standard is 4096, smaller can in theory make tiles more compact
```
trevorgerhardt, line 159: Should this be moved into a `private static final int TILE_EXTENT = 4096;`?

src/main/java/com/conveyal/analysis/controllers/GtfsVectorTileMaker.java
```diff
@@ -0,0 +1,293 @@
...
+
+            Point point =  GeometryUtils.geometryFactory.createPoint(tileCoord);
+            point.setUserData(properties);
+            stopGeoms.add(point);
+        }
+        JtsLayer stopsLayer = new JtsLayer("conveyal:gtfs:stops", stopGeoms, tileExtent);
+
+        // Combine these two layers in a tile
+        JtsMvt mvt = new JtsMvt(patternLayer, stopsLayer);
+        MvtLayerParams mvtLayerParams = new MvtLayerParams(256, tileExtent);
```
trevorgerhardt, line 197: Should this be moved into a `private static final int TILE_SIZE = 256;`?

src/main/java/com/conveyal/analysis/controllers/GtfsVectorTileMaker.java
```diff
@@ -0,0 +1,293 @@
...
+            }
+
+            Map<String, Object> properties = new HashMap<>();
+            properties.put("feedId", stop.feed_id);
+            properties.put("id", stop.stop_id);
+            properties.put("name", stop.stop_name);
+            properties.put("lat", stop.stop_lat);
+            properties.put("lon", stop.stop_lon);
+
+            // Factor out this duplicate code from clipScaleAndSimplify
```
trevorgerhardt, line 184: Should we factor this out?
```java
    private static Coordinate projectCoordinate (Coordinate c, Envelope envelope) {
        double px = ((c.x - envelope.getMinX()) * TILE_EXTENT) / envelope.getWidth();
        double py = ((envelope.getMaxY() - c.y) * TILE_EXTENT) / envelope.getHeight();
        return new Coordinate(px, py);
    }
```

src/main/java/com/conveyal/analysis/controllers/NetworkTileController.java
```diff
@@ -0,0 +1,158 @@
...
+        sparkService.get("/:bundleId/tiles/:z/:x/:y", this::getTile);
+    }
+
+    private TransportNetwork getNetworkFromRequest (Request request) {
+        // "bundleId" is also the "graphId" in an `AnalysisWorkerTask`
+        final String bundleId = request.params("bundleId");
+        // final String modificationNonceDigest = request.params("modificationNonceDigest");
+        checkNotNull(bundleId);
+        TransportNetwork network = transportNetworkCache.getNetwork(bundleId);
+        // "61fe817974f6230b0363aae1-8c07ddd4f8bd29ac10a4a109dd27d7b58dabd56c" // elevation only
```
trevorgerhardt, line 85: These appear to be hard coded testing IDs. Should they be removed?

src/main/java/com/conveyal/analysis/controllers/NetworkTileController.java
```diff
@@ -0,0 +1,158 @@
...
+ */
+public class NetworkTileController implements HttpController {
+
+    private static final Logger LOG = LoggerFactory.getLogger(NetworkTileController.class);
+
+    /**
+     * Complex street geometries will be simplified, but the geometry will not deviate from the original by more
+     * than this many tile units. These are minuscule at 1/4096 of the tile width or height.
+     */
+    private static final int LINE_SIMPLIFY_TOLERANCE = 5;
```
trevorgerhardt, line 57: Should `clipScaleAndSimplify` and common static values like `LINE_SIMPLIFY_TOLERANCE`, `TILE_EXTENT`, and `TILE_SIZE` be factored out into a `TileUtils` file?

src/main/java/com/conveyal/r5/streets/StreetRouter.java
```diff
@@ -297,9 +299,12 @@ public StreetRouter (StreetLayer streetLayer) {
         this.timeCalculator = streetLayer.edgeStore.edgeTraversalTimes;
         // If no per-edge timings were supplied in the network, fall back on simple default timings
         if (this.timeCalculator == null) {
-            // TODO one of two things: 1) don't hardwire drive-on-right, or 2) https://en.wikipedia.org/wiki/Dagen_H
+            // TODO either: 1) don't hardwire drive-on-right, or 2) global https://en.wikipedia.org/wiki/Dagen_H
```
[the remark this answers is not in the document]
abyrd: Deferred until after daylight saving time is abolished.

src/main/java/com/conveyal/analysis/controllers/WorkerProxyController.java
abyrd: I definitely agree, this should be factored out into a shared method. In fact, as a best practice we should avoid factoring out this much code even when making a prototype because it then requires us to spot and re-merge the duplicates months down the line.

src/main/java/com/conveyal/analysis/controllers/GtfsVectorTileMaker.java
abyrd: Summarizing discussion on the call today: we will move these indexes into GTFSCache, but leave these spatial index caches as two top-level LoadingCaches within GTFSCache rather than making the spatial indexes fields on the GTFSFeed values managed by the GTFSCache. This is not because we necessarily consider this an ideal design long term, but because 1. it retains the current level of flexibility with respect to eviction policies, allowing the geometry caches to hold less objects and evict them sooner than the feed cache. This is important because we haven't used this in production yet and unlike the GTFSFeeds the spatial indexes have potentially unlimited memory usage. We want to observe memory usage in practice and be able to tune these parameters. 2. it's easier to reason about reference chains and loops when they're not nested, 3. even though lazy loading synchronization would be relatively simple as it would be per-GTFSFeed instance, it's nice to uniformly use the well-audited per-key locking of LoadingCache.

src/main/java/com/conveyal/analysis/controllers/GtfsVectorTileMaker.java
abyrd: Yes, we'll add symbolic constants for all magic numbers. This is just left over from prototyping. These also need some comments explaining that they're the usual standard values, and what they do.

src/main/java/com/conveyal/analysis/controllers/GtfsVectorTileMaker.java
abyrd: Yes, this can be factored out. I think the original code that clones a coordinate and sets its fields was the result of a series of refactors, starting from a version that tried to avoid new object creation. In the end we create a lot of small objects in the vector tile generation, which is not a big deal for contemporary garbage collectors.

src/main/java/com/conveyal/analysis/controllers/NetworkTileController.java
abyrd: Yes, just left over from experiments and prototyping.

src/main/java/com/conveyal/analysis/controllers/NetworkTileController.java
abyrd: Summarizing discussion from today's call: they're not quite static utility methods, as the goal is to reuse the envelope and buffered envelope of a particular tile as fields on an object. So we want to factor them out, but to a concrete class in a vector tile-specific package.

src/main/java/com/conveyal/r5/analyst/scenario/ShapefileLts.java
```diff
@@ -0,0 +1,84 @@
...
+import java.io.File;
+
+import static com.conveyal.file.FileCategory.DATASOURCES;
+
+/**
+ * Custom (experimental) Modification that loads LTS values from a shapefile and matches them to the street edges.
+ * As a custom modification, a model class only exists in R5, there is no separate UI/backend modification class.
+ *
+ * TODO This modification is not effective together with transit - it doesn't cause the egress tables to be rebuilt.
+ * Hmm, actually we don't ever recompute bike egress. Changing the LTS value does not invalidate bike egress tables.
```
trevorgerhardt, line 23: Does this comment need to be addressed? At the very least, is it clear in the documentation provided to the clients that use this feature?

trevorgerhardt (approved):

src/main/java/com/conveyal/r5/analyst/scenario/ShapefileLts.java
abyrd: This may be properly handled because affectsStreetLayer() returns true, but I’d like to retrace how that works to make sure it handles this case. Of course additional documentation will be in order if egress is not always affected by modifications.

PR #788, opened 2022-02-09
Ephemeral merge branch: Cost rasters plus vector tiles

abyrd:
This is a throwaway branch to test raster cost loading together with vector tiles. Please do not make any commits directly to this branch, only merge other branches into it.


abyrd:
For systematically creating this kind of branches: GH Branch protection rules have a branch name _pattern_ so we could allow only merges to branches whose names have certain prefixes. Maybe we should also require linear history on our release branch, it avoids weird problems with the inferred version numbers and generally makes things easier to reason about.

abyrd:
I just added two commits to this branch, even though it was a test merge of two other branches, because we may want to permanently merge those two branches. If not, the first commit is crafted to apply cleanly to the cost-rasters-and-lts branch, with the second adapting it to the vector tiles branch.

PR #789, opened 2022-02-16
Ignore unused egress modes in network preloader

abyrd:
This avoids building huge bike egress tables when that egress mode is specified but no transit is enabled. later during routing, the egress tables should only be retrieved by the PerTargetPropagater which should only be created when transit is in use.


src/main/java/com/conveyal/r5/analyst/NetworkPreloader.java
```diff
@@ -186,7 +184,11 @@ public Key (AnalysisWorkerTask task) {
             // transit is used). See code in TravelTimeComputer for when each is used.
             // Egress modes must be tracked independently since we need to build EgressDistanceTables for those.
             this.allModes = LegMode.toStreetModeSet(task.directModes, task.accessModes, task.egressModes);
```
trevorgerhardt, line 186: It appears that the egress modes are still added to the "all modes" enum set. Do we want that to occur? 

The following snippet could be used in order to create the "all modes" enum set from the filtered egress modes set:

```java
            // Egress modes must be tracked independently since we need to build EgressDistanceTables for those.
            this.egressModes = LegMode.toStreetModeSet(
                // Avoid unnecessarily building egress tables by
                // ignoring spurious egress modes in requests that do not have transit enabled.
                task.transitModes.isEmpty() ? EnumSet.noneOf(LegMode.class) : task.egressModes
            );
            // We need to link for all of access modes, egress modes, and direct modes (depending on whether
            // transit is used). See code in TravelTimeComputer for when each is used.
            this.allModes = LegMode.toStreetModeSet(task.directModes, task.accessModes);
            this.allModes.addAll(this.egressModes);
```

trevorgerhardt (changes requested):

src/main/java/com/conveyal/r5/analyst/NetworkPreloader.java
abyrd: We may still be "over-linking" because I believe the access mode linkages also won't be touched if transit is not used. Linking is usually faster than egress table building so this may not have much impact. We may eventually want to rewrite this so the preloader key has `egressModes` and `linkageModes` instead of `allModes`, but for now I think it's better to stick with this minimally disruptive change.

ansoncfit (approved):
Looks good.

PR #796, opened 2022-03-28
Factor out common vector tile code

trevorgerhardt:
This change factors out common code across the `GtfsVectorTileMaker` and the `NetworkTileController`. It moves many common functions into the `MapTile` util. This was done because many operations are dependent on the `envelope` that is produced by that util. Additionally, the remaining functionality of the `GtfsVectorTileMaker` was boiled down to just creating and querying the caches, so I renamed it `GtfsGeometryCache` and moved it into the `com.conveyal.gtfs` package.

Note: the naming could certainly be improved. Suggestions welcome.


src/main/java/com/conveyal/analysis/util/MapTile.java
```diff
@@ -1,23 +1,74 @@
...
+import org.locationtech.jts.geom.Geometry;
+import org.locationtech.jts.geom.LineString;
+import org.locationtech.jts.geom.Point;
+import org.locationtech.jts.simplify.DouglasPeuckerSimplifier;
 
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Utilities for creating a slippy map tile and generating it's enveloper. Also for encapsulating it's properties that
```
abyrd, line 21: Javadoc should be updated to reflect the fact that this is now vector tile specific, but could eventually be combined with raster tile uses, or be made a subclass of a general tile (not raster or vector specific). It's kind of surprising that we define tile2lat() here and apparently nowhere else in the project - maybe we really don't deal with any raster tiles in the Java code.

src/main/java/com/conveyal/analysis/util/MapTile.java
```diff
@@ -1,23 +1,74 @@
...
+    public static final int DEFAULT_TILE_EXTENT = 4096;
+    public static final int DEFAULT_TILE_SIZE = 256;
+
+    /**
+     * Complex street geometries will be simplified, but the geometry will not deviate from the original by more
+     * than this many tile units. These are minuscule at 1/4096 of the tile width or height.
+     */
+    private static final int LINE_SIMPLIFY_TOLERANCE = 5;
+
+    // Add a buffer to the tile envelope to make sure any artifacts are outside it.
```
abyrd, line 39: I will expand this comment a bit.

src/main/java/com/conveyal/gtfs/GeometryCache.java
```diff
@@ -0,0 +1,55 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.time.Duration;
+import java.util.List;
+import java.util.function.BiConsumer;
+
+/**
+ * Cache sets of geometry in a JTS STRtree using a Caffeine LoadingCache.
+ */
+public class GeometryCache<T extends Geometry> {
```
abyrd, line 19: Should be renamed - it caches spatial indexes, not just geometries. Summarizing discussion: this may be "premature abstraction" (not optimization) since long term we're not certain to maintain two separate spatial index caches. Most reasons that would lead us to maintain two or more of them would also mean they potentially benefit from having different expiration and size eviction parameters. These hard-wired constants are not really constant across all potential spatial index caches, so this doesn't quite match the implied level of abstraction.

It's also not clear whether factoring this code out makes it easier or harder to read. If we retain this abstraction I'll update the Javadoc to include the above comments.

src/main/java/com/conveyal/gtfs/GeometryCache.java
```diff
@@ -0,0 +1,55 @@
...
+                });
+    }
+
+    /** RemovalListener triggered when a spatial index is evicted from the cache. */
+    private void logCacheEviction (String feedId, STRtree value, RemovalCause cause) {
+        LOG.info("Spatial index removed. Feed {}, cause {}.", feedId, cause);
+    }
+
+    /**
+     * Find a List of all geometry within a given envelope.
```
abyrd, line 50: Update comment to clarify that this overselects, and finds things that intersect the envelope (rather than contained within the envelope).

src/main/java/com/conveyal/gtfs/GtfsGeometryCache.java
```diff
@@ -0,0 +1,91 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.time.Duration;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * This class maintains spatial indexes of data from GTFS feeds.
```
abyrd, line 20: It's important that this indexes only _geometries for display_ from the GTFS feeds and does not attempt to provide full indexing for everything available in the GTFS API. Will update Javadoc to clarify.

abyrd (approved):
Comments mostly reflect things already discussed on a call. I will make the planned changes after merge.

PR #798, opened 2022-04-04
Copy street class to split edges

ansoncfit:
When new edges are created at split points for transit stops, the new edges have flags and speeds copied from the original unsplit edge. The streetClass values should be copied in the same place, as proposed in this PR.

To test, check to make sure that streets don't switch to class 4 at transit stops. Before the fix:
image

Based on discussions, @abyrd added another change here, addressing an existing FIXME related to Trove lists for streetClasses and in/out angles.


trevorgerhardt (approved):
Looks good to me 👍

Issue #803, opened 2022-04-15
Handle GTFS Pathways


abyrd:
We are seeing problems from feeds using the GTFS pathways extension. These contain stops with location_type > 1 for which coordinates are optional. This wreaks havoc with the stop spatial index used in generating GTFS vector tiles, but that's just one symptom of the underlying need to handle non-stop, non-station transit nodes that may have no coordinates.

We need to at least gracefully tolerate feeds with these extensions, and eventually fully reconstruct the pathways within the station even when they have no coordinates.

@ansoncfit mentioned MBTA use of pathways in https://github.com/conveyal/r5/issues/642#issuecomment-771647656_

derhuerst:
> […] stops with `location_type` > 1 for which coordinates are optional.

Nit-pick: For entrances/exits (`location_type = 2`), coordinates are required.

The challenge described persists of course.

derhuerst:
I'm curious if and how this would affect other implementations, and therefore trying to understand the problem.

Why is it a problem for routing if generic nodes (`location_type = 3`) and boarding areas (`location_type = 4`) are missing in the spatial index?

ansoncfit:
Fully reconstructing pathways would help address #330

abyrd:
@derhuerst it's not a problem that these nodes are not in the spatial index. The problem was that the nodes without coordinates had coordinate values of NaN which broke the sort algorithm in the spatial index. This didn't just leave them out of the index, it caused a failure when opening the feed.

Indeed my initial comment was wrong with respect to `location_type = 2` - the conditions in the PR should now exactly match the rules in the spec.

Issue #777, opened 2022-01-11
Requirement of `calendar.txt`


RicoFio:
Hi there,

First of all, thanks a lot to all the developers at Conveyal! R5 is an amazing concept!

While I was deploying a development environment locally on my machine, I noticed during test data ingestion that the API requires a `calendar.txt` file to be present in the uploaded `gtfs.zip`. [Here](https://transfer.sh/9RGFvW/gvb-gtfs-20210528-filtered-GVB.zip) the GTFS feed from the [original feed](https://transitfeeds.com/p/ov/814) which I then filtered by agency (GVB) using transitland's CLI and also re-validated using the same CLI.

![image](https://user-images.githubusercontent.com/3929674/148992148-a1eb13b4-fbdc-42a0-8c7c-1ebe82bb50df.png)

However, according to the [specifications](https://gtfs.org/reference/static#calendar_datestxt) `calendar.txt` and/*or* `calendar_dates.txt` can be present (both are valid gtfs feeds):

> Alternate: Omit calendar.txt, and specify each date of service in calendar_dates.txt. This allows for considerable service variation and accommodates service without normal weekly schedules. In this case service_id is an ID.

So I was wondering whether this is a bug or whether there is a rationale behind it?

Cheers,
Riccardo

abyrd:
Hi @RicoFio, this is due to a limitation in our GTFS importer. Each file must be defined as required or non-required. Calendar is the one special case where at least one of calendar or calendar-dates must be present. We'd either have to change the mechanism or mark both these files non-required and add a separate validation module to check that at least one of them was present. These changes really should be made, but the problem was not as visible in the past until we created a better display for these error messages.

For now we have made missing table errors yellow "warnings" instead of hard failures, so you can still use the feed when the file is missing. In the long term this validation should be changed.

abyrd:
Hi @RicoFio, we have a pull request lined up that should make these errors/warnings on the feeds more meaningful.

abyrd:
In `com.conveyal.gtfs.loader.Table`, `CALENDAR` and `CALENDAR_DATES` are now marked as optional.   The method `com.conveyal.gtfs.validator.PostLoadValidator#validateCalendarServices` now checks once the feed is loaded to make sure some services were defined, independent of which file they came from.

PR #804, opened 2022-04-15
Handle GTFS feeds with pathway nodes

abyrd:
This PR contains first steps to address issue #803. The most visible immediate effect is that it avoids spatial index errors from feeds with stops that lack coordinates when generating GTFS vector tiles. The underlying change is gracefully tolerating GTFS feeds that use the pathway extensions, even where we don't yet use those nodes in routing.

The code is already in working condition, but we may want to add more validation or build a full routable model from the pathways in transfers.txt. That could be done in this PR or a subsequent one.

### DIFF
--- a/src/main/java/com/conveyal/gtfs/validator/PostLoadValidator.java
+++ b/src/main/java/com/conveyal/gtfs/validator/PostLoadValidator.java
@@ -1,0 +1,104 @@
+package com.conveyal.gtfs.validator;
+
+import com.conveyal.gtfs.GTFSFeed;
+import com.conveyal.gtfs.error.EmptyFieldError;
+import com.conveyal.gtfs.error.ForbiddenFieldError;
+import com.conveyal.gtfs.error.GeneralError;
+import com.conveyal.gtfs.error.RangeError;
+import com.conveyal.gtfs.error.ReferentialIntegrityError;
+import com.conveyal.gtfs.model.Stop;
+
+/**
+ * Currently we perform a lot of validation while we're loading the rows out of the input GTFS feed.
+ * This can only catch certain categories of problems. Other problems must be found after tables are fully loaded.
+ * These include self-referential tables like stops (which reference other stops as parent_stations).
+ * This could also be expressed as a postLoadValidation method on Entity.Loader.
+ *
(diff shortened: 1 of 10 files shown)


abyrd:
In the interest of getting this fix into the next maintenance release, we will not implement full pathway routing in this PR. In its current form the PR should now at least tolerate non-stop nodes in stops.txt and fully validate whether they must or must not have coordinates based on their types. In order to perform this validation, which requires checking internal references between rows in stops.txt and therefore cannot be applied on the fly as rows are loaded, this PR introduces post-load validation checks that run after all tables are loaded. Using this mechanism, this PR also includes a simple fix for #777: the calendar and calendar_dates tables are both treated as non-required, but a post-load rule adds a general error on the whole feed if no service entries were generated by loading the feed.

src/main/java/com/conveyal/gtfs/validator/PostLoadValidator.java
```diff
@@ -0,0 +1,135 @@
...
+        LocType (int code) {
+            this.code = code;
+        }
+    }
+
+    private enum Requirement {
+        OPTIONAL, REQUIRED, FORBIDDEN
+    }
+
+    private static class Rule {
```
ansoncfit, line 78: At first I thought `Rule` was much more general purpose than the implementation here. Would `LocationTypeRule` or `StopStationRule` help avoid future confusion?

src/main/java/com/conveyal/gtfs/validator/PostLoadValidator.java
```diff
@@ -0,0 +1,135 @@
...
+                                parentStation.location_type)
+                        );
+                    }
+                }
+            }
+            return true;
+        }
+    }
+
+    private static final List<Rule> rules = List.of(
```
ansoncfit, line 127: Similar to comment above -- I initially thought this specified rules that applied to validation methods beyond `validateStopStationConstraints`

ansoncfit (commented):
I added a couple minor renaming suggestions. Otherwise, this looks good to me.

ansoncfit (approved):

PR #807, opened 2022-04-26
Zoom: add to opportunity dataset model

trevorgerhardt:


ansoncfit (approved):
This requires a database migration (adding `zoom`), which we are handling separately.

ansoncfit (approved):
This requires a database migration (adding `zoom`), which we are handling separately.

I tested the new features running this branch and the corresponding UI locally to create a zoom 11 grid. The grid creation worked as expected, and the resulting file was usable after uploading it to production.

PR #809, opened 2022-04-28
Add CodeQL Analysis Github Action

trevorgerhardt:
A free service provided by Github, this action "runs GitHub's industry-leading semantic code analysis engine, CodeQL, against a repository's source code to find security vulnerabilities. It then automatically uploads the results to GitHub so they can be displayed in the repository's security tab. CodeQL runs an extensible set of [queries](https://github.com/github/codeql), which have been developed by the community and the [GitHub Security Lab](https://securitylab.github.com/) to find common vulnerabilities in your code."

### DIFF
--- a/src/main/java/com/conveyal/r5/streets/StreetRouter.java
+++ b/src/main/java/com/conveyal/r5/streets/StreetRouter.java
@@ -35,7 +35,6 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.PriorityQueue;
 
-import static com.conveyal.r5.common.Util.isNullOrEmpty;
 import static com.conveyal.r5.common.Util.notNullOrEmpty;
 import static com.conveyal.r5.streets.LinkedPointSet.OFF_STREET_SPEED_MILLIMETERS_PER_SECOND;
 import static gnu.trove.impl.Constants.DEFAULT_CAPACITY;
@@ -968,8 +967,8 @@ public class StreetRouter {
                 durationSeconds-=seconds;
                 durationFromOriginSeconds -= seconds;
             } else {
-                durationSeconds+=seconds;
-                durationFromOriginSeconds += seconds;
+                durationSeconds += (int) seconds;
+                durationFromOriginSeconds += (int) seconds;
             }
 
         }
--- a/src/main/java/com/conveyal/r5/transit/TransportNetworkCache.java
+++ b/src/main/java/com/conveyal/r5/transit/TransportNetworkCache.java
@@ -6,8 +6,8 @@ import com.conveyal.file.FileStorage;
 import com.conveyal.file.FileStorageKey;
 import com.conveyal.file.FileUtils;
 import com.conveyal.gtfs.GTFSCache;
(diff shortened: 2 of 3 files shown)


src/main/java/com/conveyal/r5/transit/TransportNetworkCache.java
```diff
@@ -207,6 +206,10 @@ private TransportNetwork buildNetworkFromBundleZip (String networkId) {
             ZipEntry entry;
             while ((entry = zis.getNextEntry()) != null) {
                 File entryDestination = new File(dataDirectory, entry.getName());
+                if (!entryDestination.toPath().normalize().startsWith(dataDirectory.toPath())) {
```
abyrd, line 209: I think I see what this is doing (avoiding directory traversal attacks with `..`) but it may not be obvious to a future reader so probably deserves a comment. The exception message also doesn't explain in what way the entry is bad. If this is not intentional, the message could be expanded to say `An item in this zip file would extract to a location outside the intended destination directory. Refusing for security reasons.` which would also serve to explain the code.

abyrd: I see we've got something related in FileStorageKey#checkForDirectoryTraversal, I wonder if they could be combined somehow. On the other hand, this buildNetworkFromBundleZip is deprecated and only used for an "old-format bundle stored as single ZIP file". Another solution might be to fully deprecate it and just delete the code. I don't know how many of these "old-format bundles" we still have around.

src/main/java/com/conveyal/r5/streets/StreetRouter.java
```diff
@@ -968,8 +967,8 @@ public void incrementTimeInSeconds(long seconds) {
                 durationSeconds-=seconds;
                 durationFromOriginSeconds -= seconds;
             } else {
-                durationSeconds+=seconds;
-                durationFromOriginSeconds += seconds;
+                durationSeconds += (int) seconds;
```
abyrd, line 970: Of course there doesn't seem to be any harm in explicitly performing a cast, but what are the security implications of explicitly casting to an int here, and why would we do it when adding but not when subtracting? It seems like the code scanning may latch on to certain details that just match a simple pattern implicated in a known security problem without truly validating whether the problem is present. It's still probably worthwhile to try out CodeQL, but if over the longer term it doesn't seem to target and explain true security problems I wouldn't want to keep using it just out of inertia.

On further thought I'm actually kind of wary of accepting its suggestions. The difference between the subtraction and addition branches in what the tool considers "more secure" code creates a pattern that a future reader may stop and try to interpret - why is it this way, what is it trying to achieve by explicitly casting only in one branch? It's arguable that the automated system is requiring us to degrade the maintainability of the code for no benefit.

abyrd (approved):
Seems interesting. I've observed how it works by triggering another run of the CodeQL action. I'm not yet convinced it's requiring changes that are enough of a net win to let the results block our merges, but certainly worth a try.

abyrd:
Just noting one other thing I looked into - the service is free because this is an open source repo, and as a public repo the Actions minutes it uses should be free.

PR #816, opened 2022-06-14
Update issue templates

trevorgerhardt:
These are the stock templates provided by Github, they are a great start but we may want to customize them to fit our needs.


abyrd (approved):
Agreed, these templates are a good starting point and we can add more detail later.

PR #817, opened 2022-07-06
Clarify third-party support and versioning in README

ansoncfit:


abyrd (approved):
Looks good. The only comment I have is that by "our methods are relatively stable" we mean that we don't change them much over time, but this could be taken to mean "they don't crash too much". Feel free to make additional direct edits without PR review if you can think of a different word that works better there.

PR #824, opened 2022-09-02
Refine spot instance startup

ansoncfit:
When a regional analysis triggers a request to start spot instance workers, if the request would lead the total number of workers to exceed our configured limit (~6x higher on prod vs. staging), the request is totally ignored.

This PR changes that behavior and reduces the likelihood requests will be ignored:
* Reduces the number of instances requested for higher zoom and non-transit analyses
* Adds a backoff function (h/t Zeno's paradox) -- if we are halfway to the limit on the total number of workers, no more than half of the available remaining slots will be requested

Three additional refinements could be considered -- per-access group limits, further reducing the number of instances requested when analyses don't have frequency-based routes (we don't yet expose `hasFrequency()` in a convenient place though), and sizing the spot instance request based on observed tasks per minute.

### DIFF
--- a/src/main/java/com/conveyal/analysis/components/broker/Broker.java
+++ b/src/main/java/com/conveyal/analysis/components/broker/Broker.java
@@ -244,7 +244,8 @@ public class Broker implements Component {
     /**
      * Create on-demand/spot workers for a given job, after certain checks
      * @param nOnDemand EC2 on-demand instances to request
-     * @param nSpot EC2 spot instances to request
+     * @param nSpot Target number of EC2 spot instances to request. The actual number requested may be lower if the
+     *              total number of workers running is approaching the maximum specified in the Broker config.
      */
     public void createWorkersInCategory (WorkerCategory category, WorkerTags workerTags, int nOnDemand, int nSpot) {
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/conveyal/analysis/components/broker/Broker.java
```diff
@@ -257,6 +259,12 @@ public void createWorkersInCategory (WorkerCategory category, WorkerTags workerT
             return;
         }
 
+        // Backoff: reduce the nSpot requested when the number of already running workers starts approaching the
+        // configured maximum
+        if (workerCatalog.totalWorkerCount() * 2 > config.maxWorkers()) {
+            nSpot = Math.min(nSpot, (config.maxWorkers() - workerCatalog.totalWorkerCount()) / 2);
```
abyrd, line 265:
```suggestion
            nSpot = Math.min(nSpot, (config.maxWorkers() - workerCatalog.totalWorkerCount()) / 2);
            LOG.info("Worker pool over half of maximum size. Number of new spot instances set to {}", nSpot);
```

abyrd (dismissed):
Looks good to avoid the problem we've seen where way too many workers are created for a simple walk search, flooding the backend with responses from a single job. Longer term we'll want to more actively manage the worker pool using the components that were already drafted for that purpose.

trevorgerhardt (approved):

PR #827, opened 2022-09-22
Path summaries: return raw data in JSON

trevorgerhardt:
Returning raw path summary data for usage in the UI. Necessary for sorting and converting to other formats.

- [x] Return route & stop IDs along with the names.
- [x] Ensure the changes don't effect regional analysis path outputs.

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/cluster/PathResult.java
+++ b/src/main/java/com/conveyal/r5/analyst/cluster/PathResult.java
@@ -153,16 +153,16 @@ public class PathResult {
      * Wraps path and iteration details for JSON serialization
      */
     public static class PathIterations {
-        public StreetTimesAndModes.StreetTimeAndMode access; // StreetTimesAndModes.StreetTimeAndMode would be more machine-readable.
+        public StreetTimesAndModes.StreetTimeAndMode access;
         public StreetTimesAndModes.StreetTimeAndMode egress;
         public Collection<RouteSequence.TransitLeg> transitLegs;
-        public Collection<HumanReadableIteration> iterations;
+        public Collection<Iteration> iterations;
 
         PathIterations(RouteSequence pathTemplate, TransitLayer transitLayer, Collection<Iteration> iterations) {
             this.access = pathTemplate.stopSequence.access;
             this.egress = pathTemplate.stopSequence.egress;
             this.transitLegs = pathTemplate.transitLegs(transitLayer);
-            this.iterations = iterations.stream().map(HumanReadableIteration::new).collect(Collectors.toList());
+            this.iterations = iterations;
         }
     }
 
@@ -208,21 +208,4 @@ public class PathResult {
             this.totalTime = totalTime;
         }
     }
-
-    /**
-     * Timestamp style clock times, and rounded wait/total time, for inspection as JSON.
-     */
-    public static class HumanReadableIteration {
-        public int departureTime;
-        public int[] waitTimes;
-        public int totalTime;
-
-        HumanReadableIteration(Iteration iteration) {
-            // TODO track departure time for non-transit paths (so direct trips don't show departure time 00:00).
-            this.departureTime = iteration.departureTime;
-            this.waitTimes =  iteration.waitTimes.toArray();
-            this.totalTime =  iteration.totalTime;
-        }
-    }
-
 }
--- a/src/main/java/com/conveyal/r5/analyst/cluster/PathResult.java
+++ b/src/main/java/com/conveyal/r5/analyst/cluster/PathResult.java
@@ -46,7 +46,7 @@ public class PathResult {
      * details. For now, the path template is a route-based path ignoring per-iteration details such as wait time.
      * With additional changes, patterns could be collapsed further to route combinations or modes.
      */
-    private final Multimap<RouteSequence, Iteration>[] iterationsForPathTemplates;
+    public final Multimap<RouteSequence, Iteration>[] iterationsForPathTemplates;
     private final TransitLayer transitLayer;
 
     public static String[] DATA_COLUMNS = new String[]{
--- a/src/main/java/com/conveyal/r5/analyst/cluster/PathResult.java
+++ b/src/main/java/com/conveyal/r5/analyst/cluster/PathResult.java
@@ -1,5 +1,6 @@
 package com.conveyal.r5.analyst.cluster;
 
+import com.conveyal.r5.analyst.StreetTimesAndModes;
 import com.conveyal.r5.transit.TransitLayer;
 import com.conveyal.r5.transit.path.Path;
 import com.conveyal.r5.transit.path.PatternSequence;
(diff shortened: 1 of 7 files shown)


src/main/java/com/conveyal/r5/analyst/cluster/PathResult.java
```diff
@@ -218,19 +213,15 @@ public Iteration(int totalTime) {
      * Timestamp style clock times, and rounded wait/total time, for inspection as JSON.
      */
     public static class HumanReadableIteration {
-        public String departureTime;
-        public double[] waitTimes;
-        public double totalTime;
+        public int departureTime;
+        public int[] waitTimes;
+        public int totalTime;
```
ansoncfit, line 218: Comment needs to be updated: `departureTime` is seconds after midnight; `waitTimes` and `totalTime` are elapsed seconds.

src/main/java/com/conveyal/r5/analyst/cluster/PathResult.java
```diff
@@ -218,19 +213,15 @@ public Iteration(int totalTime) {
...
-            this.departureTime =
-                    String.format("%02d:%02d", Math.floorDiv(iteration.departureTime, 3600),
-                            (int) (iteration.departureTime / 60.0 % 60));
-            this.waitTimes =  Arrays.stream(iteration.waitTimes.toArray()).mapToDouble(
-                    wait -> Math.round(wait / 60f * 10) / 10.0
-            ).toArray();
-            this.totalTime =  Math.round(iteration.totalTime / 60f * 10) / 10.0;
+            this.departureTime = iteration.departureTime;
+            this.waitTimes =  iteration.waitTimes.toArray();
+            this.totalTime =  iteration.totalTime;
```
ansoncfit, line 224: Thanks for cleaning this up.

src/main/java/com/conveyal/r5/analyst/cluster/PathResult.java
```diff
@@ -153,21 +153,16 @@ public enum Stat {
      * Wraps path and iteration details for JSON serialization
      */
     public static class PathIterations {
-        public String access; // StreetTimesAndModes.StreetTimeAndMode would be more machine-readable.
-        public String egress;
+        public StreetTimesAndModes.StreetTimeAndMode access; // StreetTimesAndModes.StreetTimeAndMode would be more machine-readable.
```
ansoncfit, line 156: Comment can be deleted

ansoncfit (changes requested):
This looks good overall. Let's discuss whether it's worth keeping stopId/routeId values -- those might be helpful when trying to debug GTFS.

src/main/java/com/conveyal/r5/analyst/cluster/PathResult.java
trevorgerhardt: This class was superseded by `PathResultSummary.Iteration` class.

src/main/java/com/conveyal/r5/analyst/cluster/PathResultSummary.java
```diff
@@ -0,0 +1,175 @@
...
+        @Override
+        public int compareTo(IterationDetails o) {
+            int diff = this.departureTime - o.departureTime;
+            if (diff != 0) return diff;
+            return this.itineraryIndex - o.itineraryIndex;
+        }
+    }
+
+    /**
+     * Get the stop ID. If it does not exist, it is a new stop and return "new".
```
ansoncfit, line 167:
```suggestion
     * Get the stop ID. If the ID does not exist, the stop is a new one added by a modification, so return "new".
```

src/main/java/com/conveyal/r5/analyst/cluster/PathResultSummary.java
```diff
@@ -0,0 +1,177 @@
...
+     */
+    static class TransitLeg {
+        public String routeId;
+        public String routeName;
+        public int rideTimeSeconds;
+        public String boardStopId;
+        public String boardStopName;
+
+        public String alightStopId;
+        public String alightStopName;
```
ansoncfit, line 114: 👍 Nice to have these Id and Name values in separate fields instead of the prior concatenated strings

ansoncfit (commented):
Looks good to me. One minor suggestion below.

ansoncfit (approved):

src/test/java/com/conveyal/r5/analyst/network/SimpsonDesertTests.java
```diff
@@ -146,6 +146,13 @@ public void testGridFrequencyAlternatives () throws Exception {
         DistributionTester.assertExpectedDistribution(twoAlternatives, travelTimePercentiles);
     }
 
+    /**
+     * For evaluating results from the tests below.
+     */
+    private int[] roundPathTimesToMinutes (PathResult.PathIterations paths) {
+        return paths.iterations.stream().mapToInt(i -> (int) (Math.round(i.totalTime / 60f * 10) / 10.0)).toArray();
```
abyrd, line 153: It makes sense to factor out that repeated stream operation, and round rather than truncating when comparing against a list of estimates in minutes. But what's the thinking behind the multiplication and division by 10? If I'm not mistaken this is rounding to the nearest 1/10 minute, then shifting the decimal point back before truncating, giving a similar result to just truncating the original value except for input values within 0.05 below an integer. If this is the intended behavior the purpose should probably be explained in the javadoc. Also, minor detail: floating point literals in Java are double precision by default, so there are extra type conversions between float and double in this series of multiplications and divisions. In places where we don't need to minimize storage space it's typical to just use doubles (d suffix or no suffix) instead of single-precision floats (f suffix).

trevorgerhardt: I do not know the original thinking behind how or why it was written this way. This code was copied from the original implementation here: 

https://github.com/conveyal/r5/blob/a528745a51c5b5e9a8d975ffb6f94cafe601bfe4/src/main/java/com/conveyal/r5/analyst/cluster/PathResult.java#L230-L232

The `SimpsonDesertTests` were written to compare against those results. But after this change, the API no longer uses the `HumanReadableIteration` class or its results. So instead of keeping that class I just moved the transform to this new method within the `SimpsonDesertTests`.

abyrd: That makes sense. I think you mentioned this change the other day and I now see what you mean, this is just composing the old (now unused) function with the stream operation that was applied to it in the test. The old function was just rounding to the nearest tenth of a minute for display purposes (hence "human readable"). These happened to be the values readily available for comparison in the test. Once the two steps are combined though with no use for human-friendly display rounding, it creates an expression that can be confusing to read later - it had me thinking there must be some subtle trick or nuance to this combination of conversions, multiplication, and division.

Now that we don't need rounded human readable values, what we're really after here is an equality check with tolerance ("epsilon"). The testing framework should supply this, I'll see if I can track it down and replace it.

abyrd: Patch at https://github.com/conveyal/r5/pull/830

PR #838, opened 2022-11-17
Add table of per-edge factors to network when modification requires it

abyrd:
Some of the options for ModifyStreets modifications adjust generalized costs for walking and biking. These options would only work on special networks containing a special generalized costs data table that is not present by default in most networks. Those data tables are currently only created in networks built from very particular OSM data where every way has all of the special tags contributing to the LADOT generalized cost (com.conveyal.r5.streets.LaDotCostTags).

Some users have tried out the walk cost and bike cost ModifyStreets parameters and seen cryptic messages like "bikeGenCostFactor can only be set to values other than 1 on networks that support per-edge factors." Rather than keeping and repeatedly explaining this disappointing/confusing behavior, I decided to change the modification apply() code to simply create the data table in the scenario copy of the network as needed. This means each scenario will have its own (potentially large) generalized cost data table instead of just extending a shared one in the baseline network. This is a rarely used specialty modification, so such less-than-optimal implementation is acceptable at least as a stopgap. The other alternatives would be:
1. Add the table to the baseline network whenever it's any scenario needs to extend it. This breaks a lot of conventions we have about treating loaded networks as read-only, and incurs a lot of extra memory access and pointless multiplication by 1 on every scenario including the baseline.
2. Require the table to be enabled on the base network when it's first built, using a parameter in TransportNetworkConfig. This incurs the same overhead, but respects the immutable character of loaded networks and is an intentional choice by the user.

This works as expected in my own testing.

### DIFF
--- a/src/main/java/com/conveyal/r5/streets/EdgeTraversalTimes.java
+++ b/src/main/java/com/conveyal/r5/streets/EdgeTraversalTimes.java
@@ -3,6 +3,7 @@ package com.conveyal.r5.streets;
 import com.conveyal.osmlib.Way;
 import com.conveyal.r5.profile.ProfileRequest;
 import com.conveyal.r5.profile.StreetMode;
+import gnu.trove.list.array.TDoubleArrayList;
 
 import static com.conveyal.r5.streets.LaDotCostTags.Direction.BACKWARD;
 import static com.conveyal.r5.streets.LaDotCostTags.Direction.FORWARD;
@@ -105,5 +106,14 @@ public class EdgeTraversalTimes implements TraversalTimeCalculator {
     /** see getWalkTimeFactor() */
     public double getBikeTimeFactor (int edgeIndex) {
         return bikeTraversalTimes.perceivedLengthMultipliers.get(edgeIndex);
+    }
+
+    /**
+     * As a neutral starting point for building up generalized costs in modifications, as opposed to starting from
+     * tags is OSM data as it's read in, set all scaling factors to 1 and constant costs to 0.
+     */
+    public void setAllUnity () {
+        walkTraversalTimes.setAllUnity();
+        bikeTraversalTimes.setAllUnity();
     }
 }
--- a/src/main/java/com/conveyal/r5/streets/EdgeTraversalTimes.java
+++ b/src/main/java/com/conveyal/r5/streets/EdgeTraversalTimes.java
@@ -3,7 +3,6 @@ package com.conveyal.r5.streets;
 import com.conveyal.osmlib.Way;
 import com.conveyal.r5.profile.ProfileRequest;
 import com.conveyal.r5.profile.StreetMode;
-import gnu.trove.list.array.TDoubleArrayList;
 
 import static com.conveyal.r5.streets.LaDotCostTags.Direction.BACKWARD;
 import static com.conveyal.r5.streets.LaDotCostTags.Direction.FORWARD;
@@ -81,10 +80,10 @@ public class EdgeTraversalTimes implements TraversalTimeCalculator {
         bikeTraversalTimes.copyTimes(oldEdge, newEdge, bikeFactor);
     }
 
-    // Stopgap to pad out the traversal times when adding new edges
-    public void addOneEdge () {
-        walkTraversalTimes.setOneEdge();
-        bikeTraversalTimes.setOneEdge();
+    /** Pad out the traversal multipliers/costs with neutral values. */
+    public void addOneNeutralEdge () {
+        walkTraversalTimes.addOneNeutralEdge();
+        bikeTraversalTimes.addOneNeutralEdge();
     }
 
     public void setWalkTimeFactor (int edgeIndex, double walkTimeFactor) {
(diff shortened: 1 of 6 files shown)


trevorgerhardt (changes requested):
This looks good overall, and I agree with the premise.

Is `unity` descriptive enough? I am less familiar with this area of work, so I may not have the proper context, but method names of 
`EdgeTraversalTimes.generateNeutralEdges(...)` and `setAllEdgesToNeutral()` might be more appropriate for what they end up doing.

src/main/java/com/conveyal/r5/streets/EdgeTraversalTimes.java
```diff
@@ -106,4 +107,17 @@ public double getWalkTimeFactor (int edgeIndex) {
...
+
+    /**
+     * Factory method returning a newly constructed EdgeTraversalTimes where all scaling factors are 1 and constant
+     * costs are 0. This serves as a neutral starting point for building up generalized costs in modifications,
+     * as opposed to starting from pre-existing generalized costs derived from special purpose OSM tags.
+     */
+    public static EdgeTraversalTimes unity (EdgeStore edgeStore) {
+        var times = new EdgeTraversalTimes(edgeStore);
+        times.bikeTraversalTimes.setAllUnity();
+        times.walkTraversalTimes.setAllUnity();
```
trevorgerhardt, line 119: Additionally, this section could be replaced with
```suggestion
        for (int i = 0; i < edgeStore.nEdges(); i++) {
            times.bikeTraversalTimes.setOneEdge();
            times.walkTraversalTimes.setOneEdge();
        }
```
and the new `setAllUnity` method could be removed.

abyrd:
> Is `unity` descriptive enough?

I guess this depends on perspective. By "unity" I meant an element that behaves like the number 1, emphasizing that some item scales another item to the same size it already is, rather than being a specific quantity. Applying elements of the type EdgeTraversalTimes to elements of the type EdgeStore effectively transforms the EdgeStore to one with different perceived travel times. There's one specific element of EdgeTraversalTimes that has no effect on any EdgeStore, which is the identity element. I almost used the name "identity" for that reason but "identity" seemed too ambiguous. I guess "unity" is not any  better. According to https://en.wikipedia.org/wiki/Identity_element it looks like "neutral element" is a synonym so yes, "neutral" might be more intuitive. Some description of how types behave in terms of algebra can help reason about code and would be helpful for some readers, but that can just be a note in the Javadoc.

src/main/java/com/conveyal/r5/streets/EdgeTraversalTimes.java
abyrd: I think I just had a reflex of aiming for better encapsulation ("tell the object for each mode to set its own internal fields") but the EdgeTraversalTimes and SingleModeTraversalTimes are already essentially friend classes. If it's easier to understand without the extra method I'll remove it. I'll rename the existing method though because "setOneEdge" should be something like "addOneNeutralEdge".

abyrd:
Changes have been made to names and function call nesting as requested. There's also potential to add a test here, of the general functionality of setting generalized costs through modifications, which will indirectly test that the data structure is instantiated when it's missing. I'll re-request review when a test is added.

src/test/java/com/conveyal/r5/analyst/scenario/ModifyStreetsTest.java
```diff
@@ -0,0 +1,96 @@
...
+        int totalGreater = 0;
+        for (int i = 0; i < reachedVertexes.size(); i++) {
+            int baselineTime = reachedVertexes.get(i);
+            int modTime = reachedVertexesWithModification.get(i);
+            Assertions.assertTrue(baselineTime >= modTime);
+            if (baselineTime > modTime) {
+                totalGreater++;
+            }
+        }
+        Assertions.assertEquals(454, totalGreater);
```
abyrd, line 66: Checking against an exact number probably makes the test more brittle with respect to changes in routing. Ideally we'd do more checks that will verify expected outcomes even after unforeseen changes to routing details - like your check that baselineTime >= modTime. The first checks that come to my mind are: when increasing the travel speed, the post-modification results should always reach many more vertices, all reached vertices in pre-modification results should be present in post-modification results, and travel times should all be higher. Given the high speedup factor, all post-modification travel times should be 1/10 of the pre-modification travel times (could relax to "at most 1/3" to make it more flexible).

src/test/java/com/conveyal/r5/analyst/scenario/ModifyStreetsTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+        var scenario = new Scenario();
+        scenario.modifications = Lists.newArrayList(ms);
+
+        var reachedVertexes = getReachedVertexes(network, new Scenario());
+        var reachedVertexesWithModification = getReachedVertexes(network, scenario);
+
+        int totalGreater = 0;
+        for (int i = 0; i < reachedVertexes.size(); i++) {
+            int baselineTime = reachedVertexes.get(i);
```
abyrd, line 59: `reachedVertexes` is a map from vertex ID to the value of the optimization variable at that vertex. This iteration is using positional indexes in as if they are vertex IDs. Most of the get() calls are probably returning -1 which will be equal in both sets of output.

abyrd (commented):
Thanks for adding the test @trevorgerhardt. This is the kind of thing I had in mind - check the effect of the modification itself on routing, which indirectly checks that the data tables are created. I think there are a couple of changes needed (see comments). I'll push a commit with some suggested fixes for your review.

src/test/java/com/conveyal/r5/analyst/scenario/ModifyStreetsTest.java
trevorgerhardt: My mistake! Thanks for correcting that.

src/test/java/com/conveyal/r5/analyst/scenario/ModifyStreetsTest.java
trevorgerhardt: I agree, I wasn't totally sure how to structure this test. Thanks for updating that.

trevorgerhardt (approved):

PR #843, opened 2022-11-28
Avoid setting multiple LTS flags on edges via modifications

abyrd:
Add method that clears all LTS flags before setting a new one. This method is then used in ModifyStreets and ShapefileMatcher. Code comments have been clarified to avoid creating similar problems in the future.

Some refactoring was also done to improve readability (enum value imports and braces around conditional blocks).

This enforces certain behavior in Modifications. It is still possible for the LevelOfTrafficStressLabeler to set multiple flags during initial network build, though it seems to have been designed not to, and LTS data has always been set with this mechanism in the past.  Further checks and normalization should be added to always allow only a single LTS flag, at every place they are ever set. In theory this would be doing extra work. In practice it's a series of bitwise ORs on the same register and should not incur any slowdown. Safety and predictably all but require these extra checks, as we're stuck with a data structure susceptible to representing multiple mutually exclusive values at once.

There should also be a simple test that loads a network and applies an modify-streets modification, checking at each stage that no edge has conflicting LTS values associated with it.

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/scenario/ModifyStreets.java
+++ b/src/main/java/com/conveyal/r5/analyst/scenario/ModifyStreets.java
@@ -20,7 +20,7 @@ import java.util.ArrayList;
 import java.util.EnumSet;
 import java.util.List;
 
-import static com.conveyal.r5.labeling.LevelOfTrafficStressLabeler.intToLts;
+import static com.conveyal.r5.streets.EdgeStore.intToLts;
 
 /**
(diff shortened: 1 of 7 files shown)


abyrd:
@trevorgerhardt I updated this branch to bring in your recently added test, then refactored that ModifyStreetsTest class to test both the generalized cost data tables and the LTS representation, as well as testing that both modifications actually changed the contents of the street layer. This should be ready for final review.

PR #845, opened 2022-12-21
Limit number of error/warning/info messages

abyrd:
Messages are added via a method only when set is below a certain size. If a set reaches the maximum size, a message is added explaining that some messages were omitted for brevity.

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/error/TaskError.java
+++ b/src/main/java/com/conveyal/r5/analyst/error/TaskError.java
@@ -7,6 +7,8 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 
+import static com.google.common.base.Preconditions.checkArgument;
+
 /**
  * This is an API model object for reporting a single error or warning that occurred on a worker back to the UI via
  * the backend. The most common errors a user will see are problems applying scenario modifications, so this provides
@@ -35,6 +37,7 @@ public class TaskError {
     public TaskError(Modification modification, Collection<String> messages) {
         this.modificationId = modification.comment;
         this.title = "while applying the modification entitled '" + modification.comment + "'.";
+        checkArgument(messages.size() <= Modification.MAX_MESSAGES + 1);
         this.messages.addAll(messages);
     }
 
--- a/src/main/java/com/conveyal/r5/analyst/scenario/AddStreets.java
+++ b/src/main/java/com/conveyal/r5/analyst/scenario/AddStreets.java
@@ -70,24 +70,24 @@ public class AddStreets extends Modification {
     @Override
     public boolean resolve (TransportNetwork network) {
         if (allowedModes == null || allowedModes.isEmpty()) {
-           addError("You must supply at least one StreetMode.");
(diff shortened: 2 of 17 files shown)


abyrd:
I'm going to add a few assertions and tests, should be ready in a few minutes.

abyrd:
I'm testing this out and fixing a couple of bugs - needs a bit of work before final review.

abyrd:
This is ready for final review. Tested as follows: create a remove-stops modification, remove several stops from the middle of a route or two that have lots of patterns. This will generate tons of warnings. In the UI I see only 5, with a message that some were omitted.

trevorgerhardt (changes requested):
Looks good overall, but in the Github diff view it appears a lot of the changes were off by 1 space. Was the surrounding spacing incorrect or is the new spacing incorrect?

Related: I had set up `google-java-format` to run on save for all changed lines for a little while and I believe introducing that as part of a common configuration for development (along with "optimize imports" on save) will increase productivity by reducing bike-shedding in the long run. The trade-off is it will increase total lines of code changed in the short term.

abyrd:
Good catch. That's a search and replace mistake. I really wanted to do some intelligent, syntax-aware search and replace but it wasn't working and I ended up doing it textually. I had searched for a string starting with a space (to find word beginnings), and mistakenly replaced it with one not starting with a space. The textual replace had also messed up some other unrelated method calls that happened to have similar names. I really just need to learn to properly use syntax-aware replacement in IntelliJ which would have avoided all these problems.

In principle I'd like to apply uniform formatting rules and regularly check that they're applied. In practice, most sets of formatting rules I've worked with in the past made certain code much messier or harder to read, or just failed to work properly. I'll have to check out `google-java-format` to see how it fares - I imagine it's improved since the last time I tried this kind of thing.

Ah, I see there's an IntelliJ plugin, I'll try that out. https://plugins.jetbrains.com/plugin/8527-google-java-format/ 
When you say you set the formatter to "run on save for all changed lines", is this what you mean (triggered within the IDE)? 

I'd want to ensure all big whitespace changes are applied in separate commits from intentional code commits. Grouping whitespace changes together with code changes in the same commit, or even the same set of commits, can make it hard to review changesets. Maybe we could reformat a few key files in individual commits, and if we like the style we occasionally do project-wide or full-file reformats to get it out of the way.

I'll fix those space issues and update the PR.

trevorgerhardt (approved):

trevorgerhardt:
Yes, I was using that plugin.

To keep formatting changes separate I would definitely be in favor of running a formatter with the adopted format / code style repo wide in a single PR. Subsequently we could adopt a common configuration that would keep that format for all new changes.

abyrd:
I tried it out on a couple of files. It looks good overall but the line width and two-space indent are a significant departure from what we've been using (and what is standard in a lot of other Java projects), and some of the punctuation and wrapping looks less clear. Obviously in the big picture consistency is more important than bikeshedding any one detail, but I'd like to try a couple of other formatters before settling on one. I've made a note to look into that.

abyrd:
Issue created at #846

PR #848, opened 2022-12-28
Single-mode exceptions in isImpassable / Limit launched workers to half of remaining capacity

abyrd:
See discussion in comments on #841 and in code comments. Even when access=no|private, there are frequently exceptions for single modes. We want to default to the preexisting behavior of allowing passage (and rely only on edge traversal permissions) whenever we see exception tags that aren't clearly blocking access.

### DIFF
--- a/src/main/java/com/conveyal/analysis/components/broker/Broker.java
+++ b/src/main/java/com/conveyal/analysis/components/broker/Broker.java
@@ -1,6 +1,5 @@
 package com.conveyal.analysis.components.broker;
 
-import com.conveyal.analysis.AnalysisServerException;
 import com.conveyal.analysis.components.Component;
 import com.conveyal.analysis.components.WorkerLauncher;
 import com.conveyal.analysis.components.eventbus.ErrorEvent;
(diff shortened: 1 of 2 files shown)


abyrd:
I have added another unrelated change to this PR to apply different limits when starting large numbers of spot instances. This gives us a version of R5 with both patches applied for immediate testing.

abyrd:
Trying out the worker limiting on staging: on running a first analysis I see in the logs:
```
2022-12-28 12:12:16,549 [qtp1775504837-33] WARN  c.c.a.c.b.Broker - Request for 249 workers is more than half the remaining worker pool capacity.
2022-12-28 12:12:16,549 [qtp1775504837-33] WARN  c.c.a.c.b.Broker - Lowered to 0 on-demand and 123 spot workers.
2022-12-28 12:12:16,551 [qtp1775504837-33] INFO  c.c.c.EC2WorkerLauncher - Requesting 123 spot workers on graph ID '63aa9bc9c84e4152d58fbab2', worker version 'v6.7' for abyrd@conveyal.com (conveyal)
```

Then, on running a second analysis:
```
2022-12-28 12:13:33,006 [qtp1775504837-30] WARN  c.c.a.c.b.Broker - Request for 249 workers is more than half the remaining worker pool capacity.
2022-12-28 12:13:33,006 [qtp1775504837-30] WARN  c.c.a.c.b.Broker - Lowered to 0 on-demand and 67 spot workers.
2022-12-28 12:13:33,008 [qtp1775504837-30] INFO  c.c.c.EC2WorkerLauncher - Requesting 67 spot workers on graph ID '63ac10c2ac64097d90768f6f', worker version 'v6.7-15-g7a3689e' for abyrd@conveyal.com (conveyal)
```

Note that 67 is more than half of the 123 capacity that should have been remaining after the first run. This is probably because there's a time lag in the backend becoming aware of the newly started workers - not all of them had reported in yet when I ran the second job, so `workerCatalog.totalWorkerCount()` was artificially low. This is a pre-existing problem though, and will only be resolved by a bigger overhaul of the worker pool management.

src/main/java/com/conveyal/analysis/components/broker/Broker.java
```diff
@@ -249,31 +248,49 @@ public void createOnDemandWorkerInCategory(WorkerCategory category, WorkerTags w
      */
     public void createWorkersInCategory (WorkerCategory category, WorkerTags workerTags, int nOnDemand, int nSpot) {
 
+        // Log error messages rather than throwing exceptions, as this code often runs in worker poll handlers.
+        // Throwing an exception there would not report any useful infomation to anyone.
```
ansoncfit, line 252:
```suggestion
        // Throwing an exception there would not report any useful information to anyone.
```

ansoncfit (approved):
Thanks for the revised worker startup logic and node tag checks.

ansoncfit (approved):
Looks good

PR #850, opened 2022-12-29
Increase autoscaling limits for freeform analyses

ansoncfit:


abyrd (approved):
Freeform pointsets have been in use for a while, and these limits directly affect user experience. Bumping these guardrails seems safe as long as we run a few freeform analyses during testing and see no surprises.

PR #854, opened 2023-01-19
Allow UUIDs without hyphens

trevorgerhardt:
We used UUIDs without hyphens for a period in the past. This PR modifies the `BrokerController.checkUuid` method to allow those.

Migrating these IDs would be ideal. But migrating IDs is not simple. We have a bigger DB migration planned in the future and will migrate IDs to a single type then.

### DIFF
--- a/src/test/java/com/conveyal/analysis/controllers/ArgumentValidationTest.java
+++ b/src/test/java/com/conveyal/analysis/controllers/ArgumentValidationTest.java
@@ -7,7 +7,13 @@ import org.junit.jupiter.api.Test;
 import java.util.UUID;
 
 /**
+ * Tests that check how parameters received over the HTTP API are validated.
+ * These validators should be fairly strict about what they accept, and should not tolerate the presence of things
+ * like semicolons or double dashes that indicate attempts to corrupt or gain access to database contents.
  *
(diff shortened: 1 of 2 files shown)


abyrd:
I just added a commit that will check it's actually hex digits in the 32-character case, and a few tests to make sure it works.

trevorgerhardt:
Thanks for the additions. This looks good to me! 

@ansoncfit can you do the final review as the only non-contributor?

src/test/java/com/conveyal/analysis/controllers/ArgumentValidationTest.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+public class ArgumentValidationTest {
+
+    @Test
+    void testIdValidation () {
+        Assertions.assertThrows(IllegalArgumentException.class, () -> {
+            BrokerController.checkIdParameter("hello", "param");
+        });
+        Assertions.assertThrows(IllegalArgumentException.class, () -> {
+            BrokerController.checkIdParameter("Robert'); DROP TABLE Students;--", "param");
```
ansoncfit, line 26: 👍

ansoncfit (approved):
Thanks for the fixes.

PR #856, opened 2023-01-25
Return correct route info for path display

### LINKED ISSUE #855: Single-origin path results return incorrect transit route
When summarizing path details for display in the UI, the index of a routes used in a particular path is mistakenly used as an index into the routes in the network's transit layer at https://github.com/conveyal/r5/blob/v6.8/src/main/java/com/conveyal/r5/analyst/cluster/PathResultSummary.java#L53

The resulting errors are fairly conspicuous: the same route (whichever happens to be first in the transit network) will be displayed as the first leg of every path, another route will be displayed as the second leg of every path, etc., regardless of origin/destination.

ansoncfit:
Fixes #855

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/cluster/PathResultSummary.java
+++ b/src/main/java/com/conveyal/r5/analyst/cluster/PathResultSummary.java
@@ -2,6 +2,7 @@ package com.conveyal.r5.analyst.cluster;
 
 import com.conveyal.r5.analyst.StreetTimesAndModes;
 import com.conveyal.r5.transit.TransitLayer;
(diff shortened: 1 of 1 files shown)


abyrd (approved):
The changes to the logic look correct to me. I have verified that the results are produced and returned to the UI with no noticeable errors, and look superficially correct in the UI itinerary viewer: with the patch applied I see a variety of routes for different itineraries and destinations, while all routes are listed as low integers like 1, 2, or 4 without the patch.

trevorgerhardt:
Are there tests we can add to r5 to test this? It would help with ensuring changes to paths recording are correct in the future.

PR #857, opened 2023-01-26
Convert `OSMCache`'s `getKey` and `cleanId` to be static methods

trevorgerhardt:
`BundleController` now no longer depends on `OSMCache` as a component. Removes a `FIXME`.


abyrd (approved):
Thanks for following up on that FIXME. Many of those were inserted when I was working on something else, noticed some detail like this, but didn't want to mix unrelated changes into a topical changeset (and didn't want to divert attention from the main task). It looks like this couldn't affect anything elsewhere (I just took a quick look at cluster).

PR #858, opened 2023-01-26
Update MongoDB driver to 4.8.2

trevorgerhardt:
MongoDB currently provides documentation for versions v4.3 to v4.8. When making changes, its useful to have available documentation for the driver we are using.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -147,7 +147,7 @@ dependencies {
     }
 
     // Database driver.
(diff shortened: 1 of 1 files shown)


trevorgerhardt:
This update must be done when the `Persistence.java` class is removed.

abyrd:
I just rebased this and noticed that the most recent MongoJack has a transitive dependency on 4.8.1 instead of 4.8.2, so aligned all the transitive dependencies. There seem to be some build problems though.

PR #859, opened 2023-01-27
Dead code elimination: Remove Unused HTTP Endpoints

trevorgerhardt:
HTTP endpoints that only interact with the database are no longer in use here. These Controller endpoints and their methods can be removed.


ansoncfit:
Even though they are unused, these endpoints can serve as helpful reminders for me that some requests need to be directed to the database instead.

PR #861, opened 2023-02-01
Bail out of ways that reference missing nodes

### LINKED ISSUE #851: Check that OSM nodes are not null before checking whether passable
Poorly cropped OSM uploads may include ways that reference nodes that have been cropped out. That was recently observed to trigger:

```
Caused by: java.lang.NullPointerException
        at com.conveyal.r5.streets.StreetLayer.isImpassable(StreetLayer.java:1096)
        at com.conveyal.r5.streets.StreetLayer.loadFromOsm(StreetLayer.java:315)
        at com.conveyal.r5.streets.StreetLayer.loadFromOsm(StreetLayer.java:214)
        at com.conveyal.r5.transit.TransportNetworkCache.buildNetworkFromConfig(TransportNetworkCache.java:270)
```

abyrd:
This should fix #851. It will `break` to the next way rather than `continue` to the next node to avoid making edges that bypass the missing nodes via straight lines.


ansoncfit (approved):
Thanks for the fix.

PR #862, opened 2023-02-03
Pre-build linkages and egress tables for additional street modes via TransportNetworkConfig

abyrd:
This is a user-configurable version of [010970c](https://github.com/conveyal/r5/pull/786/commits/010970c6af2b7250b26290301ec338b48b848a23) in #786.

The user can add `"buildGridsForModes": ["BICYCLE", "CAR"]` to the TransportNetworkConfig to pre-build linkages and distance tables and serialize them in the network. This should alleviate the problem of hundreds of regional workers all building the same distance tables (though not during some scenario applications, which will require additional optimizations).

Currently there is no way to inject TransportNetworkConfig during a bundle upload. We must manually edit the network config JSON file on the backend server (in `~/cache/bundles/networkid.json`) and `aws s3 cp` to overwrite the copy on S3 before the network is built by a single point worker.

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/cluster/TransportNetworkConfig.java
+++ b/src/main/java/com/conveyal/r5/analyst/cluster/TransportNetworkConfig.java
@@ -42,7 +42,7 @@ public class TransportNetworkConfig {
     public List<Modification> modifications;
 
     /**
-     * Additional modes other than walk for which to pre-build a full regional grid and distance tables.
+     * Additional modes other than walk for which to pre-build large data structures (grid linkage and cost tables).
      * When building a network, by default we build distance tables from transit stops to street vertices, to which we
      * connect a grid covering the entire street network at the default zoom level. By default we do this only for the
      * walk mode. Pre-building and serializing equivalent data structures for other modes allows workers to start up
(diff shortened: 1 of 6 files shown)


abyrd:
This appears to be working. After creating a bundle on production I edited its scenario JSON `cache/bundles/63dcf6f7fdc3560d3366d09f.json` to contain `"buildGridsForModes":["CAR"]` then copied it to s3 with `aws s3 cp cache/bundles/63dcf6f7fdc3560d3366d09f.json s3://scenario-editor-prod-bundles/63dcf6f7fdc3560d3366d09f.json`.

I then started two single point analyses in different tabs with worker versions `v6.8` and `v6.8-5-g8a509ae`. Worker logs show the `v6.8` worker computing only WALK data and returning a result quickly. Worker logs on `v6.8-5-g8a509ae` show it computing both WALK and CAR linkages and egress tables (very slowly for CAR).

So existing workers are not bothered by the new field, and new workers build data for the requested modes.

New workers log the mode when they make linkages or egress tables:

```
INFO  c.c.r.s.LinkedPointSet - Linking pointset to street network for mode CAR...
INFO  c.c.r.s.LinkedPointSet - Done. Linked 297k of 297k PointSet points to streets for mode CAR.
INFO  c.c.r.s.EgressCostTable - Computed new stop-to-point tables from 100 of 7k transit stops for mode CAR.
```

ansoncfit (approved):
Thanks for the changes and updated comments.

PR #864, opened 2023-02-28
Optimize allocation management

JustinPihony:
In larger data scenarios the memory allocation scenario for the `put` results in a lot of discarded arrays due to re-sizing. So, this sizes it more intelligently. In the small scenarios it shouldn't matter, but in the larger ones it can make a huge difference.

### DIFF
--- a/src/main/java/com/conveyal/r5/util/TIntObjectHashMultimap.java
+++ b/src/main/java/com/conveyal/r5/util/TIntObjectHashMultimap.java
@@ -17,7 +17,12 @@ public class TIntObjectHashMultimap<V> implements TIntObjectMultimap<V> {
 
     @Override
     public boolean put(int key, V value) {
-        if (!wrapped.containsKey(key)) wrapped.put(key, new ArrayList<>());
+        if (!wrapped.containsKey(key)) {
+            wrapped.put(key, new ArrayList<>(2));
+        } else {
+            ArrayList<V> wrappedList = (ArrayList<V>) wrapped.get(key);
+            wrappedList.ensureCapacity(wrappedList.size() * 2);
+        }
         return wrapped.get(key).add(value);
     }
 


abyrd:
Hi @JustinPihony. Indeed, as the ArrayLists grow their backing arrays will be discarded and replaced with larger ones. However, if adding an item will cause the ArrayList to exceed its capacity, by default the backing array is increased by 50%. This occurs when `java.util.ArrayList#grow(int)` calls `java.util.ArrayList#newCapacity` whose Javadoc states: *Returns the current capacity increased by 50% if that suffices.*

I believe the effect of your PR is to change this growth factor from 1.5 to 2. Can you confirm whether this was the intended outcome, or were you working under some other assumption about how ArrayLists grow?

We should consider that creating and discarding small objects can also be a very low-cost operation using the JVM's generational garbage collector. The old wisdom about avoiding small object creation is often not valid on current runtimes. 

A slight-to-moderate reduction in the amount of allocations (from expanding by 2x instead of 1.5x) could also be neutralized by creating poorer locality via a greater number of gaps in the memory layout, as many of the array slots wouldn't be used.

The default `ArrayList#newCapacity` method also contains logic to make it overflow-aware. We should be cautious about replacing such heavily tested and reviewed code with something simpler unless we have a solid reason.

I would be hesitant to supply and maintain our own logic for growing ArrayLists unless this can be shown to be consistently and measurably more performant than the default ArrayList implementation. Even then I'd only want to change this if the difference had a clearly significant impact that was not specific to certain situations or data sets.

Some additional context: this method appears to be used only in the StreetRouter, but `TIntObjectHashMultimap#put` would indeed be called heavily during the street routing phase of a search.

JustinPihony:
I had misread the underlying Java code and had read it as incrementing +1 each time. Closing this out

PR #867, opened 2023-03-15
Update build.gradle for gradle 8.x (and future 9.x)

abyrd:
This is an alternative approach to #866.

Updated build.gradle to use newer versions of plugins that are compatible with Gradle 8.x.
Also fixed deprecation warnings for Gradle 9.x and switched to gradle dependency caching built in to setup-java action.
See further discussion on https://github.com/conveyal/r5/pull/866#issuecomment-1469820507

I have tested all the build tasks we normally use.


ansoncfit (approved):
Thanks, this looks good.

PR #868, opened 2023-03-16
Grow envelope from projected coords

### LINKED ISSUE #833: Coordinate rounding error when checking projected shapefile upload
The check at https://github.com/conveyal/r5/blob/44c683941de3035bb4f5de0d60e71bbba27db29f/src/main/java/com/conveyal/analysis/datasource/ShapefileDataSourceIngester.java#L49-L51 recently failed when a user uploaded a shapefile of selected block groups in southern Louisiana (projection NAD 83).

Envelope for the shapefile reader has `maxy = 30.212154999179035`, while `((Geometry) f.getDefaultGeometry()).getEnvelopeInternal()` yields `maxy = 30.21215499917904`. Because of this apparent rounding difference, the feature's envelope is not contained by the shapefile reader's envelope.

abyrd:
Fixes #833. I believe the true point of the assertion was not to check the envelope, but to provide a sink for the features in the shapefile, looping over all of them such that the whole file is decoded and projected and ensuring it's a valid usable file.

This did highlight that the envelope coming from `reader.wgs84Bounds()` does not necessarily contain all the points (particularly if the axes of the source projection are not parallel to those of WGS84), so I just built up the WGS84 envelope using the decoded and transformed points.

### DIFF
--- a/src/main/java/com/conveyal/analysis/datasource/ShapefileDataSourceIngester.java
+++ b/src/main/java/com/conveyal/analysis/datasource/ShapefileDataSourceIngester.java
@@ -44,24 +44,25 @@ public class ShapefileDataSourceIngester extends DataSourceIngester {
         try {
             ShapefileReader reader = new ShapefileReader(file);
             // Iterate over all features to ensure file is readable, geometries are valid, and can be reprojected.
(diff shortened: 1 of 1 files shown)


ansoncfit (approved):
Thanks for the fix.

Issue #792, opened 2022-03-10
Log an error when loading GTFS with duplicated stop_id values


ansoncfit:
A user uploaded a GTFS feed produced by another vendor, and it apparently had a few rows in `stops.txt` representing different stops with the same `stop_id`. Our validation did not flag any warnings/errors, so it was not immediately obvious why routing results were strange.

ansoncfit:
Based on https://github.com/ipeaGIT/r5r/issues/237, we may also want validation checks for duplicate `route_id` values.

ansoncfit:
In `Stop`, we could check that `feed.stops.get(s.stop_id)` returns `null` before calling `feed.stops.put(s.stop_id, s)`, and record a `DuplicateStopError` if a stop with the given `stop_id` has already been recorded.

PR #869, opened 2023-03-16
Check duplicate and missing GTFS entity keys

abyrd:
Addresses #792. A method is defined and applied for all entity types with simple string IDs, including stops and routes.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -125,10 +125,8 @@ repositories {
     // Put Open Source Geospatial before Maven Central to get JAI core, see https://stackoverflow.com/a/26993223
     maven { url 'https://repo.osgeo.org/repository/release/' }
     mavenCentral()
-    // TODO review whether we really need the repositories below
+    // Polyline encoder 0.2 is now in Maven repo
     maven { url 'https://maven.conveyal.com' }
-    // For the polyline encoder
-    maven { url 'https://nexus.axiomalaska.com/nexus/content/repositories/public-releases' }
 }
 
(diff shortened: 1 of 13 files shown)


trevorgerhardt (approved):
Looks good to me 👍

abyrd:
I just made a few different defective feeds with missing and duplicate stop and route IDs. Everything works as expected, but this revealed another preexisting problem: when there are problems with references to entities like stops, later stages in the import can fail hard with null pointer exceptions. This aborts the entire import and all we see is the NPE instead of the detailed error messages. I made the pattern-finding stage of the import conditional on there being no severe errors recorded (including referential integrity). This had the intended effect: I now see detailed error messages for all the problems instead of one cryptic NPE stack trace.

I noticed that `DuplicateKeyError` has only medium severity, when it might be considered a sign of very bad input data. But this change was made in 2021, a year more recently than the rest of the file, as part of reverting some other problematic change. I'm going to assume for now that we set this to medium for a good reason (related to some real-world input files).

abyrd:
CodeQL build is failing for the same reasons discussed at https://github.com/conveyal/r5/pull/872#issuecomment-1473656401

abyrd:
Build problems were resolved by copying the polyline encoder v0.2 into the Conveyal Maven repo and removing the axiomalaska repo from our gradle build config.

ansoncfit (approved):
Thanks -- it will be helpful to see relevant error messages rather than just the NPEs at later import stages.

Issue #844, opened 2022-12-15
Check grid size when converting from uploaded spatial data


ansoncfit:
We check the number of grid cells implied by analysis requests: https://github.com/conveyal/r5/blob/43068e168138bc5636d895861209c1c0049e0f1a/src/main/java/com/conveyal/analysis/models/AnalysisRequest.java#L274-L287

We should conduct a similar check for uploaded spatial data. Otherwise, users can upload data with large geographic extents at a high zoom level, leading to linkages with tens of millions (or really an unbounded number) of points, which have been observed to stall regional analyses.

Should we move the existing `checkGridSize` method from `AnalysisRequest` to `WebMercatorExtents`, where it could be called for both analysis requests and data uploads?

ansoncfit:
Closed by #872

Issue #871, opened 2023-03-16
Polyline Encoder does not exist


MatthewLehane:
When building r5 on my local machine, I have an error relating to the polyline encoder:

https://nexus.axiomalaska.com/nexus/content/repositories/public/

Looks like this page is down or does not exist anymore although it is a dependency for r5? I am unable to access it. Not sure if this is just an issue on my end, I understand you do not provide help to third party installs, I understand if so -

abyrd:
Hi @MatthewLehane, thanks for the report. We are also running into this issue in our automated build. It looks like their repo server is up (see http://nexus.axiomalaska.com/nexus/content/repositories/) but returning 500 error response codes for resources inside the repo.

This is the PR where we patched the polyline-encoder to more recent Geotools and JTS versions, yielding at 0.2 release:
https://github.com/axiom-data-science/polyline-encoder/pull/1
This release was on Axiom's repo. Let's wait a day or two and see if it becomes available again. If not, we can contact them and coordinate.

Digging a bit deeper, we only need this repository to get the dependency `com.axiomalaska:polyline-encoder`, which is only used in `com.conveyal.r5.util.EncodedPolylineSerializer`, which is only used in `com.conveyal.r5.transitive.TransitivePattern` line 23. This is then only used in `com.conveyal.r5.transitive.TransitiveNetwork`, which in turn we only use in `com.conveyal.r5.analyst.cluster.AnalysisWorker#saveTauiMetadata`. So all of these classes and this dependency are only used for one very specific workflow that no one outside Conveyal is likely to ever use. 

You can simply delete EncodedPolylineSerializer and remove the dependency and the repo from the Gradle config, and the whole project should work fine for most use cases.

abyrd:
I have manually copied the artifacts over from the axiomalaska Maven repo to the Conveyal Maven repo. We usually don't use this repo anymore, but it can serve as a source for this artifact until / unless we have another solution. If you just remove the axiomalaska repo from build.gradle (as in https://github.com/conveyal/r5/pull/869/commits/8545de9f2ee12f240e910d42a944c96e57459348) the build should complete.

PR #872, opened 2023-03-17
Always checkGridSize when creating grids

abyrd:
Addresses issue #844. All Grid and WebMercatorExtents constructors / factory methods will now hit the zoom and extents checks. Tests updated to use acceptable zoom levels. Comments in code further explaining the approach.

As far as I know every place we use Grid and WebMercatorExtents is legitimately subject to these restrictions. We should think about whether there are legitimate places we need to use other zoom levels, though I think the cell-count restrictions would be valid everywhere (retina raster map tiles should never exceed 512x512).

If we used WebMercatorExtents to make map tiles for example, we'd need an exception to the zoom level restrictions, but our VectorMapTile class does not make use of WebMercatorExtents. As an aside, WebMercatorExtents contains its own methods for computing tile2lon and tile2lat which are an exact copy of those in the `com.conveyal.osmlib.display.WebMercatorTile` utility class. Some deduplication would be possible but these are tiny functions copied verbatim from the OSM wiki.

### DIFF
--- a/src/main/java/com/conveyal/analysis/models/AnalysisRequest.java
+++ b/src/main/java/com/conveyal/analysis/models/AnalysisRequest.java
@@ -28,9 +28,6 @@ import java.util.stream.Collectors;
  * sends/forwards to R5 workers (see {@link AnalysisWorkerTask}), though it has many of the same fields.
  */
 public class AnalysisRequest {
-    private static int MIN_ZOOM = 9;
-    private static int MAX_ZOOM = 12;
-    private static int MAX_GRID_CELLS = 5_000_000;
 
     /**
      * These three IDs are redundant, and just help reduce the number of database lookups necessary.
@@ -207,7 +204,6 @@ public class AnalysisRequest {
         // TODO define class with static factory function WebMercatorGridBounds.fromLatLonBounds().
         //      Also include getIndex(x, y), getX(index), getY(index), totalTasks()
         WebMercatorExtents extents = WebMercatorExtents.forWgsEnvelope(bounds.envelope(), zoom);
-        checkGridSize(extents);
         task.height = extents.height;
         task.north = extents.north;
         task.west = extents.west;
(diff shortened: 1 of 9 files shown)


abyrd:
The build is failing for CodeQL because Gradle can't find the polyline encoder dependency, as mentioned in #871. We need to fix that broken repo reference anyway, but I do wonder why this action doesn't have a cached copy of the dependency like our regular build does. I've noticed a lot of actions will include a hash the build config in the cache key, which seems overly restrictive to me. If the directory being cached is a true cache, it should be reusable even after the build config changes. I know there's some system for partial cache key hits which I need to read up on - maybe that was meant to come into play in this action.

src/main/java/com/conveyal/analysis/models/AnalysisRequest.java
```diff
@@ -271,21 +267,6 @@ public void populateTask (AnalysisWorkerTask task, UserPermissions userPermissio
         }
     }
 
-    private static void checkGridSize (WebMercatorExtents extents) {
```
ansoncfit, line 274: 👍 Moved to `WebMercatorExtents`

I did want to check whether this breaks an implicit dependency hierarchy we wanted to maintain -- not relying on "analysis server" classes (in this case, `AnalysisServerException`) in R5 classes.

src/main/java/com/conveyal/r5/util/EncodedPolylineSerializer.java
```diff
@@ -11,9 +11,15 @@
...
- * so is pulling in extra dependencies and requiring conversions (toLegacyLineString).
+ * Serialize JTS LineString to Google encoded polyline.
+ *
+ * This class is the only use of dependency com.axiomalaska:polyline-encoder, and it is only used in
+ * com.conveyal.r5.transitive.TransitivePattern, which is in turn only used in
+ * com.conveyal.r5.transitive.TransitiveNetwork, which is in turn only used in
+ * com.conveyal.r5.analyst.cluster.AnalysisWorker#saveTauiMetadata.
+ * That dependency has required maintainance on a few occasions and is hosted at a repo outside Maven Central which has
+ * become unavailable on a few occations. We should evaluate licence compatibility (LGPL) for copying it into this repo
+ * ("vendoring" it).
```
ansoncfit, line 22: Should this comment be updated now that the polyline encoder is in the Conveyal Maven repo?

src/main/java/com/conveyal/analysis/models/AnalysisRequest.java
abyrd: I did hesitate to use these exceptions given the relationship between the packages, but the overriding consideration was keeping the code mostly unchanged. We've had situations in the past where error messages became invisible or more cryptic to end users because of how they are mapped or interpreted as they pass up from the HTTP server to the UI. So to keep changes minimal and avoid such problems I left the types the same.

ansoncfit (approved):
👍

PR #880, opened 2023-05-30
Fix an NPE when a task does not exist

trevorgerhardt:
Handle the case when `taskScheduler.getTaskForUser` returns `null` so that we to prevent `NullPointerException`s which we've seen in the logs from time to time.

### DIFF
--- a/src/main/java/com/conveyal/analysis/controllers/UserActivityController.java
+++ b/src/main/java/com/conveyal/analysis/controllers/UserActivityController.java
@@ -58,6 +58,10 @@ public class UserActivityController implements HttpController {
         UserPermissions userPermissions = UserPermissions.from(req);
         String id = req.params("id");
         Task task = taskScheduler.getTaskForUser(userPermissions.email, id);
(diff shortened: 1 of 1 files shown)


abyrd (approved):
Looks good to eliminate the log noise. Bigger picture, I do notice that we've got parameter validity checks split across the internal programming API (`if (tasks == null) return false`) and here in the HTTP API, with the innermost method only communicating boolean success/failure. Since the `ACTIVE` check is already "up" here in the HTTP handler we might as well put the `null` check here for now, but it might make sense to check everything at the innermost layer, enforcing those requirements and reporting causes on any call. Not just in this one method - we don't really have a uniform policy or approach for this, but we could start applying one the next time we're refreshing the background task handling or similar things called by the HTTP API. Java does not make it pleasant to return rich error types (without going overboard with `Optional<enum DeleteTaskError>` types for every single method) so this probably entails throwing meaningful exception types or at least generic exceptions with meaningful messages from all the innermost methods.

Issue #714, opened 2021-03-22
Monte Carlo tests fail


abyrd:
Some of the Simpson Desert tests occasionally fail in GH actions test runs. This is probably because they test the closeness of our Monte Carlo results to theoretical results, and there's always some probability that the MC results will be way off. For reproducible testing we could seed all our random number generators but arguably that reduces thoroughness, and in any case small changes to routing could still change the order in which numbers are produced and cause the tests to fail again. Maybe we should just use really high numbers of MC draws on these tests.

abyrd:
Note that some of the slowness in these tests is due to building histograms at every destination, even though the test only looks at one of them. Removing that behavior would probably speed them up significantly, making it more reasonable to spend more time on more MC draws.

ansoncfit:
Recent tests, including with the frequency-heavy network of Sao Paulo, prompted me to think about this again. Letting users toggle on deterministic seeding would be straightforward to implement (e.g. using a similar approach to the one in the multi-criteria router, at https://github.com/conveyal/r5/blob/v6.9/src/main/java/com/conveyal/r5/profile/McRaptorSuboptimalPathProfileRouter.java#L120-L123) and would help users resolve a common headache when they are doing scenario comparisons. We could still recommend networks with frequency-based routes be analyzed with fully randomized schedules first, to get a sense of the noise/uncertainty.

abyrd:
We discussed this again recently. Results are expected to converge on stable values with an adequate number of MC draws. In the context of these stochastic methods, there does not seem to be a legitimate use for fixed seeds. Any use would amount to an illusion of artificial precision and could lead to inadvertent cherry-picking of results.

If expectations for stable results are not met, there are two main explanations:
1. The number of MC draws is just too low. Trying to simulate a truly frequency based line without adequately sampling the parameter space of headways.
2. The modeler is assuming constraints on the line that are not included in the scenario. Departure times are assumed to be synchronized to other routes or specific clock times. Scenario should be updated to use exact-times or phasing.

For tests, the solution is probably to increase the number of MC draws until they pass reliably. They would remain by nature deterministic but the probability of failure can be dropped until it essentially never happens.

For regular use we should provide guidance on exact-times, phasing, and number of MC draws, and explain clearly in documentation how and why these stabilize results.

ansoncfit:
[...]

PR #881, opened 2023-06-13
Allow setting seed for frequency offsets

ansoncfit:
Addresses #714: with frequency-based routes, users are often distracted by spurious changes in access from random schedule differences between scenarios. 

This PR allows "locking" schedules by setting the seed used for randomizing frequency offsets to a deterministic value for each origin (namely, a multiple of `fromLat`, as is done in the multi-criteria router: see `FrequencyRandomOffsets`). It also adds a field to the analysis request to activate this mode (see `AnalysisRequest`). With `lockSchedules: true`, a given origin will use the same frequency offsets across scenarios, so resulting changes in access should be systematic ones. This PR also includes other minor changes for related tests and documentation.

To test, fetch isochrones repeatedly along an infrequent frequency-based route. Without `lockSchedules`, the isochrone should "waver." With this PR and `lockSchedules: true` (or using build `v6.9-7-ged8378e`, which hardcodes the logic to avoid needing to start a backend server on this branch), there should be no variability with repeatedly fetched isochrones.

In documentation, we should encourage users to run analysis without locking schedules first, to get a sense of variability. If certain corridors show large variability, they may want to use phasing. We've discussed other longer-term changes related to optimized schedules, but the approach proposed here is a minimally invasive one that addresses multiple user requests.

### DIFF
--- a/src/main/java/com/conveyal/r5/profile/FrequencyRandomOffsets.java
+++ b/src/main/java/com/conveyal/r5/profile/FrequencyRandomOffsets.java
@@ -1,5 +1,6 @@
 package com.conveyal.r5.profile;
 
+import com.conveyal.analysis.models.AnalysisRequest;
 import com.conveyal.r5.transit.TransitLayer;
 import com.conveyal.r5.transit.TripPattern;
 import com.conveyal.r5.transit.TripSchedule;
@@ -12,8 +13,6 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
 
-import static com.google.common.base.Preconditions.checkElementIndex;
-import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
 
 /**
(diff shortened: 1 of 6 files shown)


src/main/java/com/conveyal/r5/profile/FrequencyRandomOffsets.java
```diff
@@ -44,13 +43,19 @@ public class FrequencyRandomOffsets {
     private final Map<TripSchedule, int[]> offsetsForTripSchedule = new HashMap<>();
 
     /** The mersenne twister is a higher quality random number generator than the one included with Java */
-    private MersenneTwister mt = new MersenneTwister();
+    private MersenneTwister mt;
 
-    public FrequencyRandomOffsets(TransitLayer data) {
+    public FrequencyRandomOffsets(TransitLayer data, ProfileRequest request) {
```
trevorgerhardt, line 48: Is the entire request necessary to be passed as a parameter? I would recommend passing either the seed itself or a `MersenneTwister`.

abyrd:
In light of recent discussions summarized at https://github.com/conveyal/r5/issues/714#issuecomment-1665124612 do we still want to add a `lockSchedules` parameter or just use a custom/experimental worker build if this is ever needed? Have we been able to handle most cases where this was an issue by simply switching to exact-times or phased schedules, or increasing the number of MC draws?

src/main/java/com/conveyal/r5/profile/FrequencyRandomOffsets.java
abyrd: Assuming we want to allow locking the schedules at all I agree: it would be better to pass in only the seed to deter further coupling with the ProfileRequest. The seed should probably include both lat and lon to avoid correlation between all origins in a row across a grid.

I don't see a clear justification for changing the schedules from one origin to another, but locking them at the per-origin (or per-row) level. This would lead to apparently stable differences between neighboring origins that use the same routes to reach most destinations, which are likely to be misinterpreted. 

The effects of knowingly deriving results from one single permutation of the schedules are clearer if that permutation is the same for all origins. In that case the seed could just be derived from the TransitLayer (e.g. its center point), and the parameter can just be a boolean for whether to do this.

Issue #875, opened 2023-03-20
Feature: time to nth nearest ("dual" accessibility)


ansoncfit:
Users from multiple organizations in California have requested "dual" accessibility measures (e.g. time required to reach some threshold number of opportunities, such as one school or three grocery stores). The TransitCenter Equity Dashboard includes these types of measures (specifically, the time needed to reach the 3rd Grocery Store, 3rd Pharmacy, 1st Urgent Care, 1st Hospital, and 1st College/University).

ansoncfit:
We also had a recent request for this feature in the Netherlands.

abyrd:
Users will often use this feature with many freeform destination layers at once (TransitCenter uses five layers). So we should explore allowing multiple freeform destinations in one regional analysis which may allow big efficiency gains (not running multiple regional analyses), or at least update documentation to explain why gridded destinations or combining all freeform destinations into a single layer is beneficial here.

PR #884, opened 2023-08-04
Temporal Opportunity Density and Dual Accessibility

abyrd:
This is an initial draft of the feature described in #875: finding the closest N destinations to each origin in regional analyses and reporting them back to the broker. Implementation should eventually be improved (e.g. some abstraction or ordering in the sorting of retained nearby destinations). There is not yet any way to collate the results to CSV or grid files for display and use.


abyrd:
Hmm, we have `com.conveyal.r5.analyst.cluster.TravelTimeResult#histograms`. The dual of accessibility can be derived very efficiently from a binned histogram of travel times. It wouldn't tell us which specific destinations were reached but I'm not sure anyone is asking for that information. However in TravelTimeResult we're recording separate histograms at each destination, not merging them across all destinations. It's very inefficient to retain the travel times and histograms to every destination and sort thousands or millions of items just to report the time to the few closest destinations. So it might make sense to reimplement `nearestNResult` to accumulate a histogram of travel times across all destinations, while also tracking which specific destinations (by ID) were the closest, in case anyone wants that info for only the top 2-3 points.

abyrd:
On further thought, a histogram of how many destinations are reached at each minute is just the discrete derivative of accessibility. In single-point requests where we ask the worker to report all 120 cutoffs, the marginal increase in accessibility as we increase the cutoff by 1 minute is the number of destinations in that histogram bin. In fact we don't even need to materialize the histogram at all. Iterating over the cutoffs starting at 1, we can simply note the cutoff value where the cumulative access curve crosses the desired threshold and bail out. 

It might be tricky though to selectively enable all 120 cutoffs in only those regional analyses where we want to report dual accessibility, without inadvertently generating 120 sets of gridded accessibility results. It's probably simpler and more maintainable to build up that one-minute-resolution histogram separately, independent of how many cutoffs are specified.

And then consider that the histogram or dual of accessibility can only be easily derived from the cumulative accessibility curve when using the step function. When using other decay functions we might still want to know at exactly which minute the opportunities were located - this allows some interesting/informative visualizations alongside the resulting cumulative accessibility curve. This is another argument for accumulating and reporting the histogram independently from the accessibility values.

Of course maybe someone actually wants to record dual accessibility where the threshold accessibility indicator value is computed using a custom decay function. For example, how many minutes of travel to reach 100k jobs, using logistic decay to weight the jobs. In that case you would again need to compute the indicator value at every cutoff - this can be done by essentially issuing a single point request at each origin and scanning over the 120 cutoffs until the threshold value is exceeded. So it can currently be done in a script calling our API but not in a regional analysis.

src/main/java/com/conveyal/r5/analyst/TemporalDensityResult.java
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * An instance of this is included in a OneOriginResult for reporting how many opportunities are encountered during each
+ * minute of travel. If we use more than one destination point set they are already constrained to all be aligned with
+ * the same number of destinations.
+ *
+ * The data retained here feed into three different kinds of results: "Dual" accessibility (the number of opportunities
+ * reached in a given number of minutes of travel time); temporal opportunity density (analogous to a probability density
+ * function, how many opportunities are encountered during each minute of travel, whose integral is the cumulative
+ * accessibility curve).
```
ansoncfit, line 17: Perhaps part of this comment was unintentionally deleted? Otherwise,

```suggestion
 * The data retained here feed into two different kinds of results: "Dual" accessibility (the number of minutes of
 * travel time needed to reach a given number of opportunities); and temporal opportunity density (analogous to a probability density
 * function, how many opportunities are encountered during each minute of travel, whose integral is the cumulative
 * accessibility curve).
```

src/main/java/com/conveyal/r5/analyst/cluster/TravelTimeResult.java
```diff
@@ -37,6 +37,9 @@ public class TravelTimeResult {
      * This is optional, generally used only for debugging and testing, so may be null if no histograms are recorded.
      * Since this is used for debugging and testing it might even be interesting to do it at one-second resolution or
      * just store every travel time, but binning makes it easy to compare with probability distributions.
+     * Note that this is one histogram _per target_ showing on how many iterations each travel time is the fastest,
+     * _not_ one histogram per origin/percentile showing how many destinations are reached at each travel time. The
+     * latter is essentially the discrete derivative of step-function accessibility and is tracked elsewhere (TemporalDensityResult).
```
ansoncfit, line 42: Helpful to avoid future confusion 👍

src/main/java/com/conveyal/analysis/results/TemporalDensityCsvResultWriter.java
```diff
@@ -0,0 +1,90 @@
...
+import com.conveyal.r5.analyst.cluster.RegionalWorkResult;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * This handles collating regional results into CSV files containing temporal opportunity density
+ * (number of opportunities reached in each one-minute interval, the derivative of step-function accessibility)
+ * as well as "dual" accessibility (the amount of time needed to reach n opportunities).
```
ansoncfit, line 14: "Dual" accessibility will be recorded as -1 if `n` is not specified, or if the time needed exceeds 120 minutes.

ansoncfit (approved):
Looks good. I added a few minor clarifications/suggestions in-line, and these can be addressed later. To test, we'll check that with freeform origins, `includeTemporalDensity: true`, and `dualAccessibilityThreshold` set to a positive integer, regional results yield a CSV with the expected travel time density and dual accessibility result.

src/main/java/com/conveyal/r5/common/Util.java
ansoncfit, line 1: Is the convenience method added here used elsewhere? I did not see it in a quick search of the diff.

PR #890, opened 2023-10-05
Remove unused pattern trips field

trevorgerhardt:
This field is not used anywhere in R5 and will no longer be used in the UI. This corresponds to https://github.com/conveyal/ui/pull/1978.


abyrd:
Discussed in call:
- R5 worker version of this class `com.conveyal.r5.analyst.scenario.AdjustFrequency` does not even have a corresponding field, so there should be no issue with worker version backward compatibility.
- Deserializer handling objects coming out of Mongo has "fail on unknown properties" set to false, so we don't expect problems even if R5 encounters old database records containing this field.

abyrd (approved):

PR #894, opened 2023-10-17
Use pixel centers as points in web Mercator grids

abyrd:
Travel time and accessibility results will be slightly different when computed with workers after this change. Pre-built linkages in saved networks will not be affected - the entire network must be rebuilt to ensure they are consistent, so this **requires a network version bump**. However we are also updating to Kryo 5 serialization format, so if both changes are introduced at once in a release we need only bump the network version number once.


abyrd:
Simpson desert tests are failing, which is not surprising as travel times are slightly different with this linking strategy.

abyrd:
Tests are now passing with the Simpson Desert test changes merged in. I think this is ready for final review.

trevorgerhardt (approved):

abyrd:
@trevorgerhardt merge commit to combine the use of extents with pixelToCenterX invalidated your review.

trevorgerhardt (approved):

PR #895, opened 2023-10-17
Compose WebMercatorGridPointSet with WebMercatorExtents to reduce duplication

abyrd:
DRAFT PR trying out reuse of code from WebMercatorExtents. Some logic was slightly different between the two, this makes it more uniform with less duplication.


abyrd:
Duplicate of #900, closing

Issue #893, opened 2023-10-17
Send IDs for origin and destination pointsets in HTTP API


abyrd:
R5 did not properly handle the edge case where WGS84 extents fall exactly on a web Mercator pixel boundary, or slightly outside it due to floating point math. This is being addressed in https://github.com/conveyal/r5/pull/892

In discussion of this problem we found the ideal solution to be specifying Mercator extents using integers on the HTTP API (instead of using floating point WGS84). However, our existing way of sending origin/destination extents has weird inconsistent behavior that we don't want to replicate for both WGS84 and Mercator. The extents parameters also behave differently for single point and regional analyses, and freeform or gridded point sets. We don't want to replicate or perpetuate this behavior.

We also don't want to bake more Mercator-specific details into the API, because increasing use of GL-accelerated maps means Conveyal may eventually deprecate web Mercator. 

My suggestion is that we work toward eventually having IDs for each pointset created by the user, whether they're freeform, Mercator, or grids in some other coordinate system, and pass those IDs in as the origin and destination points in the API, having separate fields for origins and destinations.

This does mean the workers need access to some additional information via API or direct database access (if grid characteristics are stored in a database). Currently workers never contact the database so this change requires careful consideration.

Alternatively, we could have a shared type for origin and destination points fields, which allows specifying web Mercator extents or freeform points by ID in separate properties. But I feel like web Mercator is on the way out. Related to #735.

PR #897, opened 2023-10-19
Make repeated WGS84-Mercator conversion stable

abyrd:
This PR supersedes #892, which has been split into multiple PRs. This is a more minimal changeset than originally proposed, to facilitate review and contain any risk from changing heavily used math functions.

As recently discussed: regional request bounds are specified in WGS84 coordinates, which are converted to web Mercator extents when regional tasks are processed. Repeating a regional analysis with the same bounds as an existing one requires the the UI to look at the web Mercator bounds of an existing analysis, then convert them to WGS84 to create the new regional analysis request. This creates a feedback loop between WGS84 and web Mercator where WGS84 bounding boxes lie exactly on the edges of Mercator pixels, creating a literal edge case for the function `WebMercatorExtents#forWgsEnvelope` and `populateTask`.

This PR revises the way web Mercator extents are created to handle this edge case. This change is reinforced with additional factory methods that should handle any numerical instability or lack of precision in the repeated conversions.

The eventual real fix for this is https://github.com/conveyal/r5/issues/893.

This PR includes a test of repeated conversions, which fails without these changes and passes with these changes. 
Javadoc was also added and updated.


src/main/java/com/conveyal/r5/analyst/WebMercatorExtents.java
```diff
@@ -114,23 +118,82 @@ public WebMercatorExtents expandToInclude (WebMercatorExtents other) {
...
+        // Note that web Mercator coordinates increase from north to south, so maximum latitude is minimum Mercator y.
         int north = latToPixel(wgsEnvelope.getMaxY(), zoom);
         int west = lonToPixel(wgsEnvelope.getMinX(), zoom);
-        int height = (latToPixel(wgsEnvelope.getMinY(), zoom) - north) + 1; // minimum height is 1
-        int width = (lonToPixel(wgsEnvelope.getMaxX(), zoom) - west) + 1; // minimum width is 1
-        WebMercatorExtents webMercatorExtents = new WebMercatorExtents(west, north, width, height, zoom);
-        return webMercatorExtents;
+        // Find width and height in whole pixels, handling the case where the right envelope edge is on a pixel edge.
+        int height = (int) Math.ceil(latToFractionalPixel(wgsEnvelope.getMinY(), zoom) - north);
+        int width = (int) Math.ceil(lonToFractionalPixel(wgsEnvelope.getMaxX(), zoom) - west);
```
abyrd, line 141: Note that this is a key part of the PR, using ceil() to handle the case where the projected value already lies exactly on a pixel boundary.

ansoncfit (approved):
These changes look good to me.

abyrd:
A corresponding or related UI PR is https://github.com/conveyal/ui/pull/1985, where I made some comments about coordinating behavior between the two sides.

trevorgerhardt (approved):
👍

PR #898, opened 2023-10-19
Additional coordinate stability considerations

abyrd:
This builds on #897. The coordinate stability considerations addressed there also apply to building grids around sets of points, including destination opportunities and street vertices in transportation networks. Due to conversion back and forth from WGS84 to web Mercator, it is possible hat a grid made to be completely tight around the set of points may not actually contain some of the points.

This PR adds `WebMercatorExtents.forBufferedWgsEnvelope` which complements `forTrimmedWgsEnvelope` to handle this case.

This PR is a draft until the basic WGS-Mercator PR is merged, at which point this PR's base can be changed to dev.


trevorgerhardt (approved):
The ways this PR is using `forBufferedWgsEnvelope` appear correct to me. 👍

PR #900, opened 2023-10-19
Compose with WebMercatorExtents to reduce duplication

abyrd:
WebMercatorGridPointSet contained a lot of duplicated fields and logic from WebMercatorExtents. 
Objects constructed with WebMercatorGridPointSet constructor were often one unit too narrow (not consistent with WebMercatorExtents).

With this PR, WebMercatorExtents becomes Serializable and will be stored in TransportNetwork files so **this requires a network version bump** and should ideally be included in a release version with other such changes.


abyrd:
This causes the Simpson Desert tests to fail, but likely for a different reason than the pixel-center change does. The PR revising the Simpson Desert tests would also make them pass here, but we should first find out why they're failing here.

trevorgerhardt (dismissed):
This refactor looks good to me 👍

abyrd:
The main reason I left this as a draft was uncertainty around whether there were serialized objects lying around that would become incompatible. Because changing WebMercatorGridPointSet to have a nested WebMercatorExtents required making the latter serializable, we've clearly got old WebMercatorGridPointSets around that we need to deserialize without conflicts. I believe that's only in the serialized network files, which would only be read by the workers, which understand the idea of versioned network files. I just need to make sure this isn't used anywhere else before merging it.

abyrd:
Some serialization systems (e.g. for serializing to JSON) will not only serialize fields but also public getter methods. The result of WebMercatorExtents.getMercatorEnvelopeMeters could be serialized in this case. But I believe we're only serializing this class with Kryo, whose default serializer writes out only the fields of the class. For standard Java serialization the serialization version ID is important but messy to manually maintain. We should only be serializing this class with Kryo, not other systems, and I think that's currently the case.

abyrd:
As I was reviewing this I noticed another obvious place where WebMercatorExtents were being copied and compared field by field. I added a commit that updates WebMercatorGridPointSetCache.GridKey in a similar way, composing with WebMercatorExtents.

Following up on my comment above, I don't see anywhere WebMercatorGridPointSet is being stored except in TransportNetworks so I think this PR will be sufficiently backward compatible.

abyrd:
@trevorgerhardt or @ansoncfit please give this one another look to verify the semantic equality logic in WebMercatorGridPointSetCache and then feel free to approve/merge.

src/main/java/com/conveyal/r5/kryo/KryoNetworkSerializer.java
```diff
@@ -46,7 +46,8 @@ public abstract class KryoNetworkSerializer {
      * We considered using an ISO date string as the version but that could get confusing when seen in filenames.
      *
      * History of Network Version (NV) changes:
-     * nv3 use Kryo 5 serialization format
+     * nv4 2023-11-02 WebMercatorGridPointSet now contains nested WebMercatorExtents
+     * nv3 2023-18-01 use Kryo 5 serialization format
```
trevorgerhardt, line 50:
```suggestion
     * nv3 2023-01-18 use Kryo 5 serialization format
```

abyrd:
Build is now passing (code now compiling with changed symbolic constant). I think this is ready for final review.

trevorgerhardt (approved):

Issue #747, opened 2021-08-28
Support GeoPackage and GeoJSON uploads


abyrd:
Shapefiles are completely archaic and we've wasted a lot of time diagnosing and dealing with Shapefile-specific issues. The use of sidecar files, 10-character property names with no awareness of UTF-8, etc. are ridiculous in 2021.

We should be part of the movement to eliminate this mess, by making it possible for our users to use other formats and encouraging them to do so in the documentation. An informal poll cited on the "Shapefile must die" page finds that the most favored replacements are GeoPackage and GeoJSON, which are supported by almost 100% of the voters.

See:
http://switchfromshapefile.org
http://www.geopackage.org

PR #902, opened 2023-10-26
Support GeoJSON uploads for creating opportunity datasets.

trevorgerhardt:
Closely mirrors the way Shapefiles are converted into grids.

Refs #747


ansoncfit (commented):
Based on a discussion last year, we might want to follow the approach in https://github.com/conveyal/r5/issues/761. If we're not going to switch to that workflow yet, it might be good at this stage to reduce some of the repetition here, by refactoring some of the `fromShapefile` code into reusable methods.

trevorgerhardt:
> Based on a discussion last year, we might want to follow the approach in #761. 

This doesn't prevent or hinder following the approach in #761 whenever that is implemented.

> If we're not going to switch to that workflow yet, it might be good at this stage to reduce some of the repetition here, by refactoring some of the `fromShapefile` code into reusable methods.

Ok, I'll take a look and see if there is something to pull out there 👍

PR #912, opened 2023-11-22
Change dual accessibility CSV headers and column order

abyrd:


ansoncfit (approved):
Thanks for the updates. As we discussed on the call, results look sensible.

abyrd:
Noting here that the final commit is a workaround for a weird detail of Java: strangely it's possible to read the value of an uninitialized final variable in a constructor and this is not considered an error. This would often fail fast for null references but not for primitives - their final values are not actually final. Java also disallows setting such final fields before calling the superclass constructor. We may be able to reorder some initialization to lessen this effect but for now we're just leaving the dual threshold out of the column headers.

abyrd:
I just discovered that the problem described above is the subject of a change introduced in Java 22:
https://openjdk.org/jeps/447 "statements before super()"

PR #917, opened 2023-12-04
Prevent ArrayIndexOutOfBoundsException when recording TemporalDensityResult

ansoncfit:
We decided our temporal density results have 120 columns (0 to 119). In the `TemporalDensityResult` constructor, the third dimension is 120:

`this.opportunitiesPerMinute = new double[destinationPointSets.length][nPercentiles][120]`

When recording results, opportunities should only be recorded if they are less than 120 minutes away, as proposed in this PR.


abyrd (approved):
It's clear from the updated Javadoc and the constant array dimension in line 48/49 where the error came from, and that the patch will keep m within array index bounds. As mentioned on the call, we must not have generated a travel time of exactly 120 minutes in testing.

PR #918, opened 2023-12-19
Fix freeform guardrails

ansoncfit:
We currently limit the number of destinations and OD pairs in regional analyses with time/path results. This PR fixes a couple issues with the related guardrails:
- Fixes a conditional so that when `oneToOne = true`, the broker checks the number of OD pairs (which equals the number of origins) against the correct limit, rather than just allowing the analysis to proceed.
- Checks limit on number of destinations for path analyses in broker. Workers were applying a stricter limit (5000 destinations for path analyses) than the broker, so the broker could potentially enqueue millions of tasks that would trigger an immediate error once they reached the workers.

Small refactor for brevity/legibility: `task = job.templateTask`

### DIFF
--- a/src/main/java/com/conveyal/analysis/results/MultiOriginAssembler.java
+++ b/src/main/java/com/conveyal/analysis/results/MultiOriginAssembler.java
@@ -7,6 +7,8 @@ import com.conveyal.analysis.persistence.Persistence;
 import com.conveyal.file.FileStorage;
 import com.conveyal.file.FileStorageFormat;
 import com.conveyal.r5.analyst.PointSet;
+import com.conveyal.r5.analyst.cluster.PathResult;
+import com.conveyal.r5.analyst.cluster.RegionalTask;
 import com.conveyal.r5.analyst.cluster.RegionalWorkResult;
 import com.conveyal.r5.util.ExceptionUtils;
 import org.slf4j.Logger;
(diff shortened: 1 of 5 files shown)


src/main/java/com/conveyal/r5/analyst/cluster/PathResult.java
```diff
@@ -38,7 +38,7 @@ public class PathResult {
      * These results are returned to the backend over an HTTP API so we don't want to risk making them too huge.
      * This could be set to a higher number in cases where you know the result return channel can handle the size.
      */
-    public static int maxDestinations = 5000;
+    public static int MAX_PATH_DESTINATIONS = 5_000;
```
abyrd, line 41: It looks like this was originally left as a variable to allow adjustments, but it's not clear how that adjustment would be made now given how these objects are created. So it does make sense to treat this as a constant, and all-caps is conventionally reserved for constants. But in that case there's a missing `final` here.

Similar comment for the DATA_COLUMNS below, looks like we're missing a `final`.

abyrd:
The underlying problem where tasks continue to be distributed to workers after an error was reported should be solved by #921. With that fixed, mismatches in validation rules between different components (UI, backend, workers) should be much less problematic and just result in a few error messages.

src/main/java/com/conveyal/analysis/results/MultiOriginAssembler.java
```diff
@@ -89,21 +91,27 @@ public MultiOriginAssembler (RegionalAnalysis regionalAnalysis, Job job, FileSto
...
+            // destination pointsets are not too big for generating CSV files.
+            RegionalTask task = job.templateTask;
+            if (!task.makeTauiSite && task.destinationPointSetKeys[0].endsWith(FileStorageFormat.FREEFORM.extension)) {
+                // This requires us to have already loaded this destination pointset instance into the transient field.
+                PointSet destinationPointSet = task.destinationPointSets[0];
+                int nDestinations = destinationPointSet.featureCount();
+                int nODPairs = task.oneToOne ? nOriginsTotal : nOriginsTotal * nDestinations;
+                if (task.recordTimes &&
+                    (nDestinations > MAX_FREEFORM_DESTINATIONS || nODPairs > MAX_FREEFORM_OD_PAIRS)) {
+                    throw new AnalysisServerException(String.format(
```
ansoncfit, line 104: Is `throw AnalysisServerException.badRequest(...` preferable here?

Is it straightforward to apply `filterStackTrace` (moved to ExceptionUtils in #919) to the message eventually passed along to the UI?

abyrd: I think so. It looks like we could apply `filterStackTrace` to the `message` parameter in each call to `HttpApi#respondToException` (there are five of them just above the method definition).

abyrd: On your first question: it looks like calling badRequest would yield an identical AnalysisServerException instance but with type BAD_REQUEST instead of UNKNOWN. It looks like the only effect would be the string BAD_REQUEST instead of UNKNOWN in a field in the JSON returned to the UI. Assuming this does not change UI behavior in any problematic way, yes badRequest does seem to better represent the source of the exception.

abyrd (dismissed):
I am approving this as-is in case you want to merge it now. Feel free to add the exception changes discussed in the PR comments if you like in which case I can merge immediately after re-review.

src/main/java/com/conveyal/analysis/results/MultiOriginAssembler.java
ansoncfit: Thanks for pointing out that method. I think we want to apply it to the stackTrace, not the message. I committed that below for testing on staging, but feel free to revert.

abyrd: Oh right. It was late when I was reading that and I mistakenly thought the callers were passing `e.getStackTrace()` as the message. This was probably driven by some kind of wishful thinking because I realized filterStackTrace is so narrowly suited to output from one method (throwable.printStackTrace via ExceptionUtils.stackTraceString) that it probably shouldn't be public and applicable to the more general String type. This is a minor detail but it's located in error handling code that should be foolproof (things could rapidly get annoying to debug if the signal path for the debugging messages themselves is failing). It looks like it is failsafe and would at worst return "Unknown stack frame" but there could be some edge case we're not thinking of. 
Of course in current usage, both places the method is used are `filterStackTrace(errorEvent.stackTrace)` so we know it's always being applied to a final field that's the output of `ExceptionUtils.filterStackTrace()`. This should work fine but I'm still thinking about ways to clean up the method signature.
Option A: have the public method be `public static String filterStackTrace (Throwable throwable)` then call that in the ErrorEvent constructor. This would cause all console log traces to be filtered. All others (Slack) are already filtered.
Option B: move filterStackTrace back to a private method on ErrorEvent and always use it via a public wrapper method that only reads from ErrorEvent.stackTrace. (Won't work because we need it in RegionalWorkResult)
Option C: Add a field to ErrorEvent so it has both the full and filtered stack trace
Option D: keep the entire Throwable in a private field on the ErrorEvent and generate the stack traces on demand

abyrd: I did option C as it ensures the filtered trace always came from the right method but still allows for verbose stack traces in console logs. D looks cleaner, but it defers processing of the stack trace to the moment the log or slack message is written, thus deferring some potential failures to those background event handling threads. I'd rather front-load the stack trace processing to the point where the ErrorEvent is created, where any delays and problems are more self-contained.

Issue #919, opened 2023-12-19
Reduce errors sent to broker


ansoncfit:
When workers encounter an exception on a regional task, they send a String error message back to the broker: https://github.com/conveyal/r5/blob/v7.0/src/main/java/com/conveyal/r5/analyst/cluster/AnalysisWorker.java#L294-L300

Many workers sending many error messages at once can overwhelm the broker. If there is an exception, it will likely affect or invalidate all origins, so most of these messages are duplicative. Some ideas for revising worker behavior:

- Only report one error per batch of tasks received
- Stop returning results on a job after a certain number of errors have been reported

abyrd:
It would be possible to limit error reporting on the worker side. But the workers are intended to report at most one batch (`4*n_cores`) of errored tasks before the backend stops delivering tasks to them. The fact that worker error reporting became a problem revealed an underlying problem in job cancellation (see #887 and #921).

Workers do not currently associate tasks with the batch they were received in, so the first suggestion would add a bit of complexity. The second suggestion would only involve keeping a set of errored jobs, which could be global to the worker or just one per polling operation. This would be relatively simple, but it does add one more step to the reporting process (thus one more thing that can go wrong). 

I'm inclined to just get the job cancellation working properly, in which case we should only ever get one batch of errors before an errored job stops. We could then reinforce it with the second suggestion (perhaps limited to the scope of a single polling operation) if we feel there's still risk to be managed.

abyrd:
Closing this because we merged #918 to the default branch. In the future I'll try to more regularly set the associated issues on pull requests, which should auto-close them upon merge.

Issue #887, opened 2023-08-17
Regional tasks enqueued if result assembler creation fails


abyrd:
As noted in code comments in `Broker#enqueueTasksForRegionalJob`, if creation of MultiOriginAssembler fails with an exception the job and its tasks are already enqueued, possibly being processed, with no assembler to receive the results. This is not exactly catastrophic, but it's not a good behavior to proceed with running the analysis in these circumstances.

We may be able to simply create the result assembler before enqueueing the tasks so the analysis does not start if the assemblers can't be created, but we first need to check whether this change has any side effects.

abyrd:
This can actually be more problematic than it seems, in the specific case where the assembler creation fails and the tasks distributed to the workers also fail. The fact that the job exists but no assembler exists violates an assumption in `Broker#handleRegionalWorkResult` causing it not to record errors. The job then continues to be redelivered to workers many times, and the workers flood the backend with huge amounts of long stack traces.

The simplest solution to the main problem is to defer adding the new job to the map of active jobs until we're sure the assembler creation has completed without errors. I see no reason this will cause any problems, because the MultiOriginAssember constructor receives a reference to the new job and doesn't go looking for it in the list of active jobs.

Several other things could make error reporting and job cancellation more robust. One is reversing the order of two operations in `Broker#handleRegionalWorkResult` so errors are recorded on the job before any other conditionals that can exit the method. This alone would have made the problem discussed in the present issue "not exactly catastrophic" as originally described.

Another improvement would be shortening the error messages reported by the workers and recording only one of them on each job.

PR #921, opened 2023-12-20
Prevent runaway error reporting

abyrd:
This PR addresses problems discussed in #919 and #887. It makes two small changes, each of which alone would prevent the runaway error reporting observed on 2023-12-18 (EST). In addition it reduces the volume of errors sent and recorded.

### DIFF
--- a/src/main/java/com/conveyal/analysis/components/broker/Broker.java
+++ b/src/main/java/com/conveyal/analysis/components/broker/Broker.java
@@ -398,7 +398,11 @@ public class Broker implements Component {
      */
     private synchronized void recordJobError (Job job, String error) {
         if (job != null) {
-            job.errors.add(error);
+            // Limit the number of errors recorded to one.
+            // Still using a Set<String> instead of just String since the set of errors is exposed in a UI-facing API.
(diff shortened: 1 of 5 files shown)


src/main/java/com/conveyal/analysis/components/broker/Broker.java
```diff
@@ -385,14 +389,20 @@ public synchronized void markTaskCompleted (Job job, int taskId) {
...
-     * This method ensures synchronization of writes to Jobs from the unsynchronized worker poll HTTP handler.
      */
     private synchronized void recordJobError (Job job, String error) {
         if (job != null) {
-            job.errors.add(error);
+            // Limit the number of errors recorded to one.
+            // Still using a Set<String> instead of just String since the set of errors is exposed in a UI-facing API.
+            if (job.errors.isEmpty()) {
+                job.errors.add(error);
+            }
```
ansoncfit, line 405: This seems like a sensible safeguard -- in most cases, error messages coming in from workers will be identical. And if they vary slightly (e.g. including a different origin id), we probably don't want to record all messages separately.

ansoncfit (approved):

Issue #621, opened 2020-10-09
Regional analysis persistence cleanup


ansoncfit:
## How scenarios are saved (from @abyrd)

For a few reasons, we want to save a full scenario for each regional analysis run. This should allow exactly reproducing an old regional analysis on its associated worker version, or rendering a PDF/HTML report of the scenario for a completed regional analysis.

We currently save the full list of R5 modifications in MongoDB inside the regional analysis object. This doesn't meet the above goals for a few reasons: first, our scenario report rendering works on UI/backend modifications, not R5 modifications. Second, we're intentionally trying to retain potentially outdated modification types, or modification types specific to a certain version of R5. When the backend tries to fetch and deserialize such modifications it fails. This is already happening now that we have custom modifications.

Possible solutions are: 1. Store UI/backend modifications in Mongo instead of R5 modifications, avoiding deserialization of the stored objects in case their structure is different than the ones known to the backend. 2. Store R5 modifications but don't store them in Mongo.

Regarding option 1: Storing UI/Backend modification objects in Mongo should make for less problematic migrations in the event that modifications change, because the same migration process could be applied to these stored scenarios as to the main store of scenarios. However this defeats the purpose of storing old scenarios - you might not be able to exactly reproduce the R5 modifications that were effectively used to run a specific regional analysis.

Regarding option 2: We already in fact do this, because every regional analysis saves its R5-formatted scenario to a file on S3. So the only problem here is that there is no way to generate an HTML report on a finished scenario (because the reports are generated from UI/backend modifications). If modification structure may change, then there will never be a reliable way to generate reports after the fact anyway. So maybe we should just generate a report for every regional analysis and store it on S3 alongside the regional results.

In the short term @trevorgerhardt has added exclusions to all the HTTP API endpoints that fetch Regional analyses so they don't fetch the subfield containing the scenario (in conveyal/analysis-backend#214).

Later we may just remove all this data from the MongoDB entirely, since the R5 snapshot of each scenario is already being saved on S3 as a JSON file.

## Superfluous copying and duplication of RegionalAnalysis fields (from @abyrd)
[...]

PR #926, opened 2024-01-18
Exclude modifications for regional analysis lookup

trevorgerhardt:
This change adds a new helper method to pass a `fields` object to `MongoMap#findByIdIfPermitted` and utilizes that to exclude modifications while looking up a regional analysis to generate a `scenarioJsonUrl`.

This will prevent `JacksonDBDecoder` errors from occurring if custom modifications are used in a worker version that do not also exist in the server. Sample errors like:

```
java.lang.RuntimeException: IOException encountered while reading from a byte array input stream
at org.mongojack.internal.stream.JacksonDBDecoder.decode(JacksonDBDecoder.java:67)
at com.conveyal.analysis.persistence.MongoMap.findByIdIfPermitted(MongoMap.java:51)
```

### DIFF
--- a/src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
+++ b/src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
@@ -527,8 +527,11 @@ public class RegionalAnalysisController implements HttpController {
      * the given regional analysis.
      */
     private JsonNode getScenarioJsonUrl (Request request, Response response) {
-        RegionalAnalysis regionalAnalysis = Persistence.regionalAnalyses
-                .findByIdIfPermitted(request.params("_id"), UserPermissions.from(request));
+        RegionalAnalysis regionalAnalysis = Persistence.regionalAnalyses.findByIdIfPermitted(
+                request.params("_id"),
+                DBProjection.exclude("request.scenario.modifications"),
+                UserPermissions.from(request)
+        );
         // In the persisted objects, regionalAnalysis.scenarioId seems to be null. Get it from the embedded request.
         final String networkId = regionalAnalysis.bundleId;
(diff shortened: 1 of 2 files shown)


abyrd (approved):
Reviewed together with @ansoncfit at the Thursday meeting. We talked through the underlying problem.

- Backend should be able to deserialize unknown types without blowing up (into CustomModificationHolder).
- Arguably these should be written into the database as type=custom as the original intent was to only ever deserialize custom modifications in a typed fashion on workers.
- Considered whether it is necessary to save this scenario detail into the database at all if we’re always providing it by fetching this JSON file.
- Read through #621 recalling that we have long discussed representing the scenario modifications in persistent storage with the UI types instead of the R5 types, to better decouple them from worker versions.
- Noted that this PR specifically eliminates the Scenario, while in fact the API endpoint only needs two fields from the entire regional analysis object. So a more basic way to circumvent the problem is to request only the two fields actually needed from the database. However the database wrapper available here focuses solely on fetching into typed objects, probably making that awkward.

Considering all of the above, we decided to go ahead with the approach in this PR, and will create or update an issue to capture the various observations for future reference.

abyrd:
I just tried out replacing `DBProjection.exclude("request.scenario.modifications")` with `DBProjection.include("bundleId", "scenarioId", "accessGroup"` to grab only the fields necessary to return the API response. This does work... almost. But the field `regionalAnalysis.request.scenarioId` is different than `regionalAnalysis.scenarioId` in the baseline scenario case (UUID versus the string "baseline" respectively). Fetching `regionalAnalysis.request.scenarioId` means fetching `regionalAnalysis.request` which runs into other problems with type fields and quickly becomes more trouble than it's worth. The other underlying problem to note here is that we have multiple ways to reference the baseline scenario as a string and mix them together in a single object.

Issue #920, opened 2023-12-20
A feed_id for regional analysis path results


edasmalchi:
@hhmckay and I are working on a process for using Conveyal/r5 to find common transfer points within a given region in California. We can provide our own origin/destination points and have a working proof of concept using the regional analysis export paths feature for Sacramento, but many trips involve multiple transit operators and thus multiple GTFS feeds.

All feeds are present in the Network Bundles we've updated so they're correctly considered in routing, but we would like to link the path results back to our own GTFS data warehouse. This will be very difficult without a way to tell which GTFS feeds the routes and stops in the path results refer to.

**Describe the solution you'd like**

Some sort of GTFS feed id present for routes in the path analysis output CSV. `routeFeeds`?

We can accept an arbitrary feed id and do some work on our end, but it would be even better if they can be assigned in some sort of documented, deterministic way. 

@ansoncfit mentioned that `feed_info.txt` may be used to generate these, it would be great to have that process documented (which fields within feed_info contribute to the id, etc). If so we might consider editing feeds on our end prior to network bundle upload to ensure `feed_info.txt` contains our preferred identifiers...

Our GTFS data warehouse uses various keys derived from our internal organization records, feed urls, etc. Since that context isn't present within Conveyal, it's not possible to regenerate exact matches.

**Describe alternatives you've considered**

Nothing compelling, would potentially involve significant manual effort, or standing up another trip planner/routing engine.

**Additional context**

https://calsta.ca.gov/subject-areas/sb125-transit-program
https://github.com/cal-itp/data-analyses/issues/888
https://github.com/cal-itp/data-analyses/tree/main/conveyal_update
https://dbt-docs.calitp.org/#!/model/model.calitp_warehouse.dim_gtfs_datasets#details

<img width="667" alt="Screenshot 2023-12-19 170403" src="https://github.com/conveyal/r5/assets/54728455/1535445d-e518-4ddf-bd10-b85de5f39824">

abyrd:
I just looked into this a bit. Adding some notes that could assist with future work:

[...]

PR #927, opened 2024-01-25
Add feed id column to path results

ansoncfit:
This PR addresses #920, adding a column for feedIds to CSV path detail results. It also adds a "group" column to these results, reserved for future use. As noted in the discussion below and in #920, we are deriving the feedId from the boarding stop for convenience; this should generally correspond with the feedId of the route, but it may not (if, for example, a route has been subject to a reroute modification sending it to a stop from a different feed).

Because of some clumsy rebasing, the diff here looks larger than it is. See instead https://github.com/conveyal/r5/compare/dev...results-with-feedids

### DIFF
--- a/src/main/java/com/conveyal/analysis/components/broker/Broker.java
+++ b/src/main/java/com/conveyal/analysis/components/broker/Broker.java
@@ -95,15 +95,14 @@ public class Broker implements Component {
         boolean testTaskRedelivery ();
     }
 
-    private Config config;
+    private final Config config;
 
     // Component Dependencies
     private final FileStorage fileStorage;
     private final EventBus eventBus;
     private final WorkerLauncher workerLauncher;
 
-    private final ListMultimap<WorkerCategory, Job> jobs =
-            MultimapBuilder.hashKeys().arrayListValues().build();
+    private final ListMultimap<WorkerCategory, Job> jobs = MultimapBuilder.hashKeys().arrayListValues().build();
 
     /**
      * The most tasks to deliver to a worker at a time. Workers may request less tasks than this, and the broker should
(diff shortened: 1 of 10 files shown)


abyrd:
As recently discussed: the number of fields will remain fixed. The feed ID field will always be present, but will be empty unless it is intentionally switched on. That is, it won't be automatically enabled even when there are multiple feeds in a bundle. It will be enabled using the "flags" approach demonstrated in #930. 

I'm starting to wonder though whether this merits a structured CsvOptions in the request:
```
{
  csvOptions: {
    stopRepresentation: ID_AND_NAME,
    routeRepresentation: NAME_ONLY,
    includeFeedIds: true
  }
}
```

While we're already considering increasing the hard-wired number of columns, I'd like to suggest adding another empty-by-default field called "label" or "group", which would be used to facilitate filtering and grouping CSV output rows in specialized use cases (such as select-link).

abyrd:
A few comments on the current implementation discussed during the call:

- EntityRepresentation is ignored (we only check if it's null) - we always show the ID (not name) for now
- FeedId is that of the alighting stop, which is usually the same as the feedId of the boarding stop and the route, but might be mismatched if modifications have mixed routes and stops from different feeds
- We are grabbing the feed ID from the stop rather than the route (which might seem like a better representative of the leg) because stops happen to have a readily available feed ID

abyrd (approved):
Looks good (when looking at the alternate diff view you linked to). I wish I understood better why the default diff ends up looking this way due to a rebase, but hopefully the merge will work correctly.

abyrd (dismissed):

abyrd:
Something strange is going on with the review process: 
image
Even if I approve it Github says a review is required. When I do a git diff from this branch to dev locally, I see the expected set of changes which is different than the changes visible here.

abyrd:
Superseded by #936 to overcome apparent bad state on the Github side.

PR #930, opened 2024-02-02
Include names in paths in CSV

abyrd:
This makes it possible to represent stops and routes in CSV output using their names in addition to (or instead of) their IDs. Users have commented that they greatly prefer to work with the names, and often only with the names. Though they're not guaranteed to be unique, in some data sets names are clear enough to work well.

Showing route names has the additional major benefit of showing which modification created a route (instead of a meaningless UUID) in routes that did not come from GTFS.

As recently discussed, this also introduced a freeform `Set<String>` flags field in requests for enabling experimental, undocumented, or arcane behavior in backend or workers. When making a JSON request this will look like `"flags": ["csv_names", "csv_no_ids"]`. This should progressively replace all previous special behavior flags that were embedded inside analysis names etc.

I have tested this locally and it seems to work for all three entity representations.

Now that it's implemented, a few questions emerge. One is whether it's really worth it to define an enum for `EntityRepresentation { ID_ONLY, NAME_ONLY, ID_AND_NAME }` or it should just be two boolean parameters. Another is whether this would be better represented by actually adding a real enum-typed parameter to the requests. It would require less conversions and make it easier to toggle stop and route representations separately for example.

I think the flags facility should still be retained for other purposes even if we stop using it for this purpose.

### DIFF
--- a/src/main/java/com/conveyal/analysis/models/AnalysisRequest.java
+++ b/src/main/java/com/conveyal/analysis/models/AnalysisRequest.java
@@ -21,6 +21,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.EnumSet;
 import java.util.List;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 /**
@@ -175,6 +176,11 @@ public class AnalysisRequest {
      */
     public int dualAccessibilityThreshold = 0;
 
+    /**
+     * Freeform (untyped) flags for enabling experimental, undocumented, or arcane behavior in backend or workers.
+     * This should be used to replace all previous special behavior flags that were embedded inside analysis names etc.
+     */
+    public Set<String> flags;
 
     /**
      * Create the R5 `Scenario` from this request.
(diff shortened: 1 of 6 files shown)


abyrd:
Updated to add a nested CsvRequestOptions structure to control route, stop, feed ID representation independently (as decided on 2024-02-08 meeting).

Order of name and ID have been reversed to place name first. 
String flags system continues to exist in requests but is no longer used to enable this feature.

src/main/java/com/conveyal/analysis/models/AnalysisRequest.java
```diff
@@ -12,15 +12,18 @@
 import com.conveyal.r5.analyst.scenario.Scenario;
 import com.conveyal.r5.api.util.LegMode;
 import com.conveyal.r5.api.util.TransitModes;
+import com.conveyal.r5.transit.TransitLayer;
```
ansoncfit, line 15: Unused import
```suggestion
```

ansoncfit (approved):

Issue #764, opened 2021-11-22
Audit car routing (especially turn restrictions)


abyrd:
We have a whole series of issues from 2016 (five years ago) about the speed and details of car routing, with a lot of references specifically to turn restrictions: #87, #89, #155, #189. I am closing all these issues and consolidating them into this one new issue with the goal to perform an audit of logic specific to car routing and determine if there are indeed inefficiencies or design flaws that should be resolved.

abyrd:
Coming back to this issue as I look at turn restrictions as something that could potentially handle barrier nodes. I'm seeing some things that require attention:
- A comment in `canTurnFrom` method says: "This is also coded in traverse, so change it both places if/when it gets changed." We do not want to have any duplicate code that needs to be kept in sync. On first inspection I don't actually see any duplicate code. This should be verified, and if we no longer need to factor it out then the comment should be changed/deleted.
- The `canTurnFrom` method replicates a list of turn restrictions and stores them in newly created hash maps in the result state. This seems potentially quite inefficient. 
- There is also no mention in the comments on this method that it's not a pure function and modifies its parameters. The behavior should be changed, or the comments updated to warn about this behavior.

abyrd:
Note also #806 reported more recently.

ansoncfit:
I've had this on my reading list for a bit, and it may be worth reviewing before we get underway: https://doi.org/10.1016/j.orl.2012.06.001

abyrd:
I was seeing some rather slow car routing times today and did some investigation. The specific case was a full Netherlands street graph, while performing single point analyses. For various reasons, the travel time limit is set to the maximum value of 120 minutes in this situation. It is not affected by `maxCarTime`, which only applies on transit access legs, nor by `maxTripDurationMinutes`, which in single point analyses only overwrites the TravelTimeSurfaceTask default value of 120 in special cases (fare-constrained multi-criteria routing).

### Base case (no changes to code)
Car routing single point from central Amsterdam takes 11.9 sec.
Sampling two such car searches with VisualVM, two threads each show about 14.5 sec of total CPU time. About 13.6 sec is in `AnalysisWorkerController.handleSinglePoint`. (The other 700 msec is spent in Body.serializeTo(). This is not really expected, but is a separate subject - is some spurious compression happening?)

[...]

PR #931, opened 2024-02-16
keepRoutingOnFoot only in transit access searches

abyrd:
Addresses #764. Continuing routing on foot should only be necessary for searches that are trying to access transit stops. These searches also cover a smaller area than non-transit searches, so the final foot routing step is less time consuming.


abyrd:
I think we can treat this as independent of any other changes we're working on. This is so self contained that it will be easy to back out or tweak even once it's merged to dev.

ansoncfit (approved):

Issue #673, opened 2021-01-18
Cryptic filename when downloading tiffs


erathal:
It would be nice if the tiffs with the regional analysis results would by default be named after the title that the user assigns to any regional analysis.

abyrd:
We need similar logic for giving downloads reasonable names in #610

abyrd:
I just looked into this a bit. Summary of my current thinking:
- These files are stored (within our system) under seemingly random names for a reason - this associates them with known unique IDs in a database.
- Files could of course be renamed on download and saved under a different name. 
- There are several ways to do this. One way is to add a content-disposition header to the HTTP response. Another is to download the file via an HTML anchor (link) with a `download` attribute specifying the alternate name.

### Header Option
A header like `Content-Disposition: attachment; filename="Results of Analysis 12A.csv"` would have the intended effect. However, when deployed to a cloud environment, our current implementation bypasses the Conveyal backend to retrieve the file directly from cloud object storage so the backend does not have a chance to attach a header. When the files are originally pushed to cloud storage, it is possible to set a content-disposition header, but there are a few issues with this approach: a) This would only affect newly created objects. Existing objects would need to be mass-updated with appropriate filenames. b) Stored objects do not exist solely for download by web clients, but may be used by other components of the system. There is potential for incorrect behavior in other (existing or future) parts of the system if files are always received with a header indicating they are attachments that should be saved under a name different than their internal ID. A hypothetical alternate arrangement where files were served by the backend itself would readily provide control over which headers were supplied in varying situations, but this represents a major change from our current approach. In some cases like Geotiff rasters of regional results, requested files are created on demand but are nonetheless shifted into cloud object storage, allowing them to be retrieved in the same manner as pre-existing results files (directly from cloud object storage).

### Download Link Option
I believe this can be implemented as a pure frontend change, by providing files via basic HTML links with download attributes. Files are typically offered for download in a context where all relevant information (region, regional analysis name, layer names etc.) are already needed to render the surrounding page content, so could in principle be assembled to inject meaningful file names into the `download` attribute.

PR #932, opened 2024-02-16
Make human-friendly names available and allow multi-grid downloads 

abyrd:
The backend returns download URLs as small JSON objects with a `url` property. In this PR, these JSON responses are extended to also have a humanName property, which is an alternate filename that is more human readable but not guaranteed to be unique. This enables #673 (as discussed in https://github.com/conveyal/r5/issues/673#issuecomment-1912000319) but should be backward compatible with existing UI as responses still have the same old `url` property. The new `humanName` property can be used in various ways such as the download attribute of an HTML link or as the attachment name in a content-disposition header.

Regional result CSV download links seem to be the odd one out: they return text/plain instead of a JSON wrapped URL. Switching them over to use JSON wrapped URLs like everything else would require an accompanying change on the UI side but would also enable the CSV downloads to use human readable names. The updated code to return the JSON wrapped URL with humanName is already present in RegionalAnalysisController.getCsvResults but commented out for the time being.

This PR also adds and endpoint at `/api/regional/:_id/all` that will generate all available GeoTIFF rasters for a given regional analysis and store them in a persistent ZIP file, returning one of these new-style JSON wrapped download links including a human-readable name. An attempt is also made to replace any UUIDs in the names of files inside this ZIP with relevant human-readable strings.

Many of the changes here are just small changes to return types, MIME types etc. There is one big block add/remove RegionalAnalysisController which is mostly factoring the action of extracting a single grid out of the HTTP controller, so it can be called once by the existing grid-retrieval HTTP controller, but multiple times in a loop by the new bulk-zip-creating HTTP controller. The smaller targeted changes there are somewhat camouflaged/exaggerated by the indentation and movement of that big block.

### DIFF
--- a/src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
+++ b/src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
@@ -39,6 +39,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
+import java.util.Locale;
 import java.util.zip.GZIPOutputStream;
 
 import static com.conveyal.analysis.util.JsonUtil.toJson;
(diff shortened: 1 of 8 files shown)


src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
```diff
@@ -228,89 +402,36 @@ private Object getRegionalResults (Request req, Response res) throws IOException
...
         );
+        res.type(APPLICATION_JSON.asString());
+        // Substitute human readable name and shorten destination UUID.
+        // TODO better system for resolving UUID to human readable names in single and multi-grid cases
+        //      Or maybe only allow multi-grid download for saving by end user, and single grid is purely internal.
+        int firstUnderscore = singleCutoffFileStorageKey.path.indexOf('_');
+        int secondUnderscore = singleCutoffFileStorageKey.path.indexOf('_', firstUnderscore + 1);
+        int lastDot = singleCutoffFileStorageKey.path.lastIndexOf('.');
+        String humanName = analysis.name +
+                singleCutoffFileStorageKey.path.substring(firstUnderscore, firstUnderscore + 6) +
```
ansoncfit, line 429: The first six characters of spatial datasets that were created in quick succession will be the same. Use the last six characters instead?

src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
```diff
@@ -228,89 +402,36 @@ private Object getRegionalResults (Request req, Response res) throws IOException
...
+        // Process has a lot of overhead: UI contacts backend, backend calls S3, backend responds to UI, UI contacts S3.
+        FileStorageKey singleCutoffFileStorageKey = getSingleCutoffGrid(
+                analysis, cutoffIndex, destinationPointSetId, percentile, format
         );
+        res.type(APPLICATION_JSON.asString());
+        // Substitute human readable name and shorten destination UUID.
+        // TODO better system for resolving UUID to human readable names in single and multi-grid cases
+        //      Or maybe only allow multi-grid download for saving by end user, and single grid is purely internal.
+        int firstUnderscore = singleCutoffFileStorageKey.path.indexOf('_');
+        int secondUnderscore = singleCutoffFileStorageKey.path.indexOf('_', firstUnderscore + 1);
```
ansoncfit, line 426: This parses out the destination id, cutoff, and percentile from `singleCutoffFileStorageKey.path`. 

Is there a reason for not using `cutoffIndex`, `destinationPointSetId`, `percentile` directly? Perhaps to avoid replicating the logic for the annotations (`C`, `P`)?

src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
```diff
@@ -156,18 +168,181 @@ private int getIntQueryParameter (Request req, String parameterName, int default
...
+                int d = 0;
+                for (String destinationPointSetId : analysis.destinationPointSetIds) {
+                    OpportunityDataset opportunityDataset = Persistence.opportunityDatasets.findByIdIfPermitted(
+                            destinationPointSetId,
+                            userPermissions
+                    );
+                    checkNotNull(opportunityDataset, "Opportunity dataset could not be found in database.");
+                    // Avoid name collisions, prepend an integer.
+                    String friendlyDestinationName = "D%d_%s".formatted(d++, filenameCleanString(opportunityDataset.name));
+                    friendlyReplacements.put(destinationPointSetId, friendlyDestinationName);
```
ansoncfit, line 285: The destination name is used in this new method here, and (part of the) UUID is used in the revised `getRegionalResults` below.

Would a unified method, with the last six characters of the UUID appended to the name, be useful in both places?

src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
```diff
@@ -550,10 +674,12 @@ public void registerEndpoints (spark.Service sparkService) {
         });
         sparkService.path("/api/regional", () -> {
             // For grids, no transformer is supplied: render raw bytes or input stream rather than transforming to JSON.
+            // Wait, do we have any endpoints that write grids into responses? It looks like they all return URLs now.
```
ansoncfit, line 677: See `TimeGridWriter.writeGeotiff`?

ansoncfit (commented):
I tested a few things with the UI locally, ensuring that pre-existing download options for opportunity datasets and regional analyses continue to work.

I would like to consider including the last six characters of UUIDs in the human readable names, which would be a consistent way for users to distinguish between inputs when names are not unique. And protip: entering part of UUID in most of the selectors in the UI works for filtering.

src/main/java/com/conveyal/file/FileStorageFormat.java
```diff
@@ -37,7 +39,12 @@ public enum FileStorageFormat {
...
     public static FileStorageFormat fromFilename (String filename) {
-        String extension = filename.substring(filename.lastIndexOf(".") + 1);
-        return FileStorageFormat.valueOf(extension.toUpperCase());
+        String extension = filename.substring(filename.lastIndexOf(".") + 1).toLowerCase(Locale.ROOT);
+        for (FileStorageFormat format : FileStorageFormat.values()) {
+            if (format.extension.equals(extension)) {
+                return format;
+            }
+        }
+        return null;
```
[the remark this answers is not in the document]
ansoncfit: Previous version did not work correctly, but it was only used for retrieving pre-existing data in the local file server implementation.

src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
```diff
@@ -228,89 +402,36 @@ private Object getRegionalResults (Request req, Response res) throws IOException
...
+        int secondUnderscore = singleCutoffFileStorageKey.path.indexOf('_', firstUnderscore + 1);
+        int lastDot = singleCutoffFileStorageKey.path.lastIndexOf('.');
+        String humanName = analysis.name +
+                singleCutoffFileStorageKey.path.substring(firstUnderscore, firstUnderscore + 6) +
+                singleCutoffFileStorageKey.path.substring(secondUnderscore, lastDot);
+        return fileStorage.getJsonUrl(singleCutoffFileStorageKey, humanName, format.extension);
     }
 
-    private String getCsvResults (Request req, Response res) {
+    private Object getCsvResults (Request req, Response res) {
```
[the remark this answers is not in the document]
abyrd: I'd usually say yes, but in this particular case a reference to this method is being used as a `@FunctionalInterface spark.Route`, and the actual type of the method in that interface is `Object handle(Request request, Response response)`.

src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
abyrd: I think the reason the two use different naming schemes is that including the user-specified name of the destination set means cascading N database lookups to find those names. I was willing to introduce this potential source of lag or bugs in the multi-grid case. It needs to zip things and move files around so might be expected to lag a little. The multi-grid case was also new and intended to be manually triggered.

On the other hand, I was trying to keep the existing single-cutoff codepath more stable, so didn't want to cascade a new database operation there.

In the multi-grid case, I don't really like this UUID-to-name string substitution approach. Practically speaking it works fine. The principle behind it is even reasonable (a one way mapping from totally unique ID strings are equivalent to specific sanitized human-readable strings). But it feels like a workaround, avoiding producing the human readable names in the loop and associating them with the keys.

Assuming we're willing to cascade the database fetch in the single-grid case, what might make sense is to factor out not only a function that generates the short dataset name, but one that creates the whole human-readable name of a single-cutoff grid, then reuse both of these functions in both the single and multi-grid use cases. There's a slight complication in how the extension of the human-readable filename is kept separate in case the raw human readable name is truncated. Maybe this calls for a new data type combining or associating a FileStorageKey with a separate human-readable name and extension, or that truncates and combines the raw human readable name and extension in its constructor.

src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
abyrd: But now I'm wondering what causes the RegionalAnalysis object returned by getRegionalAnalysis to be serialized properly.

src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
abyrd: https://www.mongodb.com/docs/manual/reference/method/ObjectId/ The first four and middle five bytes are slow-changing and would not disambiguate between data sets. Only the least significant digits of counter at the end will be sure to change.

src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
```diff
@@ -156,18 +168,181 @@ private int getIntQueryParameter (Request req, String parameterName, int default
...
      */
-    private Object getRegionalResults (Request req, Response res) throws IOException {
+    private UrlWithHumanName getRegionalResults (Request req, Response res) throws IOException {
 
         // Get some path parameters out of the URL.
         // The UUID of the regional analysis for which we want the output data
         final String regionalAnalysisId = req.params("_id");
         // The response file format: PNG, TIFF, or GRID
-        final String fileFormatExtension = req.params("format");
+        final String fileFormatExtension = req.params("format"); // FIXME this may be case sensitive (could make multiple files for different requests)
```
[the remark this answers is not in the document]
abyrd: The PR is now performing all necessary case conversions. We rearranged some code and comments to make this more clear.

src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
abyrd: We looked at this today, and RegionalAnalysis inherits a toString method that produces JSON. Leaving as-is.

abyrd:
Recent commits factor out `humanNameForEntity` method that creates names in the format `User_assigned_name_a1b2c3` for any Model instance, including RegionalAnalysis and OpportunityDataset. This is used by a revised `getSingleCutoffGrid` method, which returns a new record type `(FileStorageKey storageKey, String humanName)`. The `getSingleCutoffGrid` method now also receives the cutoff in minutes and looks up the index into the cutoffs array as needed. The same approach to generating human readable names for grid files is used everywhere, in both the bulk zip download and retrieving individual files.

ansoncfit (approved):

abyrd:
Although this has been merged, I'm adding follow-up observations here to keep them together with the original discussion. On today's call we identified a problem with the multi-file ZIP download. In regional analysis with millions of origin points, creating the zip file can be quite slow, taking over 2 minutes. It's not appropriate to stall an HTTP request for so long with no indication of progress. In the absence of feedback users are likely to re-initiate the creation/download process multiple times from a command line, script, or UI. This could lead to multiple stalled HTTP handler threads all creating the same grid and zip files and overwriting each other's results.

In initial testing I had not noticed this slowness, so the first step was to identify which part of the process is slow. This could be extracting the single-cutoff values from the multi-cutoff regional results file, encoding the geotiff file (or grid file, which is similarly slow), or transferring these resulting files to S3. We don't have enough log messages to immediately see the timing of each step. But by comparing file modification times (with `ls --full-time` for high time resolution) and log messages, about 13 seconds elapse between the log message about starting to extract the single-cutoff grid and the geotiff file appearing on the filesystem. The file is timestamped one second later on S3. So apparently S3 transfer is not the culprit, and it's either extracting the single-cutoff data or encoding/compressing the geotiff and grid file that is slow. These files are only about 3-6MB so this is not expected to be slow at all - we need to investigate this. Maybe we're using some un-buffered file IO?

Assuming creating these files remains "slow" (more than a few seconds), the solution will probably involve a mechanism like the one we use in network building, where an HTTP response is always immediate. If the object exists, it's immediately returned. Otherwise the background process is started and we respond with a `202 ACCEPTED` (see AnalysisWorkerController.handleSinglePoint L68). But the AsyncLoader mechanism used for networks is not very appropriate for this situation: the results of the multi-file ZIP operation are not held as an object in memory, and though that AsyncLoader aims to be reusable/generic, it is a cacheing mechanism incorporating some network-build specific ideas.

Aside: as noted in AnalysisWorker#handleAndSerializeOneSinglePointTask, "Ideally we'd stall briefly using something like Future.get(timeout) in case loading finishes quickly" so not all requests for newly created entities result in a 202 and retry, but that's another topic.

abyrd:
Looking into the speed of individual multi-cutoff to single-cutoff conversions. I added some log statements around the lines that produce the single-cutoff files and ran a regional analysis for Portland (57k origins). It takes around 200ms to derive a single-cutoff file. This is 52x smaller than the original culprit regions with around 3M origins. 52 * 200 ms gives 10s, roughly aligning with experience on those larger regions.

Wrapping input and output with buffered streams makes little to no difference, probably because input and output are both passing through gzip streams which contain their own buffers. Sampling CPU in VisualVM shows `inflaterInputStream.read` as the most time-consuming call. Reading time is about double writing time.

Unzipping the source `.access` file and removing the gunzip input stream halves the time to 100 ms instead of 200 ms. Proportions are reversed, with 2/3 of time now in writing instead of 2/3 in reading. Removing the gzip output stream (in addition to the gunzip input stream) still produces files usable by the UI. This drops the conversion time to around 30 ms, so an 85% decrease relative to the original 200 ms. The input and output files are observed to be about 3x bigger without compression.

Using command line tools, gunzipping and re-gzipping the entire multi-cutoff source `.access` file source data takes 10 and 50 ms respectively. This is much faster than the 170 ms compression seems to be contribute in the Java case.

Initial conclusion is that streaming compression/decompression is the main culprit,  perhaps in how it interacts with the byte-by-byte little-endian DataInputStream. It would be interesting to see how much faster this would be if performed with NIO byte buffers.

PR #935, opened 2024-03-15
Dual access for grid origins

ansoncfit:
This is a temporary experimental feature, (ab)using existing features in the UI and backend so that dual access results for grid origins can be obtained without any changes to those other components of our system.

~~To activate via the UI, add `flags: {"gridDualAccess"}` to the analysis request JSON~~. With https://github.com/conveyal/r5/pull/935/commits/10a51a836004af3a6d12a87b471c770d98424adb, regional analyses are hard-coded to produce dual access results. When configuring the regional analysis, enter the desired dual access thresholds in the cutoffs field. Also enter 120 as the maximum cutoff. If you add `dualAccessibilityThreshold` to the analysis request JSON, it will be used in place of 120.

For example, if you select hospitals, grocery stores, and jobs as destination layers; 5, 10, and 120 as cutoffs; and 10000 as `dualAccessibilityThreshold`, from the regional results page you will be able to view and download grids showing the travel time required to reach 5 hospitals, 10 grocery stores, and 10k jobs (as well as the other combinations). For now though, the UI selector on the regional analysis result page will be misleadingly labeled (with the "minutes" unit, and with 120 instead of 10000 as the third value).

Thresholds of 1 or 3 might be common, but the UI currently imposes a minimum of 5 minutes.

### DIFF
--- a/src/main/java/com/conveyal/analysis/results/MultiOriginAssembler.java
+++ b/src/main/java/com/conveyal/analysis/results/MultiOriginAssembler.java
@@ -139,7 +139,8 @@ public class MultiOriginAssembler {
                     // We might want to record a grid of dual accessibility values, but this will require some serious
                     // refactoring of the GridResultWriter.
                     // if (job.templateTask.dualAccessibilityThreshold > 0) { ... }
-                    throw new IllegalArgumentException("Temporal density of opportunities cannot be recorded for gridded origin points.");
+                    // throw new IllegalArgumentException("Temporal density of opportunities cannot be recorded for " +
+                    //        "gridded origin points.");
                 } else {
                     // Freeform origins.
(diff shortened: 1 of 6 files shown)


abyrd:
The checks we discussed are in `AccessibilityResult#getIntValues`, where `AccessibilityResult#checkInvariants` is called as the first step when the accessibility values are pulled out of the AccessibilityResult into the single point response or RegionalWorkResult. By calling the separate function fakeDualAccess the checks are bypassed. 

As a step toward making gridded dual access a first-class feature, we might want separate checkPrimalAccessInvariants and checkDualAccessInvariants methods. The checkDualAccessInvariants method might need to be called at the end of fakeDualAccess after the final return values are prepared.

ansoncfit:
Superseded by #955 and following.

PR #936, opened 2024-03-21
Add feed id column to path results (sequel)

ansoncfit:
Supersedes https://github.com/conveyal/r5/pull/927 (where we encountered merge problems on GitHub)


abyrd (approved):
Approved previously as #927 but there was some problem with merge.

PR #938, opened 2024-04-11
Schedule creation of geotiff zip files in the background

abyrd:
Discussion and observations at https://github.com/conveyal/r5/pull/932#issuecomment-2048894228.

It may be possible to make creation of single-cutoff geotiffs and grids faster (e.g. using NIO buffers). But for the time being it can take several minutes to produce and archive all the geotiffs for a single large regional analysis. This PR is a working proof of concept to perform this operation in the background. The process of making the zip file has been moved almost verbatim into a Task.withAction(). Usage:

```
> curl 'http://localhost:7070/api/regional/66179beec42a7657af4fdd2d/all' -v
HTTP/1.1 202 Accepted
"Building ZIP in background."

> curl 'http://localhost:7070/api/regional/66179beec42a7657af4fdd2d/all' -v
HTTP/1.1 202 Accepted
"ZIP is already being prepared in the background."

> curl 'http://localhost:7070/api/regional/66179beec42a7657af4fdd2d/all' -v
HTTP/1.1 200 OK
{"url":"http://localhost:7070/files/results/66179beec42a7657af4fdd2d_ALL.zip","humanName":"TEST_REGIONAL_2_4fdd2d.zip"}
```

Progress is reported as each file is created:
image

### DIFF
--- a/src/main/java/com/conveyal/analysis/components/BackendComponents.java
+++ b/src/main/java/com/conveyal/analysis/components/BackendComponents.java
@@ -86,7 +86,7 @@ public abstract class BackendComponents {
                 new GtfsController(gtfsCache),
                 new BundleController(this),
                 new OpportunityDatasetController(fileStorage, taskScheduler, censusExtractor, database),
-                new RegionalAnalysisController(broker, fileStorage),
+                new RegionalAnalysisController(broker, fileStorage, taskScheduler),
                 new AggregationAreaController(fileStorage, database, taskScheduler),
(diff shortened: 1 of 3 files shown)


ansoncfit (approved):
This looks good, thanks.

For the next release cycle, we expect this to be used as an experimental feature. After that, we may want to add mechanisms to clear entries from `filesBeingPrepared` if errors occur (user triggers activity cancellation, or some timeout). We could design those mechanisms more generally, to handle other background tasks such as network building.

PR #939, opened 2024-04-17
Add flag to crop regional analysis destinations to origin extents

ansoncfit:
This PR adds a `flag` (`CROP_DESTINATIONS`) to crop destination grids to the web mercator extents used for origins. This setting can be useful for reducing computation, especially for auto analyses in large regions.


abyrd (approved):
This is implemented as a flag, so it's considered a more off-label feature. Presence of the flag only changes conditional branch selection on a small, readily identifiable bit of code in a way that's already used in Taui site generation. The method `RegionalTask#getWebMercatorExtents` appears to only be called from methods run by workers. For these reasons, the risk of adverse effects on the main backend system is low.

We noted that although this is called CROP_DESTINATIONS it just makes the destination grid match the origin grid, which could make the travel time grid larger rather than smaller. But the grids containing the opportunity densities are effectively cropped by wrapping with GridTransformWrapper. We verified that these GridTransformWrappers are applied on a per-grid basis, not in some way that would make assumptions about the conditional code changed in this PR.

Looks good.

PR #942, opened 2024-10-02
Update to latest mongodb legacy java driver

abyrd:
This is the newest driver, but with an older ("legacy") programming API that matches our code. It should be a drop-in replacement, and our Mongojack dependency also seems to cooperate with this as its transitive mongodb driver dependency.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -164,7 +164,7 @@ dependencies {
     }
 
     // Database driver.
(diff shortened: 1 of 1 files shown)


trevorgerhardt (approved):
Initial testing with MongoDB 7.0 locally worked for me 👍

Issue #751, opened 2021-09-26
Saving shape information in network.dat (r5r)


mvpsaraiva:
R5 currently doesn’t save shape information into the network.dat, which is useful for `PointToPointQuery` used by `r5r`’s `detailed_itineraries() `function. This could be easily changed in [this line,](https://github.com/conveyal/r5/blob/69329442551bff50ea528f6a21c141cf344e0e6f/src/main/java/com/conveyal/r5/transit/TransitLayer.java#L70) by making `SAVE_SHAPES` not final so we could change it at runtime from `r5r`. 

Perhaps if you prefer a more robust implementation, like passing a parameter via `TNBuilderConfig` (#644), I'm happy to help.

ansoncfit:
Thanks, @mvpsaraiva. 

Including `saveShapes` as an option in a new `networkBuildConfig` would be interesting. Beyond `PointToPoint` mode, this could also allow for more precise path display in Conveyal's publishable stakeholder engagement sites. Without saving shapes, these sites fall back to stop-to-stop lines:

https://github.com/conveyal/r5/blob/3c63f05d72fc2e0fa4f59e4e3498366e5d5e129e/src/main/java/com/conveyal/r5/transit/TripPattern.java#L193-L205

abyrd:
I think this constant `SAVE_SHAPES` exists because we had the functionality to save shapes and didn't need it for any current use case, but we didn't want to remove the code, expecting we would have use cases in the future. It is just a hardcoded placeholder parameter. It makes sense to me to promote this to a TNBuilderConfig field.

Issue #643, opened 2020-11-01
Option to disallow walking on street when separate sidewalks present


ansoncfit:
In some places, OSM has sidewalks mapped separately from the main carriageways:

> When a sidewalk is mapped separately, add the tag sidewalk=separate to the carriageway it borders – this prevents mappers from describing the sidewalk twice. https://wiki.openstreetmap.org/wiki/Guidelines_for_pedestrian_navigation#Sidewalks_and_crossings

It's straightforward to disallow walking on such carriageways, but we should allow users to specify whether they want to build networks with the stricter walking traversal permissions, because it may affect results in unexpected ways.

PR #943, opened 2024-10-02
Add custom walking traversal permissions via TransportNetworkConfig

ansoncfit:
Users want the ability to configure different traversal permission labeling approaches (e.g., disallowing walking on any links that allow driving, under the assumption complete and separate sidewalk networks are available, see https://github.com/conveyal/r5/issues/643) 

This PR should be considered together with #941. Together with those changes, this PR allows users to build networks using a new experimental `SidewalkTraversalPermissionLabeler` with stricter walking permissions that are likely to be appropriate as more complete sidewalk networks become available.

Based on discussions in a work session a couple months ago, this PR removes the `buildNetworkFromBundleZip` method that has been unsupported since 2016.

This PR partly addresses https://github.com/conveyal/r5/issues/751 by passing `TransportNetworkConfig` values in `PointToPointRouterServer` use. The remaining piece of that r5r-related issue is allowing users to specify whether to save detailed GTFS shapes (which will be useful for us in publishing Taui sites, see #644)

### DIFF
--- a/src/main/java/com/conveyal/r5/streets/StreetLayer.java
+++ b/src/main/java/com/conveyal/r5/streets/StreetLayer.java
@@ -6,12 +6,14 @@ import com.conveyal.osmlib.OSM;
 import com.conveyal.osmlib.OSMEntity;
 import com.conveyal.osmlib.Relation;
 import com.conveyal.osmlib.Way;
+import com.conveyal.r5.analyst.cluster.TransportNetworkConfig;
 import com.conveyal.r5.analyst.scenario.PickupWaitTimes;
 import com.conveyal.r5.api.util.BikeRentalStation;
 import com.conveyal.r5.api.util.ParkRideParking;
 import com.conveyal.r5.common.GeometryUtils;
 import com.conveyal.r5.labeling.LevelOfTrafficStressLabeler;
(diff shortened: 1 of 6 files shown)


src/main/java/com/conveyal/r5/streets/StreetLayer.java
```diff
@@ -209,6 +211,16 @@ public StreetLayer() {
         speedLabeler = new SpeedLabeler(SpeedConfig.defaultConfig());
     }
 
+    public StreetLayer(TransportNetworkConfig config) {
+        this();
+        if (config != null && config.traversalPermissionLabeler != null) {
+            // TODO cleaner mapping
```
abyrd, line 217: This can be done with switch expressions, for example:

```
        if (config != null) {
            permissionLabeler = switch (config.traversalPermissionLabeler) {
                case "sidewalk" -> new SidewalkTraversalPermissionLabeler();
                case null -> new USTraversalPermissionLabeler();
                default -> throw new IllegalArgumentException("Unknown traversal permission labeler: " + config.traversalPermissionLabeler);
            };
        } else {
            permissionLabeler = new USTraversalPermissionLabeler();
        }
```

trevorgerhardt (approved):
Config key appears to work as intended when testing walk times.

PR #944, opened 2024-10-04
Check if the bundle config exists, return 404 if missing

trevorgerhardt:
When retrieving the config, return a clearer error if the file cannot be found so that can be handled individually.

### DIFF
--- a/src/main/java/com/conveyal/analysis/controllers/BundleController.java
+++ b/src/main/java/com/conveyal/analysis/controllers/BundleController.java
@@ -18,8 +18,8 @@ import com.conveyal.gtfs.model.Stop;
 import com.conveyal.gtfs.validator.PostLoadValidator;
 import com.conveyal.osmlib.Node;
 import com.conveyal.osmlib.OSM;
-import com.conveyal.r5.analyst.progress.ProgressInputStream;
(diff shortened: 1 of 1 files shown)


src/main/java/com/conveyal/analysis/controllers/BundleController.java
```diff
@@ -327,13 +327,17 @@ private Bundle getBundle (Request req, Response res) {
      * and is considered the definitive source of truth. The entry in the database is a newer addition and has only
      * been around since September 2024.
      */
-    private TransportNetworkConfig getBundleConfig (Request request, Response res) {
+    private TransportNetworkConfig getBundleConfig(Request request, Response res) {
```
ansoncfit, line 330: Different style settings in IDE?

ansoncfit (approved):
For future reference, the 404 return code is not explicit in this controller -- it's specified in `AnalysisServerException#notFound()`

src/main/java/com/conveyal/analysis/controllers/BundleController.java
trevorgerhardt: I guess so? I did not intentionally change that. Similarly with the moving import, probably from "organize imports" on save.

PR #945, opened 2024-10-10
Add network config option to save GTFS shapes

### LINKED ISSUE #751: Saving shape information in network.dat (r5r)

ansoncfit:
Fixes #751

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/cluster/TransportNetworkConfig.java
+++ b/src/main/java/com/conveyal/r5/analyst/cluster/TransportNetworkConfig.java
@@ -57,4 +57,9 @@ public class TransportNetworkConfig {
      */
     public String traversalPermissionLabeler;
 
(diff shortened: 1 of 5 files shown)


trevorgerhardt:
Changes look good to me. Do we know why those tests are failing? Once we address those I will approve.

trevorgerhardt (dismissed):

PR #947, opened 2024-10-25
Update javadoc to reflect recent LinkageCache changes

abyrd:
This is a pure documentation change, fixing some obsolete information I noticed during a recent call.

The final block of documentation in the changeset was changed only for consistency with surrounding comments. It will take longer to work through the open questions in that block and verify current behavior. That would be useful, but I don't think I will reach reliable conclusions on that part by the time we do the release.

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/LinkageCache.java
+++ b/src/main/java/com/conveyal/r5/analyst/LinkageCache.java
@@ -21,7 +21,7 @@ import java.util.concurrent.ExecutionException;
  * There could instead be one instance per AnalystWorker or per JVM (static), but this would cause the mappings
  * including PointSets, StreetLayers, and Linkages (which hold references to the TransportNetwork) to stick around
  * even when we try to garbage collect a TransportNetwork. In cloud operation, this problem would not necessarily arise
- * in practice since workers are permanently associated with a single TransportNetwork.
+ * in practice since workers are permanently associated with a single base TransportNetwork.
  */
 public class LinkageCache {
 
--- a/src/main/java/com/conveyal/r5/analyst/LinkageCache.java
+++ b/src/main/java/com/conveyal/r5/analyst/LinkageCache.java
@@ -16,35 +16,34 @@ import java.util.Objects;
 import java.util.concurrent.ExecutionException;
 
 /**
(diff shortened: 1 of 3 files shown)


src/main/java/com/conveyal/r5/analyst/LinkageCache.java
```diff
@@ -16,35 +16,34 @@
...
- * Retains linkages between PointSets and the StreetLayers for specific StreetModes.
- * This used to be embedded in the PointSets themselves, now there should be one instance per TransportNetwork.
+ * Retains linkages between PointSets and the StreetLayers for specific StreetModes, including egress distance tables.
+ * LinkageCaches used to be associated with individual PointSets, but now there is a single cache per TransportNetwork.
  * There could instead be one instance per AnalystWorker or per JVM (static), but this would cause the mappings
  * including PointSets, StreetLayers, and Linkages (which hold references to the TransportNetwork) to stick around
- * even when we try to garbage collect a TransportNetwork. This is less of an issue now that we don't plan to have
- * workers migrate between TransportNetworks.
+ * even when we try to garbage collect a TransportNetwork. In cloud operation, this problem would not necessarily arise
+ * in practice since workers are permanently associated with a single TransportNetwork.
```
ansoncfit, line 24:
```suggestion
 * in practice since workers are permanently associated with a single base TransportNetwork.
```

ansoncfit (changes requested):
I made one small suggestion for clarity.

In our recent discussion, I mentioned a vague memory of another comment that might need to be updated, about moving something to a top-level cache. I tried to track this down but didn't find it.

abyrd:
>Is the description here still correct?

I think it is still correct, though the way it's phrased might be confusing. When it says "completely bypassing the async loader" it means that while we're still using the async loader class, we're using a method on that class that intentionally skips all the async or background proccessing machinery.

When a regional analysis starts up, we have N tasks all trying to preload at once, and they all block on this method. We don't want N threads performing the slow or heavy calculation steps of preloading redundantly. Nothing in the async loader class itself is causing them to take turns or wait for one preload to finish and share with the other threads. But behind the preloader class there are other caches that do enforce turn-taking and prevent redundant work.

ansoncfit (approved):
👍

PR #950, opened 2024-10-26
Set Status.PRESENT after synchronous preload

### LINKED ISSUE #949: Stalled single-point analysis after worker handles regional tasks
Once a worker completes regional tasks for a given combination of scenario and modes, subsequent single-point requests for that combination stall with the message "Linking destination grid to streets"

The issue appears to be that the synchronous preload used when handling regional tasks does not set status of loaded objects to "PRESENT"

ansoncfit:
This PR fixes #949, a longstanding issue (possibly since #746).

We have two pathways for loading transport network data on a worker: `preloadData` (asynchronously, for single-point tasks) and `synchronousPreload` (for regional tasks). Once data are loaded in an AsyncLoader, they should be marked with `Status.PRESENT`. The async pathway was doing this correctly. But once a worker switched to a regional task, it would switch the status to `LOADING` and never switch it back to `PRESENT`.

This PR factors out a `setComplete` method and calls it in both pathways.

### DIFF
--- a/src/main/java/com/conveyal/r5/analyst/NetworkPreloader.java
+++ b/src/main/java/com/conveyal/r5/analyst/NetworkPreloader.java
@@ -97,7 +97,10 @@ public class NetworkPreloader extends AsyncLoader<NetworkPreloader.Key, Transpor
      * data is prepared. 
      */
     public TransportNetwork synchronousPreload (AnalysisWorkerTask task) {
-        return buildValue(Key.forTask(task));
+        Key key = Key.forTask(task);
+        TransportNetwork scenarioNetwork = buildValue(key);
+        setComplete(key, scenarioNetwork);
+        return scenarioNetwork;
     }
 
     @Override
@@ -140,7 +143,7 @@ public class NetworkPreloader extends AsyncLoader<NetworkPreloader.Key, Transpor
                 linkedPointSet.getEgressCostTable(progressListener);
             }
         }
-        // Finished building all needed inputs for analysis, return the completed network to the AsyncLoader code.
+        // Finished building all needed inputs for analysis, return the completed network
         return scenarioNetwork;
     }
 
--- a/src/main/java/com/conveyal/r5/analyst/NetworkPreloader.java
+++ b/src/main/java/com/conveyal/r5/analyst/NetworkPreloader.java
@@ -79,7 +79,7 @@ public class NetworkPreloader extends AsyncLoader<NetworkPreloader.Key, Transpor
         this.transportNetworkCache = transportNetworkCache;
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/conveyal/r5/analyst/NetworkPreloader.java
```diff
@@ -140,7 +143,7 @@ protected TransportNetwork buildValue(Key key) {
                 linkedPointSet.getEgressCostTable(progressListener);
             }
         }
-        // Finished building all needed inputs for analysis, return the completed network to the AsyncLoader code.
```
ansoncfit, line 143: This method is also called by `synchronousPreload`, so I propose deleting this part of the comment.

abyrd:
Thanks @ansoncfit for identifying this problem and tracking down the cause! The proposed fix already looks good, but I have a sense we could reduce code duplication slightly, and reduce the risk of similar future problems by moving the synchronousPreload logic up into the AsyncLoader abstract class as a public getSynchronous() method, which is then also called by the runnable inside get(). This would have all code paths calling the exact same instance of the crucial buildValue / setComplete sequence, so it can't be done wrong in any future extensions of AsyncLoader. I'll give it some thought tomorrow and propose a patch.

abyrd:
What do you think of the approach shown in the three commits I just pushed to this branch? The NetworkPreloader has blocking and nonblocking preload methods, which call blocking and nonblocking methods on the base class. Then the nonblocking method on the base class just runs the blocking method in a background thread. So everything leads down to a single blocking get method that always includes the logic to mark the element present.

This should provide identical behavior to the previous version, I just find it a little less susceptible to similar problems in the future.

The question arises whether the try/catch error handling code should be inside getBlocking, which would make it apply to both preload and preloadSynchronous. Errors in either case would then set the status to ERROR. This would be more consistent but would interfere with the way we currently shuttle regional task exception messages back to the backend. So I just added some comments clarifying how it works currently.

abyrd:
Well, actually there was one small difference in behavior: the status was being set to "Starting..." just before a synchronous load begins. I moved it back just to keep behavior mostly identical to previous versions. This could be changed in the future if we do a more significant refactor of how exceptions are handled and passed up to the backend.

ansoncfit (commented):
Thanks for the revisions. I agree with having both paths converge on a single method.

I will make sure we test workers switching between scenarios and single-point/regional tasks on staging before our next release.

Let's go ahead and merge this PR (I would mark it as approved, but GH doesn't give me the option as the original author).

trevorgerhardt (dismissed):
Looks good to me, can be merged once the conflict is resolved 👍

PR #951, opened 2024-12-19
Patched v7.3 larger max extent for streetlayer

christophfink:
This is an attempt of finding a workaround for https://github.com/r5py/r5py/issues/447 
(ping @fbarbe00)


christophfink:
sorry, didn't want to merge this into conveyal/r5, but r5py/r5

PR #953, opened 2025-02-11
Enable returning freeform points

trevorgerhardt:
Allows `getOpportunityDataset` to return Freeform files to be visualized in the UI.


ansoncfit (approved):
👍

PR #955, opened 2025-02-17
Handle dual access grids (full implementation)

trevorgerhardt:
Create and write dual accessibility grids utilizing the same grid format as regular accessibility grids. A new `dualAccessibilityThresholds` parameter is required along with `includeTemporalDensity` set to true.

Changes:
- Switch to use a thresholds array instead of a single threshold in `AnalysisWorkerTask`s.
- Check that only a single grid is being created (they currently would have the same filename)
- Create a `MultiGridResultWriter` for dual access grids. Store dual access results in `accessibilityValues` in `RegionalWorkResult`.
- Ensure the worker version is >= v7.4.
- Switch `getRegionalResults` to take `threshold` as a parameter instead of `cutoff`.
- Requires https://github.com/conveyal/ui/pull/2065

Screenshot 2025-03-08 at 7 12 24 PM


src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
```diff
@@ -406,17 +409,26 @@ private UrlWithHumanName getRegionalResults (Request req, Response res) throws I
...
+            checkState(nThresholds > 0, "Regional analysis has no thresholds.");
+            threshold = getIntQueryParameter(req, "threshold", thresholds[nThresholds / 2]);
+            checkArgument(new TIntArrayList(thresholds).contains(threshold),
+                    "Dual access thresholds for this regional analysis must be taken from this list: (%s)",
+                    Ints.join(", ", thresholds)
+            );
+        } else if (analysis.cutoffsMinutes != null) {
+            // Handle newer regional analyses with multiple cutoffs in an array.
+            // If a query parameter is supplied, range check it, otherwise use the middle value in the list.
+            // The cutoff variable holds the actual cutoff in minutes, not the position in the array of cutoffs.
```
ansoncfit, line 427: Should this comment (specifically the part about using the middle value in the list) be moved up (as `[nThresholds / 2]` is also used above)?

src/main/java/com/conveyal/r5/analyst/cluster/AnalysisWorkerTask.java
```diff
@@ -111,9 +111,9 @@ public abstract class AnalysisWorkerTask extends ProfileRequest {
 
     /**
      * If this is set to a value above zero, report the amount of time needed to reach the given number of
```
ansoncfit, line 113: Javadoc here (and above for `includeTemporalDensity`) should be updated (as in `AnalysisRequest`)

src/main/java/com/conveyal/r5/analyst/TemporalDensityResult.java
```diff
@@ -41,56 +44,81 @@ public class TemporalDensityResult {
...
+    public int[][][] calculateDualAccessGrid() {
+        checkInvariants();
 
+        int[][][] dualAccessGrid = new int[nPointSets][nPercentiles][nThresholds];
+        for (int i = 0; i < nPointSets; i++) {
+            for (int j = 0; j < nPercentiles; j++) {
+                for (int k = 0; k < nThresholds; k++) {
+                    int threshold = dualAccessThresholds[k];
+                    int minutes = 0;
+                    double sum = 0;
```
ansoncfit, line 112: Current user-facing documentation states
> A value of -1 indicates the specified threshold is not reached within the maximum cutoff specified when the regional analysis was created

For a special value to signal the threshold cannot be reached in less than 120 minutes, I'd prefer a value other than 0.

src/main/java/com/conveyal/analysis/models/AnalysisRequest.java
```diff
@@ -164,17 +164,16 @@ public class AnalysisRequest {
...
+     * to the broker. Requires a destination pointset to be specified. If an origin pointset is specified the results
+     * will be collated to CSV.
      */
     public boolean includeTemporalDensity = false;
 
     /**
-     * If this is set to a value above zero, report the amount of time needed to reach the given number of
-     * opportunities from this origin (known technically as "dual accessibility").
+     * Report the amount of time needed to reach the given number of
+     * opportunities from this origin (known technically as "dual access").
```
ansoncfit, line 174:
```suggestion
     * If set, and if includeTemporalDensity is true, results will 
     * include the amount of time needed to reach each threshold 
     * number of opportunities (known technically as "dual access").
```

ansoncfit (commented):
Thanks for switching to the consistent "access" terminology. A few comments are in-line below.

src/main/java/com/conveyal/r5/analyst/TemporalDensityResult.java
trevorgerhardt: Updated in https://github.com/conveyal/r5/pull/955/commits/d8e68699ce196407065a78dc5f77d3a8f0d680d8

trevorgerhardt: If we set values to -1 then this statement will error when writing a grid:

https://github.com/conveyal/r5/blob/e535981ec29b0461f4d3d3111999a9d6f5ac6d7d/src/main/java/com/conveyal/r5/analyst/Grid.java#L303-L306

Possible solutions:
1. Use `0` to signify "unreached". Update the docs accordingly. Filter `0`s in the UI to show them as "unreached". 
2. Remove these checks while writing dual access Grids. Filter `-1` in the UI to show them as "unreached". 

Another thought: from a user's perspective, when we're showing values that represent minutes, do people usually expect `0` to mean "under 1" and `60` to mean under 61? For example, for regular access values we would show "Workers total within 60 minutes". So for dual access, expressing it as "threshold reached within 60 minutes" may also make sense.

ansoncfit: Possible solution # 1 seems reasonable. In geotiff downloads of single-point travel time results, 0 is already used to signify unreachable within 120 minutes (https://docs.conveyal.com/analysis#downloading)

src/main/java/com/conveyal/r5/analyst/cluster/AnalysisWorkerTask.java
ansoncfit: Could still use an update:
```
     * If set, and if includeTemporalDensity is true, results will 
     * include the amount of time needed to reach each threshold 
     * number of opportunities (known technically as "dual access").
```

src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
```diff
@@ -406,17 +409,26 @@ private UrlWithHumanName getRegionalResults (Request req, Response res) throws I
...
+                    Ints.join(", ", thresholds)
+            );
+        } else if (analysis.cutoffsMinutes != null) {
+            // Handle newer regional analyses with multiple cutoffs in an array.
+            // The cutoff variable holds the actual cutoff in minutes, not the position in the array of cutoffs.
             int nCutoffs = analysis.cutoffsMinutes.length;
             checkState(nCutoffs > 0, "Regional analysis has no cutoffs.");
-            cutoffMinutes = getIntQueryParameter(req, "cutoff", analysis.cutoffsMinutes[nCutoffs / 2]);
-            checkArgument(new TIntArrayList(analysis.cutoffsMinutes).contains(cutoffMinutes),
+            threshold = getIntQueryParameter(req, "cutoff", analysis.cutoffsMinutes[nCutoffs / 2]);
```
ansoncfit, line 430: I haven't thought this all the way through -- should all query parameters be `threshold` (regardless of whether the request is for accessibility or dual access results)?

```suggestion
            threshold = getIntQueryParameter(req, "threshold", analysis.cutoffsMinutes[nCutoffs / 2]);
```

trevorgerhardt: Yes. Updated in https://github.com/conveyal/r5/pull/955/commits/cb4334505b0f2237ae16555d9f574c545cb3af2c.

src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
```diff
@@ -599,6 +611,30 @@ private RegionalAnalysis createRegionalAnalysis (Request req, Response res) thro
                 "recordTimes can only be used with a single destination pointset, which must be freeform (non-grid)."
             );
         }
+        if (task.includeTemporalDensity) {
+            checkArgument(
+                    task.dualAccessThresholds != null &&
+                            task.dualAccessThresholds.length > 0,
+                    "dualAccessThresholds not specified when includeTemporalDensity is enabled."
```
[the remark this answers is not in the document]
trevorgerhardt: Yes, good catch. Updated in https://github.com/conveyal/r5/pull/955/commits/a69e71c15cd8effcc63ebab285639df9565bdf5d

src/main/java/com/conveyal/r5/analyst/TemporalDensityResult.java
trevorgerhardt: Changed to `0` in https://github.com/conveyal/r5/pull/955/commits/e6c6999ee8496aba8c2d8ec1981390a4480c149f

src/main/java/com/conveyal/r5/analyst/cluster/AnalysisWorkerTask.java
trevorgerhardt: Updated comment, thanks 👍

ansoncfit (approved):
Let's merge for a consolidated testing round.

src/main/java/com/conveyal/r5/analyst/TemporalDensityResult.java
```diff
@@ -41,56 +44,86 @@ public class TemporalDensityResult {
...
+                "Dual access requires at least one destination pointset."
         );
+        // TODO check that thresholds are sorted
         this.destinationPointSets = task.destinationPointSets;
+        this.dualAccessThresholds = task.dualAccessThresholds;
         this.nPercentiles = task.percentiles.length;
-        this.opportunityThreshold = task.dualAccessibilityThreshold;
-        this.opportunitiesPerMinute = new double[destinationPointSets.length][nPercentiles][120];
+        this.nPointSets = this.destinationPointSets.length;
+        this.nThresholds = this.dualAccessThresholds.length;
```
abyrd, line 54: If it is still valid to request temporal density results without enabling dual access, I believe `dualAccessThresholds` can be null, which will result in a null pointer exception here. Comments elsewhere say that dualAccessThresholds doesn't need to be set, and that calculating dual access is an optional add-on to temporal density.

src/main/java/com/conveyal/r5/analyst/cluster/RegionalWorkResult.java
```diff
@@ -65,7 +65,12 @@ public RegionalWorkResult(OneOriginResult result, RegionalTask task) {
         this.travelTimeValues = result.travelTimes == null ? null : result.travelTimes.values;
         this.accessibilityValues = result.accessibility == null ? null : result.accessibility.getIntValues();
         this.pathResult = result.paths == null ? null : result.paths.summarizeIterations(PathResult.Stat.MINIMUM);
-        this.opportunitiesPerMinute = result.density == null ? null : result.density.opportunitiesPerMinute;
+        if (result.density != null) {
+            this.opportunitiesPerMinute = result.density.opportunitiesPerMinute;
+            if (task.originPointSet == null) {
+                this.accessibilityValues = result.density.calculateDualAccessGrid();
```
abyrd, line 71: Relatedly, it looks like this could try to generate dual access grids even when no dual thresholds are specified, resulting in a null pointer exception in `calculateDualAccessGrid()`. However, here I think the `originPointSet == null` condition is excluding the case where we can calculate density without dual access (freeform pointsets specified as origins, producing CSV output).

src/main/java/com/conveyal/r5/analyst/TemporalDensityResult.java
trevorgerhardt: Hmmm, yes so "temporal density" can still be generated without thresholds being set. However, in practice we only generate "temporal density" to generate a dual access CSV, which requires a threshold to be set, and now dual access grids, which also require threshold to be set.

I may have responded too quickly to @ansoncfit's original comment about this issue when I moved the dual access thresholds validation. 

Will follow up with a PR to address.

trevorgerhardt: PR with dual access validation here: https://github.com/conveyal/r5/pull/961

src/main/java/com/conveyal/r5/analyst/cluster/RegionalWorkResult.java
trevorgerhardt: While this code path should not be hit without `dualAccessThresholds` being set, I added an additional check in the PR here: https://github.com/conveyal/r5/pull/961

src/main/java/com/conveyal/r5/analyst/TemporalDensityResult.java
abyrd: In practice in this particular context (regional analyses triggered from the web UI) we only use temporal density to generate access CSV or grids. But it could be used differently in single-point requests, or by API requests from scripts outside the standard web UI workflows. On the PR that originally added this functionality there's some discussion of other ways the temporal density output might be used: https://github.com/conveyal/r5/pull/884#issuecomment-1666745956

The combination in https://github.com/conveyal/r5/pull/961 seems appropriate for current capabilities. TemporalDensityResult can still be created and used without computing dual access (for example by batch single-point API requests from a script). But in regional analyses, I don't think there is any way for TemporalDensityResults to make their way into regional analysis results if dual access is not enabled. So it makes sense to validate that those dual access thresholds are present in incoming regional analysis API requests.

PR #956, opened 2025-02-18
Refactor: Remove deprecated / unused regional analysis fields

trevorgerhardt:
Remove old fields in regional analysis that haven't been used in many years. 

#957, which contains a MongoDB migration to clean up old regional analyses, should be merged and ideally run first.

We recently removed a similar worker version related flag from the UI here: https://github.com/conveyal/ui/pull/2064

### DIFF
--- a/src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
+++ b/src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
@@ -402,17 +402,8 @@ public class RegionalAnalysisController implements HttpController {
         final UserPermissions userPermissions = UserPermissions.from(req);
         RegionalAnalysis analysis = getAnalysis(regionalAnalysisId, userPermissions);
 
-        // Which channel to extract from results with multiple values per origin (for different travel time cutoffs)
-        // and multiple output files per analysis (for different percentiles of travel time and/or different
-        // destination pointsets). These initial values are for older regional analysis results with only a single
-        // cutoff, and no percentile or destination gridId in the file name.
-        // For newer analyses that have multiple cutoffs, percentiles, or destination pointsets, these initial values
-        // are coming from deprecated fields, are not meaningful and will be overwritten below from query parameters.
-        int percentile = analysis.travelTimePercentile;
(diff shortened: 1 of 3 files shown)


src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
```diff
@@ -220,38 +218,16 @@ private HumanKey getSingleCutoffGrid (
...
-            }
             LOG.debug("Single-cutoff grid {} not found on S3, deriving it from {}.", singleCutoffKey, multiCutoffKey);
 
-            InputStream multiCutoffInputStream = new FileInputStream(fileStorage.getFile(multiCutoffFileStorageKey));
+            InputStream multiCutoffInputStream = FileUtils.getInputStream(fileStorage.getFile(multiCutoffFileStorageKey));
             Grid grid = new SelectingGridReducer(cutoffIndex).compute(multiCutoffInputStream);
 
             File localFile = FileUtils.createScratchFile(fileFormat.toString());
-            FileOutputStream fos = new FileOutputStream(localFile);
+            OutputStream fos = FileUtils.getOutputStream(localFile);
```
abyrd, line 230: Looks like some of the methods that take `fos` as a parameter (at least Grid.write) don't use buffered output streams, so using this utlity method that buffers the output should greatly improve write speeds.

src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
```diff
@@ -220,38 +218,16 @@ private HumanKey getSingleCutoffGrid (
...
 
             switch (fileFormat) {
```
abyrd, line 232: This switch is not exhaustive, and when it receives an unexpected value it just falls through without writing anything to the file (or closing it). We should probably convert this to the new exhaustive switch syntax with a default block and no "break".

ansoncfit (approved):
With the migration in #957 run and the changes in this PR merged, the `travelTimePercentile` field will be unused and completely replaced by `travelTimePercentiles`.

After a version including this PR is deployed, we will run a separate migration removing the `travelTimePercentile` field from the database.

PR #957, opened 2025-02-26
Generate old regional analysis results

trevorgerhardt:
Generate all final results for old regional analyses with a task so that we can eliminate old code which was required to handle them. See #956 for the R5 code that can be simplified.

Regional analysis results had two changes to how they were stored over the years. However, the last change was several years ago. We still handle the old style of results in multiple locations in code, but if we migrate the database entries and pre-generate all of the regional analysis results we would be able to eliminate those code paths. This will make future changes easier to make.

This should be run on staging and production before #956 is merged.

### DIFF
--- a/src/main/java/com/conveyal/analysis/GenerateRegionalAnalysisResults.java
+++ b/src/main/java/com/conveyal/analysis/GenerateRegionalAnalysisResults.java
@@ -64,15 +64,15 @@ public class GenerateRegionalAnalysisResults implements TaskAction {
                         }
                     }
                 }
-                LOG.info("Finished generating files for {} of {}. Migrating db entry now...", regionalAnalysis._id, regionalAnalysis.accessGroup);
+                /* LOG.info("Migrating db entry for {} of {} now...", regionalAnalysis._id, regionalAnalysis.accessGroup);
 
                 // Save as modern types
                 regionalAnalysis.cutoffsMinutes = cutoffs;
                 regionalAnalysis.travelTimePercentiles = percentiles;
                 regionalAnalysis.destinationPointSetIds = destinationPointSetIds;
-                Persistence.regionalAnalyses.put(regionalAnalysis);
+                Persistence.regionalAnalyses.put(regionalAnalysis); */
 
-                LOG.info("Finished processing {} of {}.", regionalAnalysis._id, regionalAnalysis.accessGroup);
+                LOG.info("Finished generating files for {} of {}.", regionalAnalysis._id, regionalAnalysis.accessGroup);
             }
         } catch (Exception e) {
             LOG.error(ExceptionUtils.shortAndLongString(e));
--- a/src/main/java/com/conveyal/analysis/GenerateRegionalAnalysisResults.java
+++ b/src/main/java/com/conveyal/analysis/GenerateRegionalAnalysisResults.java
@@ -1,7 +1,6 @@
 package com.conveyal.analysis;
 
 import com.conveyal.analysis.controllers.RegionalAnalysisController;
-import com.conveyal.analysis.models.OpportunityDataset;
 import com.conveyal.analysis.models.RegionalAnalysis;
 import com.conveyal.analysis.persistence.Persistence;
 import com.conveyal.file.FileStorage;
@@ -48,14 +47,13 @@ public class GenerateRegionalAnalysisResults implements TaskAction {
 
                 // Iterate through all values and generate all possible formats for them.
                 for (String destinationPointSetId : destinationPointSetIds) {
-                    OpportunityDataset destinations = Persistence.opportunityDatasets.get(destinationPointSetId);
                     for (int cutoffMinutes : cutoffs) {
                         for (int percentile : percentiles) {
                             for (FileStorageFormat format : validFormats) {
                                 RegionalAnalysisController.getSingleCutoffGrid(
                                         fileStorage,
                                         regionalAnalysis,
-                                        destinations,
+                                        destinationPointSetId,
                                         cutoffMinutes,
                                         percentile,
                                         format
--- a/src/main/java/com/conveyal/analysis/GenerateRegionalAnalysisResults.java
+++ b/src/main/java/com/conveyal/analysis/GenerateRegionalAnalysisResults.java
@@ -51,15 +51,20 @@ public class GenerateRegionalAnalysisResults implements TaskAction {
                     for (int cutoffMinutes : cutoffs) {
                         for (int percentile : percentiles) {
                             for (FileStorageFormat format : validFormats) {
-                                RegionalAnalysisController.getSingleCutoffGrid(
-                                        fileStorage,
-                                        regionalAnalysis,
-                                        destinationPointSetId,
-                                        cutoffMinutes,
-                                        percentile,
-                                        format
-                                );
-                                filesGenerated++;
+                                try {
+                                    RegionalAnalysisController.getSingleCutoffGrid(
+                                            fileStorage,
+                                            regionalAnalysis,
+                                            destinationPointSetId,
+                                            cutoffMinutes,
+                                            percentile,
+                                            format
+                                    );
+                                    filesGenerated++;
+                                } catch (Exception e) {
+                                    LOG.error("Error generating single cutoff grid for {}", regionalAnalysis._id);
+                                    LOG.error(ExceptionUtils.shortAndLongString(e));
+                                }
                             }
                         }
                     }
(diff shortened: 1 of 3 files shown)


abyrd:
It does seem like a good simplification to get rid of this fallback code for very old formats. Removing some of these fields means old worker versions that rely on them would break, but these are very old versions that are already effectively deprecated and shouldn't be used.

Just thinking through how this works: In recent years we generate multi-cutoff results files, from which single-cutoff files are generated in multiple formats. The multi-cutoff files are never used directly; they are only used to generate single-cutoff files on demand. For old analyses where multi-cutoff results files do not exist (or are named differently because analyses were limited to one set of destinations), this PR doesn't attempt to generate new-style results files. It just hits all code paths where multi-cutoff results would normally be used to derive single-cutoff results files, hitting the fallback code that works in the absence of new-style results files. Once that's done, it should be possible to remove that fallback code. This is assuming we've exhaustively hit all the ways the functions containing the fallback code can be invoked.

What's the planned way to run this on staging and production before #956 is merged? I guess we would perform one deployment of the patched version, observe the logs, and perform another deployment with the post-956 version soon thereafter if everything looked good?

src/main/java/com/conveyal/analysis/GenerateRegionalAnalysisResults.java
```diff
@@ -0,0 +1,64 @@
...
+                RegionalAnalysis regionalAnalysis = cursor.next();
+                int[] percentiles = Objects.requireNonNullElseGet(regionalAnalysis.travelTimePercentiles, () -> new int[]{regionalAnalysis.travelTimePercentile});
+                int[] cutoffs = Objects.requireNonNullElseGet(regionalAnalysis.cutoffsMinutes, () -> new int[]{regionalAnalysis.cutoffMinutes});
+                String[] destinationPointSetIds = Objects.requireNonNullElseGet(regionalAnalysis.destinationPointSetIds, () -> new String[]{regionalAnalysis.grid});
+
+                // Save as modern types
+                regionalAnalysis.cutoffsMinutes = cutoffs;
+                regionalAnalysis.travelTimePercentiles = percentiles;
+                regionalAnalysis.destinationPointSetIds = destinationPointSetIds;
+                Persistence.regionalAnalyses.put(regionalAnalysis);
```
abyrd, line 41: #956 mentions a MongoDB migration, and it looks like this PR is intended to perform that migration (via Java code) while generating the missing single-cutoff results files. If so we might want to mention that in #956 for future reference.

src/main/java/com/conveyal/analysis/GenerateRegionalAnalysisResults.java
```diff
@@ -0,0 +1,64 @@
...
+
+                // Iterate through all values and generate all possible formats for them.
+                for (String destinationPointSetId : destinationPointSetIds) {
+                    OpportunityDataset destinations = Persistence.opportunityDatasets.get(destinationPointSetId);
+                    for (int cutoffMinutes : cutoffs) {
+                        for (int percentile : percentiles) {
+                            for (FileStorageFormat format : FileStorageFormat.values()) {
```
abyrd, line 48: It looks like this is iterating through all 12 or so FileStorageFormat values and calling `getSingleCutoffGrid` with all of them. The `getSingleCutoffGrid` method only recognizes and handles three of those formats (`GRID`, `PNG`, and `GEOTIFF`) and does not write anything to the output stream for any other cases. So it looks like this might leave the output stream open and try to move an underlying empty file into storage with each of the nine other extensions.

src/main/java/com/conveyal/analysis/GenerateRegionalAnalysisResults.java
```diff
@@ -0,0 +1,64 @@
...
+    @Override
+    public void action(ProgressListener progressListener) throws Exception {
+        DBObject query = QueryBuilder.start().or(
+                QueryBuilder.start("travleTimePercentiles").is(null).get(),
+                QueryBuilder.start("cutoffsMinutes").is(null).get(),
+                QueryBuilder.start("destinationPointSetIds").is(null).get()
+        ).get();
+        try (DBCursor<RegionalAnalysis> cursor = Persistence.regionalAnalyses.find(query)) {
+            while (cursor.hasNext()) {
+                RegionalAnalysis regionalAnalysis = cursor.next();
```
abyrd, line 32: It might be good to add some logging in here so we can verify the effects of the migration when it runs. Or maybe better, enable DEBUG level logging when we run the migration to hit the log statements inside `getSingleCutoffGrid`.

src/main/java/com/conveyal/analysis/GenerateRegionalAnalysisResults.java
trevorgerhardt: Updated to use only those three formats in https://github.com/conveyal/r5/pull/957/commits/b85e57c5408eaa3a10321b30db705ec44e4d00b3

src/main/java/com/conveyal/analysis/GenerateRegionalAnalysisResults.java
```diff
@@ -0,0 +1,64 @@
...
+public class GenerateRegionalAnalysisResults implements TaskAction {
+    private final FileStorage fileStorage;
+    public GenerateRegionalAnalysisResults (FileStorage fileStorage) {
+        this.fileStorage = fileStorage;
+    }
+
+    @Override
+    public void action(ProgressListener progressListener) throws Exception {
+        DBObject query = QueryBuilder.start().or(
+                QueryBuilder.start("travleTimePercentiles").is(null).get(),
```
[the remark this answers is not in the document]
trevorgerhardt: Good catch 🙏, fixed in https://github.com/conveyal/r5/pull/957/commits/b85e57c5408eaa3a10321b30db705ec44e4d00b3

src/main/java/com/conveyal/analysis/GenerateRegionalAnalysisResults.java
trevorgerhardt: Added a few log statements in https://github.com/conveyal/r5/pull/957/commits/c2cc8f4a25dcc03ac27ed56201d9d85e611be7c1

src/main/java/com/conveyal/analysis/GenerateRegionalAnalysisResults.java
trevorgerhardt: Modified the description of #956 to reference this PR and recommend that this one should be merged and run first.

trevorgerhardt:
> What's the planned way to run this on staging and production before #956 is merged? I guess we would perform one deployment of the patched version, observe the logs, and perform another deployment with the post-956 version soon thereafter if everything looked good?

As of right now, I think this approach would be best. There are changes in the dev branch that have not been deployed from PRs #953 and #954, however those changes are very minor. 

If we take that approach, #956 could also remove the code in this PR.

src/main/java/com/conveyal/analysis/GenerateRegionalAnalysisResults.java
abyrd: Thanks for adding some logging. To clarify, the underlying reason I was interested in logging was to observe whether the effects of the migration matched expectations. Rather than just counts, something that would tip us off if for example records were being missed due to an incorrect field name, or extra files were being produced etc. This is why I mentioned enabling debug for `getSingleCutoffGrid` as it would log each new file created and stored. Maybe we could also log the key fields of each regionalAnalysis record that was migrated.

ansoncfit (commented):
One thing to consider: how will the result generation handle cases where the destination opportunity dataset referenced by a regional analysis has been deleted? It should be possible to generate the single cutoff grids using just the id of the destination dataset. The name of the dataset is only required for the `resultHumanFilename`. Fallbacks for generating human-readable filenames when referenced data are missing would be a separate change. I just wanted to flag this edge case now in case it causes failures with the migration.

@trevorgerhardt, how does the following sound?
1. Log at the `DEBUG` level for the method in question as @abyrd suggested above (add `<logger name="com.conveyal.analysis.controllers.RegionalAnalysisController" level="DEBUG" />` to logback.xml?). Or if you think logging `regionalAnalysis._id` and `regionalAnalysis.accessGroup` as currently proposed (plus the existing `INFO` and `WARN` level statements in `getSingleCutoffGrid` will be enough detail, that's fine.
2. Take a snapshot of the staging database
3. Run the migration and check results on staging, reporting back on log statements and time required

abyrd:
We may also want to do a dry run or equivalent read-only database query on production without changing anything, just to see how many records / results are concerned here and how old they are. We may be addressing a limited number of very old results. If customers confirm that these results are quite dated and will rarely or never be used, it may be reasonable to abandon them or do a partial conversion that does not cover every edge case.

trevorgerhardt:
I modified the log statements to use `.info` instead of `.debug`. Additionally, I refactored the static `getSingleCutoffGrid` method to only require the destination point set string ID. The full database entry was only retrieved to generate a human name for downloading, which we don't need to do here (but would be its own problem that already exists).

There are currently 167 entries that match the MongoDB query on staging. I will take a snapshot and then run the migration there.

trevorgerhardt:
The changes in this PR were run on staging and from a local machine on production data with the results stored in our production S3 bucket. Closing.

PR #962, opened 2025-04-05
Allow access and dual access together

trevorgerhardt:
Support both types of accessibility analysis in a single regional analysis


ansoncfit (commented):
After testing this version on staging and giving the changes a quick look, I don't see any issues. I'll let Andrew take another look today before approval.

abyrd:
Looks good to me. As previously mentioned, TemporalDensityCsvResultWriter#rowValues performs its own dual access calculation which may be replicating information already available in workResult.dualAccessValues. I don't see any obvious reason to perform the calculation separately in the CSV writer but I may be missing some detail.

abyrd:
One other thing though, as discussed in chat: given that `calculateDualAccessForOrigin` is called within the RegionalWorkResult constructor, do we want to move that constructor call outside the synchronized block to keep the critical section as short as possible?

trevorgerhardt:
> One other thing though, as discussed in chat: given that `calculateDualAccessForOrigin` is called within the RegionalWorkResult constructor, do we want to move that constructor call outside the synchronized block to keep the critical section as short as possible?

Done in https://github.com/conveyal/r5/pull/962/commits/60be06a2ec18da064d6c22a3c6126f17d133ac58

trevorgerhardt:
> Looks good to me. As previously mentioned, TemporalDensityCsvResultWriter#rowValues performs its own dual access calculation which may be replicating information already available in workResult.dualAccessValues. I don't see any obvious reason to perform the calculation separately in the CSV writer but I may be missing some detail.

I opened https://github.com/conveyal/r5/pull/964 to address that issue

ansoncfit (approved):

PR #964, opened 2025-04-10
TemporalDensityCSV: Re-use dual access calculations and allow multiple thresholds

trevorgerhardt:
- Re-use the previously calculated dual access threshold values when writing out a CSV file.
- Allow more than one dual access threshold to be used when creating and writing the CSV. Uses the heading `dual ${threshold}` for the column in the CSV.
- Additionally, allow a `TemporalDensityCSV` to be created without any thresholds specified.


ansoncfit (approved):
Thanks -- the change to allow multiple thresholds to be written to a CSV ended up being smaller than I expected.

PR #965, opened 2025-04-11
Temporal CSV Fix

trevorgerhardt:
This PR modifies `CsvResultWriter` to set `this.task` before calling `columnHeaders` and changes `TemporalDensityCSVResultWriter` to use `dualAccessThresholds` directly from the regional task. The previous `columnHeaders` changes were attempting to access `dualAccessThresholds` before they were set which caused an NPE.

As previously noted in comments about `columnHeaders` there is probably a better solution that should be explored in the future.


ansoncfit (approved):
Just tested multi-threshold dual access CSVs on staging.

PR #972, opened 2025-07-25
Ignore custom cost fields in CAR routing

### LINKED ISSUE #970: GTFS and Elevation data affects carspeeds
Hi all,

In the dev of {r5r}, we have found that GTFS and Elevation data affects car speeds (see [here](https://github.com/ipeaGIT/r5r/issues/458)). We're not entirely sure this is a bug or intended behavior, but we couldn't find any info on this in R5 documentation.

ansoncfit:
This ignores custom costs (e.g. from elevation data) for the CAR street mode. Custom speeds for cars are handled separately.

Fixes #970

### DIFF
--- a/src/main/java/com/conveyal/r5/rastercost/CostField.java
+++ b/src/main/java/com/conveyal/r5/rastercost/CostField.java
@@ -13,7 +13,16 @@ import java.io.File;
 import java.util.List;
 
 /**
- * Subclasses or plugin functions will provide costs for hills, sun, noise, pollution etc.
+ * This models a field of traversal costs, in the sense of a cost that varies over geographic space having different
+ * effects on different edges. Subclasses serve as plugins that provide costs for hills, sun, noise, pollution etc.
+ *
+ * CostFields could in principle apply costs to any mode of travel, and might yield different costs for different modes,
+ * but currently they are not mode-aware. No existing implementations apply to cars, so for now CostFields are simply
+ * ignored for car routing. Ideally CostFields would know which modes they applied to, and the list of CostFields would
+ * be filtered before the street search happens. The logic currently in the StreetRouter constructor could be factored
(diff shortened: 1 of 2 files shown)


abyrd:
Note, this PR is cherry-picking commit d5722b651835fd59941d1094aa78fc1a3b70103f from the broader PR #812.

abyrd (commented):
This change assumes all CostField instances apply to all modes except cars. CostFields were intended to be generic, and could in principle apply costs to walking, bicycling, or driving. Our assumption was that users would know which modes they intended their costs to apply to, and would only employ such graphs for the intended modes.

In practice, the existing CostField implementations (elevation and sun exposure) do not apply to driving, so this change does provide an immediate improvement to car results.

However, it gives a false impression that mode is taken into account by CostFields, and bakes the undocumented assumption about which modes CostFields apply to into traversal logic. We should clearly document this assumption and the fact that it is an unrealistic simplification, as this has the potential to cause confusion in future development.

Although sun and hill costs arguably apply to both walking and biking, these costs would likely be different for biking. CostFields do not receive any information about what mode of transport is in use, so they cannot modify their behavior per mode. Implicitly, a CostField only applies to a single mode of transport (rather than "all modes but car").

Ideally CostFields would know which modes they applied to, and the list of CostFields would be filtered before the search happens. The logic currently in the StreetRouter constructor could be factored into a mode-aware factory method that would yield the correct TraversalTimeCalculator.

We  don't need to implement this now, but should document it in the Javadoc of the relevant classes. I will propose some simple Javadoc changes.

abyrd:
StreetRouter is also explicitly single-use and can only have one StreetMode, so that mode could be specified up front and the TraversalTimeCalculator chosen accordingly (avoiding the MultistageTraversalTimeCalculator construct entirely when no CostFields apply).

abyrd (approved):
Approved (added two commits with Javadoc and line comments)

PR #976, opened 2025-08-14
Additional TraversalPermissionLabelers

ansoncfit:
We've had a few requests for custom routing profiles, including:
- Disallow biking on sidewalks (`NoSidewalkCyclingTraversalPermissionLabeler` in 7f4282a56dc4f02f04141ffe35c5416fe4220f48)
- Disallow walking on links where custom tags indicate steep inclines, and optionally prevent traversal of steps/raise curbs (`NoSteepInclinesTraversalPermissionLabeler`)

This PR adds these capabilities as custom TraversalPermissionLabelers, which can be applied at network build using TransportNetworkConfig options.


src/main/java/com/conveyal/r5/streets/StreetLayer.java
```diff
@@ -219,7 +219,7 @@ public StreetLayer(TransportNetworkConfig config) {
             permissionLabeler = switch (config.traversalPermissionLabeler) {
                 case "sidewalk" -> new SidewalkTraversalPermissionLabeler();
                 case "noSidewalkCycling" -> new NoSidewalkCyclingTraversalPermissionLabeler();
-                case null -> new USTraversalPermissionLabeler();
+                case null -> new NoSidewalkCyclingTraversalPermissionLabeler();
```
abyrd: Is the intent here to make the default NoSidewalkCycling? Does this mean that now there's no way to select USTraversalPermissionLabeler?

ansoncfit: No, the intent is to keep USTraversalPermissionLabeler. This default case was just set for testing but should be reverted before final review of this PR:
> https://github.com/conveyal/r5/commit/49a205ac6ed97c1f0a96e3780dcad238b384d9c0 forces use of the NoSidewalkCyclingTraversalPermissionLabeler and should be reverted before merging.

src/main/java/com/conveyal/r5/streets/StreetLayer.java
```diff
@@ -225,6 +232,10 @@ public StreetLayer(TransportNetworkConfig config) {
         } else {
             permissionLabeler = new USTraversalPermissionLabeler();
         }
+
+        if (permissionLabeler instanceof StepFreeTraversalPermissionLabeler) {
+            stepFree = ((StepFreeTraversalPermissionLabeler) permissionLabeler).requireStepFree;
```
abyrd: It feels strange to set this via a way permission labeler, as the labeler instance can't take the action itself and we need to detect its type here. Could stepFree just be considered another separate TransportNetworkConfig option? The boolean field could be pulled up out of StepFreeTraversalPermissionLabeler into TransportNetworkConfig itself. This would then become `stepFree = config.stepFree` with no instanceof or conditional. The purpose of the field here seems to be only blocking kerbs (not stair/step ways) so perhaps it should be renamed accordingly (e.g. kerbFree or requireKerbCuts).

src/main/java/com/conveyal/r5/labeling/StepFreeTraversalPermissionLabeler.java
```diff
@@ -0,0 +1,45 @@
+package com.conveyal.r5.labeling;
+
+import com.conveyal.osmlib.Way;
+import com.conveyal.r5.streets.EdgeStore;
+
+/**
+ * USTraversalPermissionLabeler, except walking is disallowed on ways that allow cars, ways with inclines
+ * steeper than a specified limit, and ways with steps.
+ */
+public class StepFreeTraversalPermissionLabeler extends USTraversalPermissionLabeler {
```
abyrd: Ideally we should avoid the duplicate code between this and NoSteepInclinesTraversalPermissionLabeler. Could this class just extend NoSteepInclinesTraversalPermissionLabeler?  Could they just be a single class with a boolean for disallowing steps (stairs) passed into the constructor? This might be further simplified by pulling the requireStepFree field up to the enclosing NetworkConfig.

src/main/java/com/conveyal/r5/streets/StreetLayer.java
ansoncfit: Yes, it makes sense to make `stepFree` another TransportNetworkConfig option.

Ways with highway=steps tags are filtered out where appropriate in the custom TraversalPermissionLabelers.

abyrd (approved):
Thanks for the updates, this way of structuring the labelers seems quite easy to read and understand. I have not tested this on appropriate data sources with lots of sidewalk and curb/kerb tags, but would be in favor of merging it to dev as-is for testing in combination with other PRs.

Issue #934, opened 2024-02-28
Snapped distance not being calculated as crow flies "walking" distance (r5r)


Mponkane:
### Brief description of the problem:

Here is an issue I found while using r5r, perhaps an upstream R5 issue? Originally posted: https://github.com/ipeaGIT/r5r/issues/373#issue-2156986527

When origin is snapped to network, it does not account for the crow flies distance from the original location.  See picture below. 

![issue](https://github.com/ipeaGIT/r5r/assets/105248249/c46e87e8-7614-4793-ad64-c6323eb36177)

The travel time matrices result into 0 even if I increase the distance of the origin point up to 1.6 km from the network. Both origin and destination are in WGS 84

here is the ttm result:
   from_id       to_id travel_time_p50
1:  origin destination               0

here is find_snap for origin:
   point_id,      lat,      lon, snap_lat, snap_lon, distance, found
1:   origin, 65.05129, 25.42877, 65.04899, 25.42855, 255.8379,  TRUE

here is find_snap for destination:
     point_id,      lat,      lon, snap_lat, snap_lon, distance, found
1: destination, 65.04884, 25.42872, 65.04883, 25.42872, 1.176363,  TRUE

### Reproducible example here

```
options(java.parameters = "-Xmx20G")
library(r5r)
library(sf)
library(dplyr)
library(ggplot2)

  
r5r_core <- setup_r5("path")
origin <- st_read("test_origin.gpkg")
destination <- st_read("test_dest.gpkg")


det <- detailed_itineraries(
  r5r_core,
  origins = origin,
  destinations = destination,
  mode = "BICYCLE",
  max_trip_duration = 30,
  max_lts = 2,
  bike_speed = 15,
  walk_speed = 2,
  shortest_path = TRUE,
  all_to_all = TRUE
)

ttm <- travel_time_matrix(
  r5r_core,
  origins = origin,
  destinations = destination,
  mode = "BICYCLE",
  max_trip_duration = 30,
  max_lts = 2,
  bike_speed = 15,
  walk_speed = 2
)

snap_df <- find_snap(r5r_core, origin, mode = "BICYCLE")
```

### Situation report
$r5r_package_version
[1] ‘1.1.0’

$r5_jar_version
[1] "6.9"

$java_version
[1] "11.0.18"

$set_memory
[1] "-Xmx20G"

$session_info
R version 4.2.2 (2022-10-31 ucrt)
Platform: x86_64-w64-mingw32/x64 (64-bit)
[...]

PR #980, opened 2025-11-18
Use walking speed from request when linking origin and destination points to network

ansoncfit:
R5 uses a walking speed when setting an origin point for street routing, and when linking (destination) points to the street network. Previously, we mixed use of a fixed constant (`OFF_STREET_SPEED_MILLIMETERS_PER_SECOND`) and the user-requested walk speed from analysis requests. This PR uses the user-requested walk speed more consistently. 

On the origin end, the walk speed is readily accessible in the `profileRequest` field of `StreetRouter`, so the change seems straightforward.

On the destination end, Egress Cost Tables store distances (in millimeters) for walking and biking linkages, and speeds are applied at analysis time when the user-requested walk speed is readily accessible. For car, on the other hand, times (in seconds) are stored; dividing distance by speed happens when creating the Egress Cost Tables, so it is simpler to maintain the use of the constant `OFF_STREET_SPEED_MILLIMETERS_PER_SECOND` (and we don't have known cases of users wanting to tweak the walking times between the street network and origins/destination points when the street mode is car).

More background is at https://github.com/conveyal/r5/issues/934#issuecomment-2090835628

### DIFF
--- a/src/main/java/com/conveyal/r5/streets/StreetRouter.java
+++ b/src/main/java/com/conveyal/r5/streets/StreetRouter.java
@@ -337,7 +337,7 @@ public class StreetRouter {
         State startState0 = new State(split.vertex0, split.edge + 1, streetMode);
         State startState1 = new State(split.vertex1, split.edge, streetMode);
         EdgeStore.Edge  edge = streetLayer.edgeStore.getCursor(split.edge);
-        int offStreetTime = split.distanceToEdge_mm / OFF_STREET_SPEED_MILLIMETERS_PER_SECOND;
+        int offStreetTime = split.distanceToEdge_mm / (int) (profileRequest.walkSpeed * 1000);
 
         // Uses weight based on distance from end vertices, and speed on edge which depends on transport mode
         float speedMetersPerSecond = edge.calculateSpeed(profileRequest, streetMode);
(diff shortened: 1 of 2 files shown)


src/main/java/com/conveyal/r5/streets/StreetRouter.java
```diff
@@ -337,7 +337,7 @@ public boolean setOrigin (double lat, double lon) {
         State startState0 = new State(split.vertex0, split.edge + 1, streetMode);
         State startState1 = new State(split.vertex1, split.edge, streetMode);
         EdgeStore.Edge  edge = streetLayer.edgeStore.getCursor(split.edge);
-        int offStreetTime = split.distanceToEdge_mm / OFF_STREET_SPEED_MILLIMETERS_PER_SECOND;
+        int offStreetTime = split.distanceToEdge_mm / (int) (profileRequest.walkSpeed * 1000);
```
abyrd, line 340: It may make more sense to perform floating point multiplication and division and only convert the result to int before storing it in offStreetTime. But integer division was already being used on mm-per-second speeds in related code in LinkedPointSet so this consistent, and rethinking the use of integer and float operations is not within scope for this PR.

abyrd (approved):

PR #985, opened 2025-12-04
Accelerate GeoTiff extraction (experiments)

abyrd:
Extracting single channels out of an access file with the SelectingGridReducer is surprisingly slow. Past research shows this is not due to file transfer to cloud storage, but has something to do with the local file IO. The slowness is mostly tolerable when for example displaying single cutoffs of regional analysis results, but becomes a problem when working with very large data sets or extracting many channels at once, as when making a ZIP of GeoTiffs for every combination of parameters.

This branch contains some experiments to further narrow down the cause and test some alternative approaches.

Related past issues and comments:
- https://github.com/conveyal/r5/pull/932#issuecomment-2049435867 (some measurements and theories)
- https://github.com/conveyal/r5/pull/938 (current workaround, multi-channel extraction done as a background task)

Some ongoing UI changes would be much simpler if channel extraction was fast enough to be performed synchronously by the backend. We could then avoid exposing separate extract and download steps to the end user.

### DIFF
--- a/src/main/java/com/conveyal/analysis/SelectingGridReducer.java
+++ b/src/main/java/com/conveyal/analysis/SelectingGridReducer.java
@@ -2,6 +2,7 @@ package com.conveyal.analysis;
 
 import com.conveyal.r5.analyst.Grid;
 
+import java.io.BufferedInputStream;
 import java.io.DataInputStream;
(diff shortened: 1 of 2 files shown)


abyrd:
Superseded by #986

PR #987, opened 2025-12-05
Prepare regional result geotiff zip files in blocking endpoint

abyrd:
Prepares regional result geotiff zip files in a blocking endpoint. These were previously prepared in a background task action because the process was very slow.

This PR reverses PR #938 which is hopefully no longer necessary after PR #986. I have tested this endpoint on smaller regional results (32k origins) but we should test it with much larger areas to ensure it's still reasonably fast.

The code does not take extra measures to prevent simultaneous requests for the same results, as this should not happen in usual operation from a web UI, and most actions taken are idempotent so the only expected side effect is wasted compute (and maybe exceptions from threads trying set files read-only while other threads are replacing them).

### DIFF
--- a/src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
+++ b/src/main/java/com/conveyal/analysis/controllers/RegionalAnalysisController.java
@@ -270,10 +270,6 @@ public class RegionalAnalysisController implements HttpController {
         return new HumanKey(singleCutoffFileStorageKey, resultHumanFilename);
     }
 
-    // Prevent multiple requests from creating the same files in parallel.
-    // This could potentially be integrated into FileStorage with enum return values or an additional boolean method.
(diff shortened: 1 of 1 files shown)


abyrd:
@ansoncfit reports that on staging, fetching all GeoTiffs for regional analysis 63ab193ac84e4152d58fbb25 (~536k origins x 5 cutoffs x 5 percentiles) takes ~13 seconds to prepare, which is still too slow for a synchronous fetch. Interestingly running locally on exactly the same result grids, it takes only 3 seconds. I've added some logging to identify where the differences may lie when running on the staging environment.

abyrd:
On staging, I observed producing the same GeoTiffs take 27 seconds, while zipping them and uploading to s3 took 300 and 400 msec respectively. It appears that only the read-intensive part of the operation is about 10x slower.

PR #989, opened 2025-12-09
Make zips of dual access grids

ansoncfit:


abyrd:
Thanks for the change to extract valid dual thresholds @trevorgerhardt. While reviewing and testing this I noticed we have code paths to infer appropriate cutoff and percentile values when the parameters are missing. I believe this capability hasn't been used for a very long time and stricter validation would be more appropriate. I tried removing this inference and factoring out some of the repetitive validation logic, using more readable constructs like Arrays.toString, List.of, and Ints.contains. See two added commits.

I have tested this via the web UI and CURL API access, with all query parameters absent, present, correct, and incorrect, and all code paths seem to work as intended.

ansoncfit (commented):
Looks good to me. I opened the PR so I cannot select "approve"; I'll go ahead and merge.

PR #999, opened 2026-06-04
(fix) elevation costs when copying a street network

alexmgns:
Currently when using `scenarioCopy` in TransportNetwork the elevation costs are dropped during the "deep copy" if the scenario announces that it will mutate the street layer. 

I added a small fix to copy over the raster costs to the new network. This is not a deep copy similarly to how some of the other fields are copied over, as at least in R5R we do not expect to change elevation in a scenario. However I don't think making a deep copy would be difficult I just didn't want to miss something. 

Elevation is now properly preserved in R5R when we make copies, not sure how to test from your guys' end.

### DIFF
--- a/src/main/java/com/conveyal/r5/rastercost/CostField.java
+++ b/src/main/java/com/conveyal/r5/rastercost/CostField.java
@@ -44,6 +44,14 @@ public interface CostField {
      */
     double getDisplayValue (int edgeIndex);
 
+    /**
+     * Return a copy of this CostField when duplicating for scenarios.
+     * If field won't be mutated in scenario, may return this.
+     */
+    default CostField extendOnlyCopy (EdgeStore copiedEdgeStore) {
+        return this;
+    }
+
     /** Interface for classes that create a CostField for a given StreetLayer, usually by overlaying a raster file. */
     interface Loader<T extends CostField> {
         void setNorthShiftMeters (double northShiftMeters);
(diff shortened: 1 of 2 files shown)


abyrd:
Hi, although it's clearly not correct that we totally drop the `copy.costFields` and this prevents analyzing scenarios when costFields are present, I don't think simply copying the list will work. The words "copy" and "extend-only" are important here. Any scenario that changes the street layer, including transit scenarios that might split a street into two segments to connect a new transit stop, will have more edges than the base network they were applied to. The edge splitting or creation process adds items to the end of every list of edge characteristics, describing those new edges. New numbered edges are then present that require lookups at slots not present in the lists of edge characteristics for the base network. In both existing implementations of CostField, routing on a scenario that has a copy of a reference to edge data for the base network is then expected to throw an ArrayIndexOutOfBoundsException.

It's conceivable to alter the costField implementations so they'd just return zero extra seconds for any edge ID above their known range instead of throwing an exception, but this would silently replace the nonzero costs of edges in the base network with zero costs on their split fragments in the scenario network.

There also needs to be machinery for properly copying the cost data to new edges as those edges are created, or at least mapping new edge IDs to the ones they were built from so lookup is possible. If you put this machinery in place but simply copy the base network's costField references, all edge creation in scenarios will be writing to and attempting to expand the same set of costField internal structures, corrupting each other's data as well as the base network. Every scenario needs its own copy, that copy must be extensible, and the mechanism for extension and the replication of edge characteristics needs to exist.

For other kinds of edge data we use com.conveyal.r5.trove.TIntAugmentedList to wrap the base data in a list that can expand upon it without creating a complete copy. Any full solution would probably use the same kind of mechanism, but needs to be generic in a way that works across all costField implementations.

Issue #995, opened 2026-05-07
Check whether workers are needed repeatedly in a regional analysis, possibly on-demand rather than spot instances


ansoncfit:
Lately I've noticed increased terminations of spot instances (perhaps related to AWS degradations in other regions, though that may be purely coincidental). Spot instance termination can leave hosted service users with a small number of workers handling a large/slow regional analysis.

In our current setup, additional workers can be requested only once per regional analysis:
https://github.com/conveyal/r5/blob/81a25c8c562069a236d28f486e4f2bc78ac19fdc/src/main/java/com/conveyal/analysis/components/broker/Broker.java#L536-L540

The startup happens when the broker handles task 42 (a number slightly greater than the target tasks per worker). https://github.com/conveyal/r5/blob/81a25c8c562069a236d28f486e4f2bc78ac19fdc/src/main/java/com/conveyal/analysis/components/broker/Broker.java#L138

We may want to check if starting more workers is appropriate every so often after this initial startup check (say every 1000 tasks)?

I have also encountered situations where spot capacity is not available, but on-demand capacity is. If workers need to be started multiple times during the course of a regional analysis, that may be a sign we should request on-demand instances instead.

abyrd:
It is probably a good idea to perform more regular, active shaping of the worker pool. This was always the intent but we always erred on the side of a simpler policy. There will be additional, less obvious failure modes, for example where requests to start instances are queued up for a long time, additional requests pile up and we over-provision. We could just identify the most common or problematic failure modes and add a couple of checks, ensuring there is no backlog for example. As long as all spot and on-demand instances are configured to auto-shutdown quickly in the absence of work, it should be fine.

Somewhat relatedly, we may also want to look into enabling migration of workers from one network to another and keeping a small permanent fleet of reserved instances (or bare-metal servers).

ansoncfit:
Closed by #1000

PR #1000, opened 2026-07-01
Consider additional worker startup regularly during regional analysis jobs

ansoncfit:
This PR has a couple options to address #995 and make regional analysis progress smoother for users.

The first commit (9771a0d610c81a8590ec05e17d317d39346faddc) checks whether the number of running workers matches the target number every 4000 tasks in a regional job, then requests spot instances to make up any difference.

The second commit (b501a55fcf28edf32fd3424c1db8dc5c6fc1d07b) makes these subsequent requests for on-demand workers instead of spot workers, because if workers are missing in a regional job, it may be because AWS capacity issues have led to termination of our spot instances.

The third commit (a305d2164b9f81fe63d04ca24dff50311470ebef) is purely an indentation change, separated for more readable diffs.

Other approaches would be welcome, but I wanted to propose some initial minimal changes to address the tighter spot markets we've been seeing lately.

### DIFF
--- a/src/main/java/com/conveyal/analysis/components/broker/Broker.java
+++ b/src/main/java/com/conveyal/analysis/components/broker/Broker.java
@@ -135,7 +135,13 @@ public class Broker implements Component {
      * results are received for a given workerCategory. Do so after receiving results for an
      * arbitrary task toward the beginning of the job
      */
-    public static final int AUTO_START_SPOT_INSTANCES_AT_TASK = 42;
+    public static final int START_INSTANCES_TASK = 42;
+
+    /**
+     * We also want to consider starting instances occasionally as a regional analysis is running, in case previously
+     * created instances have been terminated.
+     */
+    public static final int RESTART_INSTANCES_TASKS = 4000;
 
     /** The maximum number of spot instances allowable in an automatic request */
     public static final int MAX_WORKERS_PER_CATEGORY = 250;
@@ -535,7 +541,7 @@ public class Broker implements Component {
         }
         // When non-error results are received for several tasks we assume the regional analysis is running smoothly.
         // Consider accelerating the job by starting an appropriate number of EC2 spot instances.
-        if (workResult.taskId == AUTO_START_SPOT_INSTANCES_AT_TASK) {
+        if (workResult.taskId == START_INSTANCES_TASK || (workResult.taskId + 1) % RESTART_INSTANCES_TASKS == 0) {
             requestExtraWorkersIfAppropriate(job);
         }
     }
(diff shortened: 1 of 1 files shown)


src/main/java/com/conveyal/analysis/components/broker/Broker.java
```diff
@@ -533,43 +539,50 @@ public void handleRegionalWorkResult(RegionalWorkResult workResult) {
...
-            createWorkersInCategory(job.workerCategory, job.workerTags, 0, nSpot);
+                // Do not exceed the limit on workers per category TODO add similar limit per accessGroup or user
+                targetWorkerTotal = Math.min(targetWorkerTotal, MAX_WORKERS_PER_CATEGORY);
+                // Guardrails until freeform pointsets are tested more thoroughly
+                if (job.templateTask.originPointSet != null) targetWorkerTotal = Math.min(targetWorkerTotal, 80);
+                if (job.templateTask.includePathResults) targetWorkerTotal = Math.min(targetWorkerTotal, 20);
+                int nWorkers = targetWorkerTotal - categoryWorkersAlreadyRunning;
+                if (taskId == START_INSTANCES_TASK) {
+                    // After a few tasks are completed successfully, try to start spot instances
+                    createWorkersInCategory(job.workerCategory, job.workerTags, 0, nWorkers);
```
abyrd: My first concern was that this might be a synchronous operation and block the broker waiting on a cloud computing API request. Handling that correctly is be the responsibility of the WorkerLauncher implementation. I just checked `cluster` and in our case it's done in a task scheduler "light task" in the background. So it looks nonblocking other than the very brief synchronized calls to the worker catalog above.

src/main/java/com/conveyal/analysis/components/broker/Broker.java
```diff
@@ -533,43 +539,50 @@ public void handleRegionalWorkResult(RegionalWorkResult workResult) {
...
-                float transitScaleFactor = (9f / job.templateTask.zoom);
-                targetWorkerTotal = (int) ((job.nTasksTotal / TARGET_TASKS_PER_WORKER_TRANSIT) * transitScaleFactor);
-            } else {
-                // Tasks without transit are simpler. They complete relatively quickly, and the total computation for
-                // the job increases roughly with linearly with the number of origins.
-                targetWorkerTotal = job.nTasksTotal / TARGET_TASKS_PER_WORKER_NONTRANSIT;
-            }
+    // At certain task numbers, check various conditions about the worker pool then start spot or on-demand instances
+    private void requestExtraWorkersIfAppropriate(Job job, int taskId) {
+        if (taskId == START_INSTANCES_TASK || (taskId + 1) % RESTART_INSTANCES_TASKS == 0) {
```
abyrd: A conditional gating the entire function body can leave the reader sifting through closing brackets to confirm whether there's an else clause and what it does. I suggest negating the condition and returning early: `if ( ! (taskId == START_INSTANCES_TASK || (taskId + 1) % RESTART_INSTANCES_TASKS == 0)) return;` then pulling the whole rest of the method out of the block.

abyrd (dismissed):
Looks like it would work. Keeping the change so simple and incremental should make it easy to diagnose if we get any unexpected behavior.

PR #1002, opened 2026-07-24
Reduce modification rounding and distance error

abyrd:
Accumulate fractional travel times, rounding only at stops .


abyrd:
Confirmed that merging branch `abyrd/simpson-scheduled-tests` into this one brings the number of failing tests on that other branch down from two to zero, with none of the other new tests failing.

src/main/java/com/conveyal/analysis/models/ModificationStop.java
```diff
@@ -241,11 +221,11 @@ static int[] getHopTimes (List<ModificationStop> stops) {
         if (stops == null || stops.size() < 2) {
             return new int[0];
         }
-
         // First stop has no hop.
         int[] hopTimes = new int[stops.size() - 1];
-        for (int i = 1; i < stops.size(); i++) hopTimes[i - 1] = stops.get(i).hopTimeSeconds;
-
```
ansoncfit, line 248: Whitespace changes only

src/main/java/com/conveyal/analysis/models/ModificationStop.java
```diff
@@ -81,17 +72,19 @@ static List<ModificationStop> getStopsFromSegments (List<Segment> segments, Inte
...
+        double secondsFromPatternStart = 0; // travel time from start of pattern, in unrounded double precision
+        double secondsAtLastStop = 0; // travel time from start of pattern to the previously created stop
+        double secondsAtPenultimateStop = 0; // like secondsAtLastStop, but for the stop created before it
+        double previousSegmentSpeedKph = segmentSpeedsKph.length > 0 ? segmentSpeedsKph[0] : DEFAULT_SEGMENT_SPEED_KPH;
         Coordinate previousSegmentFinalCoordinate = null;
 
-        // Iterate over segments, which are defined by control points and real (i.e. not auto-generated at a requested
-        // spacing) stops.
+        // Iterate over segments, which are real stops defined by control points
+        // (i.e. not auto-generated at a requested spacing)
```
ansoncfit: The revision seems to be equating control points and stops.

In my understanding of UI parlance, stops are not control points. If a segment has `stopAtStart = false`, its start is a control point (and similarly for the end of the segment).

ansoncfit (dismissed):
One question about a revised code comment is below. Otherwise, let's merge.

src/main/java/com/conveyal/analysis/models/ModificationStop.java
abyrd: Yes, control points and "real stops" are the two mutually exclusive ways a point in the modification can be categorized. I was focused on the "real" vs. generated distinction and moving the parenthetical somewhere I found less confusing. I must have been thinking points=coordinates rather than specifically "control points". I'll correct it.

PR #1003, opened 2026-07-24
Grid transform fix

abyrd:
This fixes a single line that was truncating the bottom of stacked grids of different dimensions. This only applied to regional analyses, not single-point, and only in cases where grids had different dimensions. But in those cases it would affect visibility of opportunities near the bottom/south edge.

Includes a couple of tests to reproduce the problem.


ansoncfit (approved):
Thanks for the fix and tests.

PR #1004, opened 2026-07-26
Expand Simpson Desert tests

abyrd:
We have a few changes lined up in other branches that could potentially affect routing results. These include rounding in conversion of add and reroute modifications, new kinds of on-demand/flex routing, stacked grid bounds fixes, and eventually a refactor of turn cost and turn restriction handling to speed up non-car street routing.

This more extensive testing of travel time and accessibility results should provide better warnings if anything changes in routing results as those changes and fixes are applied, and increase confidence in stability and correctness for a release.

Other than adding a constructor and an invariant check this PR purely adds tests, so should have no effect on production behavior. Three tests will fail intentionally, with follow up PRs making them pass. There is some replication of testing between unit tests in the fix branches and the tests here, with these hitting more layers of the production system. We may choose to merge some tests later to reduce maintenance load.


abyrd:
As expected, failing tests decreased from 3 to 2 after merging from dev, which contained the grid bounds math fix.

ansoncfit (approved):
Gave this a quick scan 👍

PR #1006, opened 2026-07-26
Rework path reporting

abyrd:
This PR responds to reports of inconsistencies in transfer times in paths CSV, as well as an interest in path data that might be more readily interpretable as a proxy for network assignment in traffic models. The shift to storing clock times (from which variable wait times are later derived) makes paths reusable at different departure minutes and simplifies things a bit while enabling more well-defined summary statistics.

This also stores paths only at those stops actually used in propagation to destination points, and performs path bookkeeping only when paths are enabled, potentially reducing resource consumption.

Note that this will change the number and meaning of CSV columns reported to the backend, so rollout requires careful coordination.


src/main/java/com/conveyal/r5/analyst/cluster/PathResult.java
```diff
@@ -95,65 +96,93 @@ public void setTarget(int targetIndex, Multimap<PatternSequence, Iteration> patt
...
+    /// Summary of iterations for each destination, suitable for writing to a CSV. Conversion to strings happens here
+    /// (on distributed workers) to minimize pressure on the central Broker's assembler.
+    ///
+    /// For each destination, a row is produced for each distinct route-based path template, summarizing all the
+    /// iterations (departure minutes and Monte Carlo schedules) in which that template was optimal. The fixed
+    /// components of a template (access, egress, ride, and transfer times) are identical in all its iterations.
+    /// Only the waiting times and the total travel time vary. The requested Stat determines how those varying
+    /// quantities are summarized:
+    ///
+    /// - MINIMUM reports the single fastest iteration. The row describes an itinerary a rider could actually follow,
```
ansoncfit, line 108: If multiple itineraries for the given path template are equally fast, the one with the earliest departure time is reported.

src/main/java/com/conveyal/r5/analyst/cluster/PathResult.java
```diff
@@ -95,65 +96,93 @@ public void setTarget(int targetIndex, Multimap<PatternSequence, Iteration> patt
...
 
+    /// Format a duration in seconds as fractional minutes with two decimal places for use in result CSV.
+    /// Two decimal places is the coarsest fractional-minute format that loses no information.
+    /// Rounding error is at most 0.005 minutes (0.3 seconds) so an exact number of seconds is recoverable
+    /// by rounding (minutes * 60). Averages of durations in whole seconds can take on fractional values
+    /// that can't be recovered exactly, but are reported in the same format for consistency.
+    public static String formatMinutes (double seconds) {
+        return String.format("%.2f", seconds / 60d);
+    }
+
```
ansoncfit, line 185: Move to Util class?

src/main/java/com/conveyal/r5/analyst/cluster/PathResult.java
```diff
@@ -95,65 +96,93 @@ public void setTarget(int targetIndex, Multimap<PatternSequence, Iteration> patt
...
-     * @param stat whether the reported per-leg wait times should correspond to the iteration with the minimum or mean
-     *             total waiting time.
-     * @return For each destination, a list of String[]s summarizing path templates. For each path template, details
-     *          of the itinerary with waiting time closest to the requested stat are included.
-     */
+    /// Summary of iterations for each destination, suitable for writing to a CSV. Conversion to strings happens here
+    /// (on distributed workers) to minimize pressure on the central Broker's assembler.
+    ///
+    /// For each destination, a row is produced for each distinct route-based path template, summarizing all the
+    /// iterations (departure minutes and Monte Carlo schedules) in which that template was optimal. The fixed
```
ansoncfit, line 103: Confirming some details for documentation:

If multiple path templates have identical arrival times, only one is retained. That one is based on the first patterns explored (because `RaptorState#setTimeAtStop` uses strict inequalities such as `time < bestTimes[stop]`).

Exact-times patterns are explored first, then frequency-based patterns are explored. Within each, the order of patterns explored at a given stop is ultimately determined by the random order of stored trips (see `TransitLayer#loadFromGtfs`: `TRIPS: for (String tripId : gtfs.trips.keySet()) {`)

src/main/java/com/conveyal/r5/analyst/cluster/PathResult.java
```diff
@@ -64,6 +64,7 @@ public class PathResult {
             "transferTime",
             "waitTimes",
             "totalTime",
+            "departureTime",
```
ansoncfit, line 67: One extra column to add to https://github.com/conveyal/docs/blob/master/docs/guides/compute-freeform.mdx#path-details

Stat.MINIMUM is still hard-coded. If we plan to make that configurable, we'll want to update this section of the user manual accordingly.

src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
```diff
@@ -531,10 +578,7 @@ private void doScheduledSearchForRound (RaptorState outputState) {
                 // This block is above the boarding search so that we don't alight from the same stop where we boarded.
                 if (onTrip != NONE && pattern.dropoffs[stopInPattern] != PickDropType.NONE) {
                     int alightTime = schedule.arrivals[stopInPattern];
-                    int inVehicleTime = alightTime - boardTime;
-                    checkState (alightTime == inputState.bestTimes[boardStop] + waitTime + inVehicleTime,
-                                "Components of travel time are larger than travel time!");
```
ansoncfit, line 536: It looks like this assertion was added ~9 years ago for Taui sites.

src/main/java/com/conveyal/r5/profile/FastRaptorWorker.java
```diff
@@ -707,8 +749,6 @@ private void doFrequencySearchForRound (RaptorState outputState, FrequencyBoardi
                                 newBoardingDepartureTimeAtStop < remainOnBoardDepartureTimeAtStop
                             ) {
                                 boardTime = newBoardingDepartureTimeAtStop;
-                                // TODO is this wait time calculation right?
-                                waitTime = boardTime - inputState.bestTimes[stop];
```
ansoncfit, line 711: Related to the reported inconsistencies in transfer times?

src/main/java/com/conveyal/r5/profile/RaptorState.java
```diff
@@ -228,10 +203,11 @@ public void minMergePrevious () {
      * When transfer is false, times can update both the bestNonTransferTime and the bestTime; when transfer is true,
      * only bestTimes can be updated.
      *
+     * @param boardTime the clock time at which the vehicle departed its boarding stop (ignored for transfers)
      * @param transfer if true, we are recording a time obtained via a transfer or the initial access leg in round 0
      * @return true if the new time was optimal and the state was updated, false if the existing values were better
      */
-    public boolean setTimeAtStop(int stop, int time, int fromPattern, int fromStop, int waitTime, int inVehicleTime, boolean transfer) {
+    public boolean setTimeAtStop(int stop, int time, int fromPattern, int fromStop, int boardTime, boolean transfer) {
```
ansoncfit, line 210: One of the key changes -- storing board (clock) time instead of wait and in-vehicle times (which can be derived from alighting and boarding times).

ansoncfit (commented):
Looks good after an initial review. We should test with a mix of exact-times and frequency-based patterns, as well as patterns with overtaking trips.

Were the reported inconsistencies in transfer times resolved?

ansoncfit (dismissed):

src/main/java/com/conveyal/r5/analyst/cluster/PathResult.java
abyrd: We discussed this and decided this particular format is specific to PathResults. We're only using it there because people are accustomed to seeing easily readable minutes in these tables, and two digits of precision are enough to exactly restore the internal precision of seconds. So we'll keep it in this class rather than the utility class.

src/main/java/com/conveyal/r5/analyst/cluster/PathResult.java
abyrd: Confirmed in recent conversation.

